PR #7, opened 2022-07-19
feat: flake

yajo:
Without this, the module cannot be used with flakes, because it fails with `cannot look up '<nixpkgs/lib>' in pure evaluation mode (use '--impure' to override)`.

@moduon MT-904

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,3 +1,7 @@
 {
-  inherit (import ./lib) config create mount;
-}
+  lib ? pkgs.lib,
+  pkgs ? import <nixpkgs> {overlays = [(import ./overlay.nix)];},
+  diskoEnv ? pkgs.diskoEnv,
+  ...
+}:
+import ./lib { inherit lib diskoEnv; }
--- a/default.nix
+++ b/default.nix
@@ -1,7 +1,4 @@
 {
-  lib ? pkgs.lib,
   pkgs ? import <nixpkgs> {overlays = [(import ./overlay.nix)];},
-  diskoEnv ? pkgs.diskoEnv,
-  ...
 }:
-import ./lib { inherit lib diskoEnv; }
+pkgs.disko.lib
--- a/flake.nix
+++ b/flake.nix
@@ -13,6 +13,7 @@
         };
       in rec {
         lib = import ./. {inherit pkgs;};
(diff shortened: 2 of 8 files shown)


default.nix
```diff
@@ -1,3 +1,10 @@
...
+(import
+  (
+    let lock = builtins.fromJSON (builtins.readFile ./flake.lock); in
+    fetchTarball {
+      url = "https://github.com/edolstra/flake-compat/archive/${lock.nodes.flake-compat.locked.rev}.tar.gz";
+      sha256 = lock.nodes.flake-compat.locked.narHash;
+    }
+  )
+  { src = ./.; }
+).defaultNix.lib.body
```
zimbatm, line 10: I think the intention was to only export `config` `create` and `mount` as official interfaces. Something like that would better fit the spirit and better compose with non-flake users:
```nix
{ lib ? import <nixpkgs/lib> }:
let my_lib = import ./lib { inherit lib; }; in
{
  inherit (my_lib) config create mount;
}
```

flake.nix
```diff
@@ -0,0 +1,12 @@
+{
+  description = "nix-powered automatic disk partitioning";
+
+  inputs.flake-compat = {
+    url = "github:edolstra/flake-compat";
+    flake = false;
+  };
+
+  outputs = { self, nixpkgs, ... }: {
+    lib = import ./lib {inherit (nixpkgs) lib;};
```
zimbatm, line 10:
```suggestion
    lib = import ./. {inherit (nixpkgs) lib;};
```
Only export `config`, `create` and `mount`

zimbatm (approved):

yajo:
@zimbatm Thanks for the review. I changed this quite a bit now. I think the new changes reflect better your intentions in the review, while still fixing the non-determinism that was still relying around.

I think this looks very better now.

Mic92:
Sorry. I did not saw your pull request first when I added my change. But I also in particular did not like the overlay it added as it re-imports nixpkgs for now reason. But maybe we can as a follow up still do something about adding missing tools. Also I see value in having the option to take those from the installer itself (to avoid copying additional binaries) - so maybe this could be made optional. Anyway its usable as a flake now.

yajo:
No problem, thanks anyway!

PR #15, opened 2022-08-22
use /dev/disk/by-path for better nvme support

Lassulus:


lib/default.nix
```diff
@@ -9,6 +9,25 @@ let {
   body.mount = mount-f {};
 
 
+  helper.find-device = device: let
+    environment = helper.device-id device;
+  in ''
+    if echo '${device} | grep -q '^/dev/disk'; then
```
zimbatm, line 15:
```suggestion
    if echo '${device}' | grep -q '^/dev/disk'; then
```
BUG!

Baitinq:
@Lassulus Thank you for your amazing work.

Mic92:
> @Lassulus Thank you for your amazing work.
> 
> EDIT: What do you think about encouraging this new way of identifying disks? Imho I think its much better as it is more deterministic.

We should document both versions but there are also good reasons to use non-deterministic approaches i.e. when provisioning servers where you don't know the ids in advance.

tests/test.nix
```diff
@@ -107,6 +107,7 @@ in makeTest' {
 
   testScript = ''
     machine.succeed("echo 'secret' > /tmp/secret.key");
+    machine.succeed("ls -la /dev/disk/by-id");
```
Mic92, line 110:
```suggestion
```

Baitinq:
@Lassulus Why did you end up choosing by-path instead of by-id? Just curious, thank you for your work again.

Lassulus:
@Baitinq because of https://github.com/nix-community/disko/pull/15#issuecomment-1223661848 :)

PR #16, opened 2022-08-22
[WIP] Implement by-id disk device partition handling

Baitinq:
Currently adding a device by id messes with the functionality of the script, as it tries appending the partition index to the name when the format for by-id disks is $NAME-part$INDEX instead of the assumed $NAME$index.

This implementation is in progress as we need a way to determine if the user has chosen the traditional or the by-id way to identify a disk in order to handle the partition names correctly.

imho I think we should promote/make-default the use of the by-id way of identifying disks as it provides a more declarative and deterministic way of identifying disks.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -35,7 +35,7 @@ let {
   config.noop = q: x: {};
 
   config.partition = q: x:
-    config-f { device = q.device + toString q.index; } x.content;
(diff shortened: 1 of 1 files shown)


Baitinq:
Closing this, I cant believe we both starting working on this at the same time 😂

Mic92:
It's on master now.

PR #18, opened 2022-08-23
raid support

Mic92:


lib/default.nix
```diff
@@ -67,76 +69,124 @@ let {
...
+
     ${optionalString (hasAttr "lvm" z) (concatStringsSep "\n" (attrValues z.lvm))}
+    ${maybeRaid}
+
     ${optionalString (hasAttr "fs" z) (concatStringsSep "\n" (attrValues z.fs))}
   '';
 
+  mount.mdraid = q: x: (recursiveUpdate (mount-f { device = "/dev/md/${x.name}"; } x.content) {
+    raid.all = ''
+      mdadm --assemble --scan || true
```
Mic92, line 186: This is a bit ugly it would be import all raids explicitly after each step...

PR #19, opened 2022-08-24
style: simpler structure

yajo:
I always get confused with so many `lib`s and `default`s.

It seems to me like all this structure is not really needed and it just complicates things.

Here I propose to simplify it: just a `default.nix` with all that's needed.

I just moved `default.nix` from `/lib` into `/`, and changed the 1st line to add a default value for the parameter.


default.nix
```diff
@@ -1,5 +1,169 @@
-{ lib ? (import <nixpkgs> {}).lib }: {
-  inherit (import ./lib {
-    inherit lib;
-  }) config create mount;
+{ lib ? <nixpkgs/lib> }:
```
Mic92, line 1:
```suggestion
{ lib ? import <nixpkgs/lib> }:
```

yajo: I think you don't really need importing to be able to use the lib.

Lassulus:
generally ok with it, I would do this after either #22 or #18 are merged. we need to rebase then so I will probably do it directly on master

Lassulus:
done in https://github.com/nix-community/disko/commit/6b0b20da18cdffd09f04faee7128c557bcb9f054

PR #25, opened 2022-08-24
Set part-type as optional when creating a partition

Baitinq:
The man page specifies that part-type may be specified only with msdos and  dvh  partition  tables, hence it shouldnt be used with gpt partitions. This change makes it able to do this by making this attribute optional


Mic92 (approved):

Mic92:
We could also simplify our examples than...

Lassulus:
Will merge this after my mega PR is done, maybe I will also just absorb it into there

Baitinq:
@Lassulus Ill also implement partition labels support when ur pr is done:)

Mic92:
> @Lassulus Ill also implement partition labels support when ur pr is done:)

This is now the case.

Baitinq:
Ill rebase this against master as soon as #30 is merged.

Mic92:
It does not seem to be an optional parameter: https://www.gnu.org/software/parted/manual/html_chapter/parted_2.html#SEC18. We could potentiality have a default value though.

default.nix
```diff
@@ -107,7 +107,7 @@ let
   create.partition = q: x: let
     env = helper.device-id q.device;
   in ''
-    parted -s "''${${env}}" mkpart ${x.part-type} ${x.fs-type or ""} ${x.start} ${x.end}
+    parted -s "''${${env}}" mkpart ${x.part-type or ""} ${x.fs-type or ""} ${x.start} ${x.end}
```
Mic92, line 110:
```suggestion
    parted -s "''${${env}}" mkpart ${x.part-type or ""} ${x.fs-type or "primary"} ${x.start} ${x.end}
```

Baitinq:
@Lassulus I've fixed the conflict with the master branch, should I squash the commits or can you do that while merging?

default.nix
Baitinq: From the arch wiki (https://wiki.archlinux.org/title/Parted), part-type is only needed when using MBR.

Baitinq: @Mic92 I believe the man page you have linked is outdated. This is the current parted man page which specifies the optionality of part-type (https://www.gnu.org/software/parted/manual/parted.html#mkpart)

Lassulus:
this was absorbed into the types rewrite. thanks!

PR #26, opened 2022-08-25
mdadm: also mount containing filesystem

Mic92:


Mic92:
I think there should be a test for this...

PR #27, opened 2022-08-25
zfs, lvm raid, btrfs subvolumes support & some fixups

Lassulus:


default.nix
```diff
@@ -180,6 +207,31 @@ let
...
+    recursiveUpdate
+    (foldl' recursiveUpdate {} (map (mount-f (q // { pool = q.name; })) x.datasets))
+    {zpool.${q.device} = ''
+      zpool list '${q.name}' >/dev/null 2>/dev/null || zpool import '${q.name}'
+    '';}
+  );
+
+  mount.zfs_filesystem = q: x: {
+      zfs.${x.mountpoint} = ''
+        if ! grep -qs '${q.pool}/${x.name} /mnt${x.mountpoint}' /proc/mounts; then
```
Mic92, line 223: findmnt does not work here?

Lassulus: it did not in my tests

Lassulus: ok, now it works

tests/zfs.nix
```diff
@@ -0,0 +1,40 @@
...
+    { config, pkgs, modulesPath, ... }:
+
+    {
+      imports = [
+        (modulesPath + "/profiles/installation-device.nix")
+        (modulesPath + "/profiles/base.nix")
+      ];
+
+      # speed-up eval
+      documentation.enable = false;
```
Mic92, line 26: This can go to a separate module that is included from every test.

tests/luks-lvm.nix
```diff
@@ -0,0 +1,39 @@
...
+}:
+let
+  makeTest' = args:
+    makeTest args {
+      inherit pkgs;
+      inherit (pkgs) system;
+    };
+  disko-config = import ../example/luks-lvm.nix;
+  tsp-create = pkgs.writeScript "create" ((pkgs.callPackage ../. {}).create disko-config);
+  tsp-mount = pkgs.writeScript "mount" ((pkgs.callPackage ../. {}).mount disko-config);
```
Mic92, line 12: Could this be a general test function that we re-ruse?
```
makeDiskoTest = diskoconfig: func: let
     tsp-create = pkgs.writeScript "create" ((pkgs.callPackage ../. {}).create disko-config);
     tsp-mount = pkgs.writeScript "mount" ((pkgs.callPackage ../. {}).mount disko-config);
     args = func { inherit tsp-create, tsp-mount; };
in
     makeTest args {
      inherit pkgs;
      inherit (pkgs) system;
    };
```

default.nix
```diff
@@ -125,6 +125,31 @@ let
...
+      -V ${x.size} \
+      ${lib.optionalString (isAttrs x.options or null) (concatStringsSep " " (mapAttrsToList (n: v: "-o ${n}=${v}") x.options))}
+    udevadm trigger --subsystem-match=block; udevadm settle
+    ${create-f { device = "/dev/zvol/${q.pool}/${x.name}"; } x.content}
+  '';
+
+  create.zpool = q: x: ''
+    zpool create ${q.name} \
+      ${lib.optionalString (!isNull x.mode || x.mode != "stripe") x.mode} \
+      ${lib.optionalString (isAttrs x.options or null) (concatStringsSep " " (mapAttrsToList (n: v: "-o ${n}=${v}") x.options))} \
```
0x4A6F, line 148: Is there a way to set filesystem options with -O on the root dataset?

tests/luks-lvm.nix
Lassulus: we probably need to define the number of disks too, but I guess we can generalize that

default.nix
Lassulus: do we need that? what would be the difference to just -o on the zpool?

tests/luks-lvm.nix
Lassulus: actually we could get the number of disks from the diskoConfig

default.nix
0x4A6F: `zpool create` will create the root zfs dataset, without -O these dataset properties can only be set with `zfs set` afterwards. `zpool create -o` does only set pool properties and not dataset properties.

default.nix
```diff
@@ -52,10 +52,10 @@ let
     boot.initrd.luks.devices.${x.name}.device = q.device;
   } // config-f { device = "/dev/mapper/${x.name}"; } x.content;
 
-  config.lv = q: x:
+  config.lvm_lv = q: x:
     config-f { device = "/dev/mapper/${q.vgname}-${q.name}"; } x.content;
```
ajs124, line 56: escaped ihr das richtig, wenn der name `-` enthält?

default.nix
```diff
@@ -98,15 +98,23 @@ let
...
+    pvcreate ${q.device}
+    LVMDEVICES_${x.vg}="''${LVMDEVICES_${x.vg}:-}${q.device} "
+  '';
+
+  create.lvm_lv = q: x: ''
+    lvcreate \
+      ${if hasInfix "%" x.size then "-l" else "-L"} ${x.size} \
+      -n ${q.name} \
+      ${lib.optionalString (!isNull x.lvm_type or null) "--type=${x.lvm_type}"} \
+      ${q.vgname}
```
ajs124, line 111: gibt es da keine extraargs oder sowas? was wenn ich `--stripes` oder sowas angeben will?

ajs124: einen thinpool und thin lvs darauf kann ich hiermit relativ sicher auch nicht erstellen :thinking:

default.nix
```diff
@@ -38,42 +39,65 @@ let
     };
   };
 
+  config.zfs_filesystem = q: x: {
```
mweinelt, line 42: q? x?

Lassulus: q is the state upwards and x the configuration downwards

Issue #29, opened 2022-08-26
Label disks?


yajo:
[In the installation instructions](https://nixos.org/manual/nixos/stable/#sec-installation-installing), it seems recommended to identify disks by label (`/dev/disk/by-label/nixos` and friends).

It would be very nice if disko could add labels to disks when creating the filesystems.

Baitinq:
I'll implement this as soon as #27 is merged

Mic92:
I think #27 also add create options for mkfs to make this feasible.

Baitinq:
@Mic92 I dont think you can use mkfs to label a partition, but I might be wrong on that.

Mic92:
Yeah I also think it needs to be done per filesystem: 

Fileystem | package | set label | show label
-- | -- | -- | --
ext2/ext3/ext4 | e2fsprogs | sudo e2label /dev/sda1 NEUES_LABEL sudo tune2fs -L NEUES_LABEL /dev/sda1 | sudo e2label /dev/sda1  sudo tune2fs -l /dev/sda1
ReiserFS | reiserfsprogs | sudo reiserfstune -l NEUES_LABEL /dev/sda1 | sudo debugreiserfs
JFS | jfsutils | sudo jfs_tune -L NEUES_LABEL /dev/sda1 | sudo jfs_tune -l
XFS | xfsprogs | xfs_admin -L NEUES_LABEL /dev/sda1 | xfs_admin -l
FAT | dosfstools | sudo fatlabel /dev/sdb1 NEUES_LABEL | sudo fatlabel /dev/sdb1
FAT | mtools | sudo mlabel -i /dev/sdb1 ::NEUES_LABEL | sudo mlabel -i /dev/sdb1 -s ::
FAT | mtools und Skript (siehe unten) | sudo ./set_msdos_label.sh /dev/sdb1 NEUES_LABEL |  
NTFS | ntfs-3g (ab Ubuntu 11.10) | sudo ntfslabel /dev/sdb1 NEUES_LABEL | sudo ntfslabel /dev/sdb1
Swap | util-linux | sudo swaplabel -L NEUES_LABEL /dev/sda1 | sudo swaplabel /dev/sda1

yajo:
According to https://github.com/nix-community/disko/pull/30/files#r956295248, a workaround for disk labeling is to use a GPT table and set the label as `part-type`.

Lassulus:
This should be done in #31

PR #30, opened 2022-08-26
Add partition labels support

Baitinq:


Baitinq:
Fixes  #29

Lassulus:
huh does is need to be positional or can we parametrize the argument?

default.nix
```diff
@@ -142,7 +142,7 @@ let
       env = helper.device-id q.device;
     in
     ''
-      parted -s "''${${env}}" mkpart ${x.part-type} ${x.fs-type or ""} ${x.start} ${x.end}
+      parted -s "''${${env}}" mkpart ${x.part-type} ${x.part-label} ${x.fs-type or ""} ${x.start} ${x.end}
```
Mic92, line 145: Are you sure this is correct?

```
  mkpart PART-TYPE [FS-TYPE] START END     make a partition
```

I don't see any label support in parted for this.

Mic92: https://www.gnu.org/software/parted/manual/html_chapter/parted_2.html#SEC18

Baitinq: @Mic92 Im seeing conflicting info. On the arch wiki for parted (https://wiki.archlinux.org/title/Parted) it says:

```
(parted) mkpart part-type-or-part-label fs-type start end

    part-type-or-part-label is interpreted differently based on the partition table:
        MBR: the parameter is interpreted as part-type, which can be one of primary, extended or logical.
        GPT: the parameter is interpreted as part-label, which sets the [PARTLABEL](https://wiki.archlinux.org/title/PARTLABEL) attribute of the partition. The partition label always has to be set, since mkpart does not allow to create partitions with empty label.
```

Mic92: Ok. In this case I would pass down the format type in `create.table` down to `create.partition`:

```
    ${concatStrings (imap (index: create-f (q // { inherit index format; })) x.partitions)}
```

And than only pass either `x.part-type` or `x.part-label` to parted based on the format type (gpt or mbr).
Otherwise disko might pass too many arguments to mkpart

yajo: [The man page](https://www.man7.org/linux/man-pages/man8/parted.8.html) is interesting:

              mkpart [part-type name fs-type] start end
                     Create a new partition. part-type may be specified
                     only with msdos and dvh partition tables, it should
                     be one of "primary", "logical", or "extended".
                     name is required for GPT partition tables and fs-
                     type is optional.  fs-type can be one of "btrfs",
                     "ext2", "ext3", "ext4", "fat16", "fat32", "hfs",
                     "hfs+", "linux-swap", "ntfs", "reiserfs", "udf", or
                     "xfs".

              name partition name
                     Set the name of partition to name. This option
                     works only on Mac, PC98, and GPT disklabels. The
                     name can be placed in double quotes, if necessary.
                     And depending on the shell may need to also be
                     wrapped in single quotes so that the shell doesn't
                     strip off the double quotes.

It seems to indicate the same as https://github.com/nix-community/disko/pull/30/files#r956295248.

Maybe a good option is to assert that you're getting either a part-type for msdos and dvh tables, or a part-label for gpt tables.

Lassulus:
this should now be possible already with the version on master

PR #31, opened 2022-08-29
Reimplement using types

Lassulus:
started working on types, sadly I have to change the current format since it caused conflicts at top-level (although I could probably revert that with the new subType option type, but the new format should also be clearer for a constraint solver I wanted to work on afterwards).
I will probably implement most of the stuff from the current default.nix as type options to make the code (hopefully) a bit clearer.


types.nix
```diff
@@ -0,0 +1,278 @@
...
+      };
+      options = mkOption {
+        type = types.listOf types.str;
+        default = [];
+      };
+      mountpoint = mkOption {
+        type = types.str; # TODO absolute-path
+      };
+      format = mkOption {
+        type = types.enum [ "ext1" "ext2" "ext3" "ext4" "btrfs" "xfs" ]; # TODO add many more
```
Mic92, line 57: Not sure if this should be type checked at all. It's kind of open-ended and potentially user extensible:

```suggestion
        type = types.enum [ "ext1" "ext2" "ext3" "ext4" "btrfs" "xfs" "vfat" "f2fs" ]; # TODO add many more
```

And this is still missing bcachefs, jfs and others...

Lassulus: hmm yeah we can leave that probably open and just let the script fail if the binary is missing

Lassulus:
nearly done, just need some more documentation, code cleanup and some optionTypes for devices and names

PR #33, opened 2022-08-31
fix(btrfs): subvolume options support in mount and config

yajo:
- Add `config.btrfs`.
- Fix `mount.filesystem` when extra `mountOptions` were used (they were split by space instead of comma).
- Support btrfs subvolumes and mount options in `mount.btrfs`.
- Support the case where a subvolume is named different than its mountpoint.

BREAKING CHANGE: This changes the btrfs layout for subvolumes from a list of strings to a set. This is needed to let the set include other options.

@moduon MT-1075

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -84,7 +84,7 @@ let
     let
       subvolumeNames = attrNames x.subvolumes;
     in ''
-    mkfs.btrfs ${q.device}
+    mkfs.btrfs ${q.device} ${x.extraArgs or ""}
     ${lib.optionalString (!isNull x.subvolumes or null) ''
       MNTPOINT=$(mktemp -d)
       (
--- a/default.nix
+++ b/default.nix
@@ -46,6 +46,16 @@ let
     };
   };
 
+  config.btrfs = q: x: {
+    fileSystems = mapAttrs' (name: value:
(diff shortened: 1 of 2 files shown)


Lassulus:
ah nice, I just implemented it very minimally since I don't use it and didn't know more about it.

Sadly you probably have to redo this after https://github.com/nix-community/disko/pull/31 is finished, which is becoming a complete rewrite of disko now. Hopefully I'm finished till the end of the week.

yajo:
There's https://nixos.wiki/wiki/Btrfs for the docs. This patch should be complete as is, can't you merge it before #31? 😅

Lassulus:
ah sure I could merge it before, but it would be gone after I merge #31 since it replaces the default.nix 

I would leave this open after #31 got merged as a reminder. ofc you could also update the code then to make it compatible.

yajo:
I won't follow up on this PR because I'm switching over to XFS due to performance reasons. It shouldn't be so hard for somebody with knowledge on nixos type system to follow up on these changes.

Sadly, without this, btrfs is not really well supported.

Thanks anyways!

Lassulus:
I will take a look at a future point. just busy right now with other stuff (like better tests)

PR #43, opened 2022-09-30
add nixos tests for disko.config

Lassulus:


example/zfs-over-legacy.nix
```diff
@@ -34,11 +49,16 @@
...
       datasets = {
-        zfs_fs = {
+        "root" = {
+          zfs_type = "filesystem";
+          options.mountpoint = "none";
+        };
+        "root/zfs_fs" = {
           zfs_type = "filesystem";
           mountpoint = "/zfs_fs";
-          options."com.sun:auto-snapshot" = "true";
```
Mic92, line 41: You don't want to keep autosnapshots here for testing zfs options?

Lassulus: was removed for debugging, will return soon (is already in my dirt)

PR #47, opened 2022-10-07
Allow setting device to a variable

pinpox:


types.nix
```diff
@@ -63,6 +64,7 @@ rec {
           if match "/dev/disk/.*" dev != null then "dev_disk" else
           if match "/dev/nvme.*" dev != null then "dev_nvme" else
           if match "/dev/md/.*" dev != null then "dev_md" else
+          if match ".*" dev != null then "dev_var" else
```
pinpox, line 67: This regex should be more specific

Lassulus:
this is sadly not possible. since we handle different device types (like nvme, mdadm, disk/by-id, disk/by-uuid) and all of them have different suffixes for partitions. we can't have this detection in shell, since we also need these device names for the nixos config.

So the device type needs to be determined in nix and we can't delay it until the shell execution phase.

PR #48, opened 2022-10-23
add a NixOS module for easier usage

Lassulus:

PR #49, opened 2022-11-05
feature: add flake shorthand (hoist from lib)

blaggacao:
it's just more ergonomic for such an atomic functionality

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -3,10 +3,12 @@
 
   inputs.nixpkgs.url = "github:NixOS/nixpkgs/nixos-unstable";
 
-  outputs = { self, nixpkgs, ... }: {
+  outputs = { self, nixpkgs, ... }: let
     lib = import ./. {
       inherit (nixpkgs) lib;
     };
+  in {
+    inherit lib;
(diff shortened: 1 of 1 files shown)


blaggacao:
it's on the table, just seeing that after https://github.com/nix-community/disko/pull/48  the namespace might not be as minimal anymore as it is today...

flake.nix
```diff
@@ -16,5 +18,7 @@
         makeTest = import (pkgs.path + "/nixos/tests/make-test-python.nix");
         eval-config = import (pkgs.path + "/nixos/lib/eval-config.nix");
       };
-  };
+  }
+  // lib # shorthand
+  ;
```
Lassulus, line 23: I think this is a bit to much. but we probably also want the stuff from ./default.nix in here?

blaggacao: Yeah, I agree. After `module` is merged, the panorama looks a bit different.

I'll close and eventually re-open, in the mesure ot makes sense.

blaggacao:
I may reopen later after reassessing the situation now that `module` branch is merged.

PR #64, opened 2022-11-27
add disko output variant without package list

Mic92:
for netboot or kexec based installer we may want to avoid uploading too many dependencies because the target could go out-of-memory.

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -51,6 +51,11 @@ in {
       ${types.diskoLib.zapCreateMount cfg.devices}
     '';
 
+    # This is useful to skip copying executables uploading a script to an in-memory installer
+    system.build.diskoNoDeps = pkgs.writers.writeBash "disko" ''
+      ${types.diskoLib.zapCreateMount cfg.devices}
+    '';
+
     # Remember to add config keys here if they are added to types
     fileSystems = lib.mkIf cfg.enableConfig (lib.mkMerge (lib.catAttrs "fileSystems" (types.diskoLib.config cfg.devices)));
     boot = lib.mkIf cfg.enableConfig (lib.mkMerge (lib.catAttrs "boot" (types.diskoLib.config cfg.devices)));


Lassulus:
but we need the dependencies to run the script? we also just get the ones we need, so its as small as possible

Mic92:
> but we need the dependencies to run the script? we also just get the ones we need, so its as small as possible

Are they not already included in the nixos installer? At least this is what I did before with my script.

Mic92:
I guess I can also do `${types.diskoLib.zapCreateMount cfg.devices}` in my own nixos configuration otherwise.

Lassulus:
> > but we need the dependencies to run the script? we also just get the ones we need, so its as small as possible
> 
> Are they not already included in the nixos installer? At least this is what I did before with my script.

not all of them, I guess zfs is not always there? not sure for the other tools. Also if they are already there, there should be no overhead?

Mic92:
> not all of them, I guess zfs is not always there? not sure for the other tools. Also if they are already there, there should be no overhead?

Ok. In my project I don't need that. I will just switch to my own disko output.

Lassulus:
but zfs is only added to the packages list if you define a zfs volume

Mic92:
```
# foo.nix
 with import <nixpkgs> {};
 
 pkgs.writeText "test" ''
   ${lib.makeBinPath [ dosfstools e2fsprogs parted btrfs-progs ]}
 ''
```

```
$ nix-build foo.nix
$ nix path-info -Sh ./result
/nix/store/yyh8bpl1m5qw8w5cz0xkpa1rhkq41l0y-test         192.6M
```

^ And now you need 200MB more RAM. I already had people running into the 2.5GB RAM limit.

Lassulus:
but those are already part of the installation image? or not?

Mic92:
> but those are already part of the installation image? or not?

No because your installer image might be on a different nixpkgs revision than your nixos installation. Which is not super uncommon.

Lassulus:
hmm, well I guess we can have it. but the remote installer should use the generated packages to be feature complete. I would prefer to have a more minimal installer image and get just the deps we need then. (maybe not-os?)

module.nix
```diff
@@ -51,6 +51,11 @@ in {
       ${types.diskoLib.zapCreateMount cfg.devices}
     '';
 
+    # This is useful to skip copying executables uploading a script to an in-memory installer
+    system.build.diskoNoDeps = pkgs.writers.writeBash "disko" ''
+      ${types.diskoLib.zapCreateMount cfg.devices}
+    '';
```
Lassulus, line 57:
```suggestion
    system.build.diskoNoDeps = pkgs.writers.writeBash "disko" (
      types.diskoLib.zapCreateMount cfg.devices
    );
```

Lassulus: ah but I guess its more in style this way

Mic92:
not-os might help but it is also not very well maintained and polished in other areas.

zimbatm:
there is also https://github.com/nix-community/NixNG but it would be nice if NixOS itself could be trimmed down to the max

Mic92:
I felt like it was mainly optimized for containers. Can it boot on bare-metal?

Mic92:
> NixNG cannot boot on real hardware due to lacking a kernel and initramfs, but it can "boot" as a container, be it LXC or OCI. Some modules are implemented but there is still a lot to be done. If you want to get a feel for it, you can look into ./examples where you will find fully functional systems, mostly catered to be containers.

zimbatm:
yeah, marked off-topic

PR #67, opened 2022-11-30
Fix syntaax in cli.nix

mweinelt:


Mic92:
fixed in https://github.com/nix-community/disko/pull/65

PR #68, opened 2022-12-01
add first flake template

Mic92:


templates/hybrid-lvm/disko-config.nix
```diff
@@ -0,0 +1,75 @@
...
+          start = "0";
+          end = "1M";
+          part-type = "primary";
+          flags = ["bios_grub"];
+        }
+        {
+          type = "partition";
+          name = "ESP";
+          start = "1MiB";
+          end = "100MiB";
```
Mic92, line 22:
```suggestion
          end = "500MiB";
```

templates/hybrid-lvm/disko-config.nix
```diff
@@ -0,0 +1,75 @@
...
+          bootable = true;
+          content = {
+            type = "mdraid";
+            name = "boot";
+          };
+        }
+        {
+          name = "root";
+          type = "partition";
+          start = "100MiB";
```
Mic92, line 32:
```suggestion
          start = "500MiB";
```

templates/hybrid-lvm/disko-config.nix
Lassulus: yeah why not. or should we put just the ESP on the partiton and /boot to /?

templates/hybrid-lvm/disko-config.nix
```diff
@@ -0,0 +1,75 @@
+# Example to create a bios compatible gpt partition
+{ lib, disks ? [ "/dev/sda" ], ... }: {
```
zimbatm, line 2: Move the disks into an option?

PR #70, opened 2022-12-03
add tmpfs support

Lassulus:
we could also have tmpfs as toplevel instead of nodev. but I guess we also want to have unionfs and bindfs or something at some point? maybe even nfs?

### DIFF
--- a/types.nix
+++ b/types.nix
@@ -834,6 +834,7 @@ rec {
       };
       type = mkOption {
         type = types.enum [ "zpool" ];
+        default = "zpool";
         internal = true;
       };
       mode = mkOption {
@@ -1212,6 +1213,8 @@ rec {
       };
       type = mkOption {
         type = types.enum [ "disk" ];
+        default = "disk";
+        internal = true;
       };
       device = mkOption {
         type = optionTypes.absolute-pathname; # TODO check if subpath of /dev ? - No! eg: /.swapfile
--- a/types.nix
+++ b/types.nix
@@ -286,7 +286,7 @@ rec {
           fs.${config.mountpoint} = ''
             if ! findmnt ${dev} "/mnt${config.mountpoint}" > /dev/null 2>&1; then
               mount ${dev} "/mnt${config.mountpoint}" \
-              ${concatStringsSep " " config.mountOptions} \
+              ${concatMapStringsSep " " (opt: "-o ${opt}") config.mountOptions} \
               -o X-mount.mkdir
             fi
           '';
@@ -360,7 +360,7 @@ rec {
           fs.${config.mountpoint} = ''
             if ! findmnt ${dev} "/mnt${config.mountpoint}" > /dev/null 2>&1; then
               mount ${dev} "/mnt${config.mountpoint}" \
-              ${toString config.mountOptions} \
+              ${concatMapStringsSep " " (opt: "-o ${opt}") config.mountOptions} \
               -o X-mount.mkdir
             fi
           '';
@@ -895,7 +895,7 @@ rec {
               if ! findmnt ${config.name} "/mnt${config.mountpoint}" > /dev/null 2>&1; then
                 mount ${config.name} "/mnt${config.mountpoint}" \
                 ${optionalString ((config.options.mountpoint or "") != "legacy") "-o zfsutil"} \
-                ${toString config.mountOptions} \
+                ${concatMapStringsSep " " (opt: "-o ${opt}") config.mountOptions} \
                 -o X-mount.mkdir \
                 -t zfs
               fi
@@ -993,7 +993,7 @@ rec {
               if ! findmnt ${zpool}/${config.name} "/mnt${config.mountpoint}" > /dev/null 2>&1; then
                 mount ${zpool}/${config.name} "/mnt${config.mountpoint}" \
                 -o X-mount.mkdir \
-                ${toString config.mountOptions} \
+                ${concatMapStringsSep " " (opt: "-o ${opt}") config.mountOptions} \
                 ${optionalString ((config.options.mountpoint or "") != "legacy") "-o zfsutil"} \
                 -t zfs
               fi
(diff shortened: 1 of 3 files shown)


Mic92:
yes I think we could cover all device less filesystems in one go here. nfs, all the different fuse, tmpfs, ramfs, 9p...

Lassulus:
Puh get me config examples. Then I'm gonna implement them

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+        readOnly = true;
+        type = diskoLib.jsonType;
+        default = {
+          fs.${config.mountpoint} = ''
+            if ! findmnt tmpfs "/mnt${config.mountpoint}" > /dev/null 2>&1; then
+              mount -t tmpfs none "/mnt${config.mountpoint}" \
+              ${lib.optionalString (!isNull config.size) "-o size=${config.size}"} \
+              ${concatStringsSep " " config.mountOptions} \
+              -o X-mount.mkdir
+            fi
```
Mic92, line 288: I think all it would take would be to not have a `size` mountoptions and add an fs type instead of hardcoding `tmpfs`.

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+            options = config.mountOptions ++
+              lib.optional (!isNull config.size) "size=${config.size}";
+          };
+        }];
+      };
+      _pkgs= mkOption {
+        internal = true;
+        readOnly = true;
+        type = types.functionTo (types.listOf types.package);
+        default = pkgs: [];
```
Mic92, line 308: Also to pull packages it could make sense to make additional filesystem types.

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+        type = types.attrsOf (diskoLib.subType { inherit tmpfs; });
+        default = {};
+      };
     };
   };
 
+  tmpfs = types.submodule ({ config, ... }: {
+    options = {
+      type = mkOption {
+        type = types.enum [ "tmpfs" ];
```
Mic92, line 249: This might work:

```suggestion
        type = types.enum [ "tmpfs" "nfs" ];
```

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+        default = "";
+      };
+      _mount = mkOption {
+        internal = true;
+        readOnly = true;
+        type = diskoLib.jsonType;
+        default = {
+          fs.${config.mountpoint} = ''
+            if ! findmnt tmpfs "/mnt${config.mountpoint}" > /dev/null 2>&1; then
+              mount -t tmpfs none "/mnt${config.mountpoint}" \
```
Mic92, line 284:
```suggestion
              mount -t ${config.type} none "/mnt${config.mountpoint}" \
```

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+              ${lib.optionalString (!isNull config.size) "-o size=${config.size}"} \
```
Mic92, line 285:
```suggestion
```

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+  tmpfs = types.submodule ({ config, ... }: {
+    options = {
+      type = mkOption {
+        type = types.enum [ "tmpfs" ];
+        internal = true;
+      };
+      size = mkOption {
+        type = types.nullOr types.str; # TODO size type
+        default = null;
+      };
```
Mic92, line 255:
```suggestion
```

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+          '';
+        };
+      };
+      _config = mkOption {
+        internal = true;
+        readOnly = true;
+        default = [{
+          fileSystems.${config.mountpoint} = {
+            device = "none";
+            fsType = "tmpfs";
```
Mic92, line 298:
```suggestion
            fsType = config.type;
```

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
+            options = config.mountOptions ++
+              lib.optional (!isNull config.size) "size=${config.size}";
```
Mic92, line 300:
```suggestion
            options = config.mountOptions;
```

types.nix
```diff
@@ -235,9 +236,80 @@ rec {
...
         default = {};
       };
+      nodev = mkOption {
+        type = types.attrsOf (diskoLib.subType { inherit tmpfs; });
+        default = {};
+      };
     };
   };
 
+  tmpfs = types.submodule ({ config, ... }: {
```
Mic92, line 246: not sure here:

```suggestion
  nodevFS = types.submodule ({ config, ... }: {
```

types.nix
Lassulus: hmm, for example for nfs. the none part needs to be the remote. maybe we need an option for the device? I have the feeling, that dedicated types could be better for UX

types.nix
Lassulus: what mounts would need extra packages?

Lassulus:
alright, I made the nodev type now more general and it supports multiple different fsTypes (I didn't want to reuse type, since we usually use that for type detection and not for specification, and this would have created more confusion in this special case)

types.nix
```diff
@@ -235,9 +236,82 @@ rec {
...
 
+  nodev = types.submodule ({ config, ... }: {
+    options = {
+      type = mkOption {
+        type = types.enum [ "nodev" ];
+        default = "nodev";
+        internal = true;
+      };
+      fsType = mkOption {
+        type = types.enum [ "tmpfs" "ramfs" "nfs" ];
```
Lassulus, line 254: do we want a str here? there are probably a lot of different fsTypes and we don't want to take care of all of them here. but for some we probably would need some extra packages.

Lassulus:
is nodev the right name if they can have a device? :D

types.nix
Mic92: A string sounds reasonable.

types.nix
Mic92: Ecryptfs, davfs, cifs... But we can also leave this empty for now.

Mic92:
> is nodev the right name if they can have a device? :D

Right, I just cannot think of one.

PR #72, opened 2022-12-05
Allow shell variables in disks parameter 

phaer:
...which are expanded at run-time.

While this limits the declarative nature of disko a bit, i believe its useful as it allows pre-generating the scripts for systems which don't run nix (yet) while still keeping the flexibility of declaring disks to use at-runtime.

Generate the script: 
```
disko-create-zfs = self.lib.createScriptNoDeps (import ./example/zfs.nix { disks = ["\${disk1}" "\${disk2}" ]; }) pkgs;
```
(one could use positional parameters like `${1}` if preferred)

Call it at runtime:

```
disk1=/dev/vda disk2=/dev/vdb bash disko-create-zfs-simple
```

This seems to work for me and was easy to implement. What do you think?

### DIFF
--- a/types.nix
+++ b/types.nix
@@ -57,6 +57,8 @@ rec {
         "${dev}-part${toString index}" # /dev/disk/by-id/xxx style
       else if match "/dev/(nvme|md/|mmcblk).+" dev != null then
         "${dev}p${toString index}" # /dev/nvme0n1p1 style
+      else if match "\\\$\\{.+}" dev != null then # shell variable: ${primaryDisk}
+        "${dev}${toString index}"
       else
         abort "${dev} seems not to be a supported disk format";
 
@@ -214,6 +216,13 @@ rec {
           isString x && (x == "/" || (length xs > 0 && all filename.check xs));
       merge = mergeOneOption;
     };
+
+    shell-variable = mkOptionType {
+      name = "Shell variable";
+      check = x: match "\\\$\\{.+}" x != null;
+      merge = mergeOneOption;
+    };
+
   };
 
   /* topLevel type of the disko config, takes attrsets of disks, mdadms, zpools, nodevs, and lvm vgs.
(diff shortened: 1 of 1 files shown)


Lassulus:
If we really want to have shell injection in the script, we lose the nixos config part and should disable it completely in that case. I guess the best way to do that is to use `builtins.throw` in _config if a `shell-variable` is used.

types.nix
```diff
@@ -57,6 +57,8 @@ rec {
         "${dev}-part${toString index}" # /dev/disk/by-id/xxx style
       else if match "/dev/(nvme|md/|mmcblk).+" dev != null then
         "${dev}p${toString index}" # /dev/nvme0n1p1 style
+      else if match "\\\$\\{.+}" dev != null then # shell variable: ${primaryDisk}
+        "${dev}${toString index}"
```
Lassulus, line 61: well this also just works if you give it a "/dev/xda" style device. maybe we need some more logic here if people want to use this with nvmes etc.

phaer: Oh right! Not entirely sure how to solve that tbh, as we can't know which scheme to use during eval time. 

The only idea I've got for the moment would be to move device numbering to run-time (at least in this case) but that seems somewhat more involved. :/ Happy for any input here :)

phaer:
> If we really want to have shell injection in the script, we lose the nixos config part and should disable it completely in that case. I guess the best way to do that is to use builtins.throw in _config if a shell-variable is used.

Right! I believe this should be fixed with my latest commit?

types.nix
Lassulus: after reading the old code, it actually did somewhat different, but we can reuse that here also. It tried to find the corresponding device in either /dev/disk/by-id or /dev/disk/by-path and continued to work with that (since the device numbering is stable with those).
https://github.com/nix-community/disko/blob/9bb4aec9640cbc30e241c267158e506278862b5e/default.nix#L7

phaer:
Just converted this to a draft. This branch is currently in use in https://github.com/dep-sys/nix-dabei/tree/disko-runtime, but we'll see how it works out after that one supports a few more use cases, like mirrored zfs disks. 

As it stands it still could turn out to be a bad idea to do this with disko as it currently exists.

PR #73, opened 2022-12-08
fix: remove existing zfs pool

jfroche:


Lassulus:
maybe we can check if zpool is in PATH (like with mdadm) but maybe thats also a bad heuristic

Lassulus:
I should probably extend PATH instead of overriding it at the top, hmm

types.nix
```diff
@@ -164,6 +164,10 @@ rec {
         umount -Rlv /mnt
       fi
 
+      for ds in $(zpool list -H -o name); do
+        zpool destroy "$ds"
+      done
```
Mic92, line 169:
```suggestion
      if command -v zpool; then
        for ds in $(zpool list -H -o name); do
          zpool destroy "$ds"
        done
      fi
```

types.nix
```diff
@@ -164,6 +164,10 @@ rec {
...
+
       # stop all existing raids
       if command -v mdadm; then
```
Mic92, line 172:
```suggestion
      if command -v mdadm &>/dev/null; then
```

types.nix
zimbatm: `s/command -v/type -p/g` because it's shorter :-p

Lassulus:
I was inspired by this PR to do this: https://github.com/nix-community/disko/pull/75

jfroche:
ok closing this one. I tested #75 and added https://github.com/nix-community/disko/pull/79 to make it work for me

PR #78, opened 2022-12-13
enable nix run

archite:
Is there a reason why something like this is not done so that `nix run .#nixosConfigurations.myhostname.config.system.build.diskoNoDeps` could setup disks prior to `nixos-install --flake .#myhostname`? This saves running build and then the result script.

Great work. Thanks for your effort.

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -20,23 +20,23 @@ in {
     };
   };
   config = lib.mkIf (cfg.devices.disk != {}) {
-    system.build.formatScript = pkgs.writers.writeBash "disko-create" ''
+    system.build.formatScript = pkgs.writers.writeBashBin "disko-create" ''
       export PATH=${lib.makeBinPath (types.diskoLib.packages cfg.devices pkgs)}:$PATH
       ${types.diskoLib.create cfg.devices}
     '';
 
-    system.build.mountScript = pkgs.writers.writeBash "disko-mount" ''
+    system.build.mountScript = pkgs.writers.writeBashBin "disko-mount" ''
       export PATH=${lib.makeBinPath (types.diskoLib.packages cfg.devices pkgs)}:$PATH
       ${types.diskoLib.mount cfg.devices}
     '';
 
-    system.build.disko = pkgs.writers.writeBash "disko" ''
+    system.build.disko = pkgs.writers.writeBashBin "disko" ''
       export PATH=${lib.makeBinPath (types.diskoLib.packages cfg.devices pkgs)}:$PATH
       ${types.diskoLib.zapCreateMount cfg.devices}
     '';
 
     # This is useful to skip copying executables uploading a script to an in-memory installer
-    system.build.diskoNoDeps = pkgs.writeScript "disko" ''
+    system.build.diskoNoDeps = pkgs.writeScriptBin "disko" ''
       #!/usr/bin/env bash
       ${types.diskoLib.zapCreateMount cfg.devices}
     '';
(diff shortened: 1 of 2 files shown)


Mic92:
Currently was mainly used by the disko cli (which might be also need to be fixed as a result of this)

Mic92:
This line needs to be fixed as well: https://github.com/nix-community/disko/blob/935feb7496cbf96119f1eeb951f258c1550b3273/disko#L124

archite:
Looking at it, it only looks like this change would affect the nixosModule. I've been using this in my flake outputs as a workaround but not ideal.

```nix
apps = forAllSystems (system: let
  pkgs = nixpkgs.legacyPackages.${system};
  hosts = pkgs.lib.filterAttrs (_: value:
    value.pkgs.system == system &&
    builtins.hasAttr "diskoNoDeps" value.config.system.build
  ) self.nixosConfigurations;
in
  if (hosts == {}) then {} else {
  disko = pkgs.lib.genAttrs (builtins.attrNames hosts) (name:
    {
      program = "${self.nixosConfigurations.${name}.config.system.build.diskoNoDeps}";
      type = "app";
    });
  }
);
```

iFreilicht:
This change makes disko incompatible with [nixos-anywhere, which calls the script directly](https://github.com/nix-community/nixos-anywhere/blob/d5ced5dc30c7a14587f002a1ab67f5201911f03d/src/nixos-anywhere.sh#L469).

That's a solvable problem, but I don't see much use in changing this now. These scripts are not supposed to be called directly by a user anyway. Unless there's a good reason to get this change merged, I'd vote for closing this PR.

Enzime:
As someone who doesn't use the disko CLI, I personally would like the ability to run `nix run github:Enzime/dotfiles-nix#nixosConfigurations.host.config.system.build.formatScript`, this would make it super easy to execute from a NixOS live installer for example

iFreilicht:
Okay good to know, thank you. I'm concerned about backwards compatibility, what could we do so we don't break compatibility with nixos-anywhere? Maybe we should submit a PR to them that detects if `formatScript` is a file or folder and then executes either `formatScript` or `formatScript/bin/format`?

Enzime:
I think the actual root cause is that `nix run` should support executing executable files, rather than forcing all of them to follow the structure `bin/<meta.mainProgram>`

But that would probably be the easiest solution

cc @Mic92 with his nixos-anywhere maintainer hat on

iFreilicht:
How easy is it right now to get PRs merged in NixCpp? Haven't followed the development for a while, but this would be relatively simple to implement. Just add some additional logic here: https://github.com/NixOS/nix/blob/68ba6ff4709d936c1a714de35da08f8ed354c710/src/nix/app.cc#L118

module.nix
```diff
@@ -20,23 +20,23 @@ in {
     };
   };
   config = lib.mkIf (cfg.devices.disk != {}) {
-    system.build.formatScript = pkgs.writers.writeBash "disko-create" ''
+    system.build.formatScript = pkgs.writers.writeBashBin "disko-create" ''
```
Mic92, line 23: If we switch to this, we should make new attributes and make the old "script" ones execute those with a warning using `lib.warn`. This would give us time to migrate tools like nixos-anywhere.

Mic92: Also `diskoNoDeps` is the only thing used by nixos-anywhere, we still have nixpkgs with the old disko cli around.

Mic92: Maybe something like:

- `diskoFormat`, `diskoMount`, `diskoRecreate`, `diskoRecreateNoDeps`

Enzime:
@iFreilicht I think a simple PR for `nix run` could get through pretty quickly if you update docs and tests

iFreilicht:
This PR would also fix the underlying issue in #454, which then only becomes a documentation issue.

iFreilicht:
Closing in favor of #842

PR #75, opened 2022-12-10
zap only devices we care about

Lassulus:

PR #79, opened 2022-12-13
Add lvm2 package to use `blkdeactivate`

jfroche:
Zap script uses `blkdeactivate` to deactivates block devices. `blkdeactivate` is currently in the lvm2 package.
Add grep/find/cut tools as well as we are fixing complete PATH.

### DIFF
--- a/types.nix
+++ b/types.nix
@@ -605,8 +605,9 @@ rec {
         internal = true;
         readOnly = true;
         type = types.functionTo (types.listOf types.package);
+        # lvm2 package is required hereafter to use `blkdeactivate` in `zapCreateMount` script
         default = pkgs:
-          [ pkgs.parted pkgs.systemdMinimal ] ++ flatten (map (partition: partition._pkgs pkgs) config.partitions);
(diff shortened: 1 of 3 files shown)


types.nix
```diff
@@ -606,7 +606,7 @@ rec {
         readOnly = true;
         type = types.functionTo (types.listOf types.package);
         default = pkgs:
-          [ pkgs.parted pkgs.systemdMinimal ] ++ flatten (map (partition: partition._pkgs pkgs) config.partitions);
+          [ pkgs.parted pkgs.systemdMinimal pkgs.lvm2 pkgs.gnugrep pkgs.coreutils pkgs.findutils ] ++ flatten (map (partition: partition._pkgs pkgs) config.partitions);
```
Mic92, line 609: Could be lvm2 moved to lvm partitions?

jfroche: We need `blkdeactivate` from lvm2 package even if we don't use lvm

Mic92: I see. Maybe add a comment why this is needed.

Mic92:
Will this be deprecated by https://github.com/nix-community/disko/pull/75 ?

PR #81, opened 2022-12-15
module: use `bin` variants to work with `nix run`

Enzime:
This allows commands like:

```
nix run github:Enzime/dotfiles-nix#nixosConfigurations.sigma-nixos.config.system.build.mountScript --override-input disko github:Enzime/disko/nix-run
```

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -18,23 +18,23 @@ in {
     };
   };
   config = {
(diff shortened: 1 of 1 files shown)


Enzime:
Duplicate of #78

PR #83, opened 2022-12-20
swap: fix typo for checking if swap is on

lilyinstarlight:
Simple little error shellcheck happened to find. If you're trying to get the exit status of `grep` in that command pipe, I believe a command group is intended, not command substitution

### DIFF
--- a/types.nix
+++ b/types.nix
@@ -655,7 +655,7 @@ rec {
         type = types.functionTo diskoLib.jsonType;
         default = dev: {
           fs.${dev} = ''
-            if ! $(swapon --show | grep -q '^${dev} '); then
+            if ! swapon --show | grep -q '^${dev} '; then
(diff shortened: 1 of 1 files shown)


types.nix
```diff
@@ -655,7 +655,7 @@ rec {
         type = types.functionTo diskoLib.jsonType;
         default = dev: {
           fs.${dev} = ''
-            if ! $(swapon --show | grep -q '^${dev} '); then
+            if ! (swapon --show | grep -q '^${dev} '); then
```
Mic92, line 658: Should this not also work?

```suggestion
            if ! swapon --show | grep -q '^${dev} '; then
```

lilyinstarlight: Ah yeah, so it does! I'll push an update

PR #84, opened 2022-12-20
types: pass through mountOptions to _config for all types

lilyinstarlight:
So the `filesystem.options` option never seemed to actually be used anywhere in disko (for creating, mounting, or configuring) and it seems to have been confused with the `mountOptions` option, which is used but often only during `_mount` and not `_config` (except for `nodev` where it was passed through correctly)

I may be off-base with that observation (still trying to fully figure disko out tbh) but if that observation is correct, this PR removes the `filesystems.options` type and passes through `filesystems.mountOptions` through to the filesystem definitions output by `_config` so that they get propagated to the system config as well

I also had to update the tests which used `filesystem.options` (which did not seem to do anything) in the place of using `filesystem.mountOptions` to specify mount options

I ran the tests to confirm this worked (and I had btrfs pass them through in #82 like this PR has), but let me know if anything in this change seems problematic or if I'm not understanding things correctly

### DIFF
--- a/example/complex.nix
+++ b/example/complex.nix
@@ -109,7 +109,7 @@
             type = "filesystem";
             format = "ext4";
             mountpoint = "/ext4_on_lvm";
-            options = [
+            mountOptions = [
               "defaults"
             ];
           };
--- a/example/luks-lvm.nix
+++ b/example/luks-lvm.nix
@@ -17,7 +17,7 @@
               type = "filesystem";
               format = "vfat";
               mountpoint = "/boot";
-              options = [
+              mountOptions = [
                 "defaults"
               ];
             };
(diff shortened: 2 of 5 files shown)


Lassulus:
this seems correct. I was unsure when initially implementing this, if the same options are supported by nixos as with `-o $option` but I guess they are. if we encounter problems later we can still fix/revert.

thanks!

lilyinstarlight:
> I was unsure when initially implementing this, if the same options are supported by nixos as with `-o $option` but I guess they are.

Yep, NixOS puts the options in the fstab after escaping and separating by commas, so it should be the exact same options. It defaults to `["defaults"]` in NixOS so I set our default value to that too

Also because setting `fileSystems.<name>.options` to an empty list results in an invalid fstab (which should probably be fixed in NixOS so I'll probably make a PR to nixpkgs later to either assert or use `defaults` in fstab when given an empty list of options)

PR #87, opened 2022-12-21
disko: add missing btrfs-progs

Mic92:


lilyinstarlight:
Oh yeah I noticed these were missing (and just giving errors when formatting using the `btrfs` type) and actually rolled effectively this same change into #82 as well when I opened it

I'll rebase that one if this is merged first, though, since that one may be a bit before it's merged

Mic92:
I also saw that you add those. I was just helping out a friend to format his drive until we get your thing reviewed.

Mic92:
No longer needed since we have the other PR merged.

PR #89, opened 2022-12-23
WIP: fix ci

Lassulus:


flake.nix
```diff
@@ -24,6 +24,7 @@
     });
     # TODO: disable bios-related tests on aarch64...
     checks = forAllSystems (system: let
+      nixpkgs = builtins.getFlake "nixpkgs/nixos-unstable";
```
Mic92, line 27: I think this is impure and not allowed. I think we need a second nixpkgs that is locked in a lockfile...

PR #95, opened 2022-12-28
pre/post create hooks

phaer:
Not sure about the best implementation here yet.

I'd like to use disk encryption with ZFS and existing disko facilities work fine to set it up, but we obviously can't use keylocation=prompt during disko formatting. So unlocking fails on boot if the file referred to by keylocation isn't included in the initrd.

Ideally I could use keylocation=file://tmp/disk.key during partitioning, but set keylocation=prompt right after that.

What do you think? Maybe a generic option to add post-creation shell commands would be a more flexible solution?

example usage:
```
    rpool = {
      type = "zpool";
      mode = "";
      options = {
        ashift = "12";
        autotrim = "on";
      };
      optionsAfterCreate = {
        keylocation = "prompt";
      };
      rootFsOptions = {
        encryption = "on";
        keylocation = "file:///tmp/disk.key";
        keyformat = "passphrase";
        compression = "zstd";
        acltype = "posixacl";
        mountpoint = "none";
        canmount = "off";
        xattr = "sa";
        dnodesize = "auto";
        normalization = "formD";
        relatime = "on";
      };
```

### DIFF
--- a/example/config.nix
+++ b/example/config.nix
@@ -5,7 +5,7 @@
       type = "lvm_vg";
       lvs = {
         root = {
-          type = "lv";
+          type = "lvm_lv";
           size = "10G";
           content = {
             type = "filesystem";
@@ -14,7 +14,7 @@
           };
         };
         home = {
-          type = "lv";
+          type = "lvm_lv";
           size = "10G";
           content = {
             type = "filesystem";
@@ -33,9 +33,8 @@
         format = "gpt";
         partitions = [
           {
-            name = "boot";
+            name = "ESP";
             type = "partition";
-            part-type = "ESP";
             start = "1MiB";
             end = "1024MiB";
             fs-type = "fat32";
--- a/example/stand-alone/configuration.nix
+++ b/example/stand-alone/configuration.nix
@@ -18,7 +18,7 @@
     disk = {
       sda = {
         device = "/dev/sda";
-        type = "device";
+        type = "disk";
         content = {
           type = "table";
           format = "msdos";
--- a/module.nix
+++ b/module.nix
@@ -20,12 +20,12 @@ in {
     };
   };
   config = lib.mkIf (cfg.devices.disk != {}) {
-    system.build.formatScript = pkgs.writers.writeDash "disko-create" ''
+    system.build.formatScript = pkgs.writers.writeBash "disko-create" ''
       export PATH=${lib.makeBinPath (types.diskoLib.packages cfg.devices pkgs)}:$PATH
       ${types.diskoLib.create cfg.devices}
     '';
 
-    system.build.mountScript = pkgs.writers.writeDash "disko-mount" ''
+    system.build.mountScript = pkgs.writers.writeBash "disko-mount" ''
       export PATH=${lib.makeBinPath (types.diskoLib.packages cfg.devices pkgs)}:$PATH
       ${types.diskoLib.mount cfg.devices}
     '';
--- a/types.nix
+++ b/types.nix
@@ -1159,7 +1159,9 @@ rec {
             ${concatStringsSep " " (mapAttrsToList (n: v: "-o ${n}=${v}") config.options)} \
             ${concatStringsSep " " (mapAttrsToList (n: v: "-O ${n}=${v}") config.rootFsOptions)} \
             ''${ZFSDEVICES_${config.name}}
-          zfs set ${concatStringsSep " " (mapAttrsToList (n: v: "${n}=${v}") config.optionsAfterCreate)} ${config.name}
+          ${lib.optionalString
+            (config.optionsAfterCreate != {})
+            "zfs set ${concatStringsSep " " (mapAttrsToList (n: v: "${n}=${v}") config.optionsAfterCreate)} ${config.name}"}
           ${concatMapStrings (dataset: dataset._create config.name) (attrValues config.datasets)}
         '';
         description = "Creation script";
--- a/types.nix
+++ b/types.nix
@@ -149,7 +149,7 @@ rec {
       };
     };
 
-    mkCreateOption = { config, options, default }:
+    mkCreateOption = { config, options, default }@attrs:
       mkOption {
         internal = true;
         readOnly = true;
@@ -159,11 +159,20 @@ rec {
             "(" # subshell for namespacing
             (diskoLib.defineHookVariables { inherit config options; })
             config.preCreateHook
-            (default args)
+            (attrs.default args)
             config.postCreateHook
             ")"
           ];
         description = "Creation script";
+      };
+
+    mkMountOption = { config, options, default }@attrs:
+      mkOption {
+        internal = true;
+        readOnly = true;
+        type = types.functionTo diskoLib.jsonType;
+        default = args: let v = attrs.default args; in lib.traceSeqN 3 [(config.name or "unnamed") (v.fs or "")] v;
+        description = "Mount script";
       };
 
 
@@ -196,8 +205,6 @@ rec {
       ${concatMapStrings (dev: ((attrByPath (dev ++ [ "_mount" ]) "" devices) {}).dev ) sortedDeviceList}
 
       # and then mount the filesystems in alphabetical order
-      # attrValues returns values sorted by name.  This is important, because it
-      # ensures that "/" is processed before "/foo" etc.
       ${concatStrings (attrValues fsMounts)}
     '';
 
@@ -334,10 +341,8 @@ rec {
         inherit config options;
         default = {}: "";
       };
-      _mount = mkOption {
-        internal = true;
-        readOnly = true;
-        type = types.functionTo diskoLib.jsonType;
+      _mount = diskoLib.mkMountOption {
+        inherit config options;
         default = {}: {
           fs.${config.mountpoint} = ''
             if ! findmnt ${config.fsType} "/mnt${config.mountpoint}" > /dev/null 2>&1; then
@@ -347,7 +352,6 @@ rec {
             fi
           '';
         };
-        description = "Mount script";
       };
       _config = mkOption {
         internal = true;
@@ -413,10 +417,8 @@ rec {
           ${concatMapStrings (subvol: subvol._create { inherit dev; }) (attrValues config.subvolumes)}
         '';
       };
-      _mount = mkOption {
-        internal = true;
-        readOnly = true;
-        type = types.functionTo diskoLib.jsonType;
+      _mount = diskoLib.mkMountOption {
+        inherit config options;
         default = {dev}:
           let
             subvolMounts = diskoLib.deepMergeMap (subvol: subvol._mount {inherit dev; parent = config.mountpoint;}) (attrValues config.subvolumes);
@@ -431,7 +433,6 @@ rec {
               '';
             };
           };
-        description = "Mount script";
       };
       _config = mkOption {
         internal = true;
@@ -506,10 +507,8 @@ rec {
           )
         '';
       };
-      _mount = mkOption {
-        internal = true;
-        readOnly = true;
-        type = types.functionTo diskoLib.jsonType;
+      _mount = diskoLib.mkMountOption {
+        inherit config options;
         default = {dev, parent}: let
           mountpoint = if (!isNull config.mountpoint) then config.mountpoint
                        else if (isNull parent) then config.name
@@ -523,7 +522,6 @@ rec {
             fi
           '';
         };
-        description = "Mount script";
       };
       _config = mkOption {
         internal = true;
@@ -592,10 +590,8 @@ rec {
             ${dev}
         '';
       };
-      _mount = mkOption {
-        internal = true;
-        readOnly = true;
-        type = types.functionTo diskoLib.jsonType;
+      _mount = diskoLib.mkMountOption {
+        inherit config options;
         default = {dev}: {
           fs.${config.mountpoint} = ''
             if ! findmnt ${dev} "/mnt${config.mountpoint}" > /dev/null 2>&1; then
@@ -606,7 +602,6 @@ rec {
             fi
           '';
         };
-        description = "Mount script";
       };
       _config = mkOption {
         internal = true;
@@ -673,10 +668,8 @@ rec {
           ${concatMapStrings (partition: partition._create {inherit dev; type = config.format;} ) config.partitions}
         '';
       };
-      _mount = mkOption {
-        internal = true;
-        readOnly = true;
-        type = types.functionTo diskoLib.jsonType;
+      _mount = diskoLib.mkMountOption {
+        inherit config options;
         default = {dev}:
           let
             partMounts = diskoLib.deepMergeMap (partition: partition._mount {inherit dev;}) config.partitions;
@@ -686,7 +679,6 @@ rec {
             '';
             fs = partMounts.fs or {};
         };
-        description = "Mount script";
       };
       _config = mkOption {
         internal = true;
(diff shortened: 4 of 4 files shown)


Lassulus:
I actually use keylocation=prompt? it works, but there is no prompt :D

phaer:
> I actually use keylocation=prompt? it works, but there is no prompt :D

Not sure I understand, where does it get the key from then? :D

Lassulus:
you can just write it into the terminal, it doesn't even get hidden

phaer:
Ah, but that would require an interactive installation (or piping the encryption key, i guess), right? I'd actually  prefer to pass the key as a file - as with diskos LUKS support - during installation. I only need `prompt` afterwards, during normal boot.

Lassulus:
hmm, I would prefer something more generic, maybe a post/pre action per type?

phaer:
To recap a chat with @Lassulus today:

* Are there good use cases for generic pre/post create actions for create & mount? Code-size should be a consideration.
* just using strings for actions would require hardcoding things like the device or ${config.name}. 
* using functions for actions would make it impossible to serialize the whole config (e.g. to json).

## possible use case

ZFS allows adding (faster) SSDs as a second-level cache L2ARC with a syntax like this:

```
zpool add rpool cache /dev/sdc
```

That's not supported with the current implementation, but could be in a post-create-actions. 
There seems to be another form of zpool create which isn't yet supported by disko, but should be. It could cover the cache use case as well, but that should be tested ;)
```
sudo zpool create -n rpool -o ashift=12 \
    mirror /dev/sda \
    mirror /dev/sdb

```

I'll need to leave it at this for at least a few days.

versedwildcat:
Another possible use case for a after-creation, before-mount hook would be to create an empty snapshot on ZFS or BTRFS, which gets rolled back on boot
To use with [impermanence](https://github.com/nix-community/impermanence)
+1 on this feature

jackinloadup:
Seems like before/after scripts would cover most use-cases. Particularly popular ones could become options as demonstrated in this PR with  its 'prompt after creation' functionality.

phaer:
@versedwildcat @jackinloadup I think at this point it's mostly about proposing an implementation which stays serializable while still supporting as many use cases as possible (.i.e. substitution of the device name and ideally other options in a hook). 

I think it should be doable with `lib.toShellVars` and will try that "soon-ish".

phaer:
Quick example on how that could work on a zpool to reference in discussion:

We define a helper to serialize the disko config of our `zpool` to shellVariables. I still need to generalize the filter to ignore things where the option has type `attrsOf submodule`. 
```
      defineHookVariables = config:
        lib.toShellVars
          (lib.filterAttrs
            (n: v: !(n == "datasets" || lib.hasPrefix "_" n || lib.hasSuffix "Hook" n))
            config);
```

If we then define a hookOption with an example as default
```
      options.postCreateHook = mkOption {
        type = types.str;
        default = "zfs set keylocation=prompt $name";
      };
```

And just quickly add it to zpools `_create` default value: 
```
          ${defineHookVariables config}
          ${config.postCreateHook}
```

We end up with a config which is still serializable to JSON but allows to keep hooks dynamic.
The changes above executed with `./example/zfs.nix` result in the following `_create_output` (truncated):

```
[...]
mode='mirror'
declare -a mountOptions=('defaults')
mountpoint='/'
name='zroot'
declare -A options=()
declare -A rootFsOptions=(['com.sun:auto-snapshot']='false' ['compression']='lz4')
type='zpool'
zfs set keylocation=prompt $name
[...]
```

We would still need to generalize that by finishing the filter to work with all of diskos submodules and than wrap their _create and _mount functions with `defineHookVariables`, `preHook`, `postHook`, etc.
And how to make sure we clear hook variables after each submodule (i.e. between two zfs datasets; subshells? functions & local?).

jackinloadup:
@phaer Thanks for the context. Keeping serializable makes sense.

> And how to make sure we clear hook variables after each submodule (i.e. between two zfs datasets; subshells? functions & local?)

I'm not familiar with the dash shell being used in the create and mount scripts at the moment. If it's like bash then you can scope variables in functions  to 'clear' the variables. 

```
#/usr/bin/env bash

# Define a global variable
greeting="Hello, global world!"

function modify_global {
  # Modify the global variable
  greeting="Hello, modified global world!"
}

function modify_local {
  # Declare a local copy of the global variable
  local greeting="Hello, local world!"
  # Modify the local copy of the global variable
  greeting="Hello, modified local world!"
}

echo $greeting # Hello, global world!

modify_global
echo $greeting # Hello, modified global world!

modify_local
echo $greeting # Hello, modified global world!
```

If I have some time I might poke it but thank you for the work you have done thus far!

phaer:
Still a work in progress. Create scripts look okay, but are mostly untested atm. mountScripts are broken because I need to adapt the hook wrapper to its recursive application style.
 I am considering splitting "device activation" out of `_mount` into an `_activate` attribute to keep it a bit simpler, but we'll see.

The idea is that each submodule gets a mixin which defines for 4 string options for the hooks: {pre,post}{Create,Mount}.
`_mount` and `_create` are wrapped by `diskoLib.mkMountOption` and `diskoLib.mkCreateOption`, respectively.

Those wrappers wrap `default` to open a subshell, define hookVariables and call the hooks. This currently means that i.e. `postCreate` for a zpool is currently called *after* all of its children/datasets have been created and that nested subshells are used. 

<details>
<summary>
example: create script of examples/zfs.nix, after shfmt.  
</summary>

```sh
#!/usr/bin/env bash
set -efux
( # disk x /dev/sdx
    device='/dev/sdx'
    name='x'
    type='disk'
    echo 'DEBUG preCreate: disk'
    ( # table   gpt
        format='gpt'
        type='table'
        echo 'DEBUG preCreate: table'
        parted -s /dev/sdx -- mklabel gpt
        ( # partition ESP
            bootable='1'
            end='64MiB'
            declare -a flags=()
            fs_type='fat32'
            index='1'
            name='ESP'
            part_type='primary'
            start='0'
            type='partition'
            echo 'DEBUG preCreate: partition'
            parted -s /dev/sdx -- mkpart ESP fat32 0 64MiB

            # ensure /dev/disk/by-path/..-partN exists before continuing
            udevadm settle
            parted -s /dev/sdx -- set 1 boot on

            # ensure further operations can detect new partitions
            udevadm settle
            ( # filesystem   vfat /boot
                extraArgs=''
                format='vfat'
                declare -a mountOptions=('defaults')
                mountpoint='/boot'
                type='filesystem'
                echo 'DEBUG preCreate: filesystem'
                mkfs.vfat \
                    \
                    /dev/sdx1

                echo 'DEBUG postCreate: filesystem'
            )

            echo 'DEBUG postCreate: partition'
        )
        ( # partition zfs
            bootable=''
            end='100%'
            declare -a flags=()
            fs_type=''
            index='2'
            name='zfs'
            part_type='primary'
            start='128MiB'
            type='partition'
            echo 'DEBUG preCreate: partition'
            parted -s /dev/sdx -- mkpart zfs 128MiB 100%

            # ensure /dev/disk/by-path/..-partN exists before continuing
            udevadm settle

            # ensure further operations can detect new partitions
            udevadm settle
            ( # zfs
                pool='zroot'
                type='zfs'
                echo 'DEBUG preCreate: zfs'
                ZFSDEVICES_zroot="${ZFSDEVICES_zroot:-}/dev/sdx2 "

                echo 'DEBUG postCreate: zfs'
            )

            echo 'DEBUG postCreate: partition'
        )

        echo 'DEBUG postCreate: table'
    )

    echo 'DEBUG postCreate: disk'
)
( # disk y /dev/sdy
    device='/dev/sdy'
    name='y'
    type='disk'
    echo 'DEBUG preCreate: disk'
    ( # table   gpt
        format='gpt'
        type='table'
        echo 'DEBUG preCreate: table'
        parted -s /dev/sdy -- mklabel gpt
        ( # partition zfs
            bootable=''
            end='100%'
            declare -a flags=()
            fs_type=''
            index='1'
            name='zfs'
            part_type='primary'
            start='128MiB'
            type='partition'
            echo 'DEBUG preCreate: partition'
            parted -s /dev/sdy -- mkpart zfs 128MiB 100%

            # ensure /dev/disk/by-path/..-partN exists before continuing
            udevadm settle

            # ensure further operations can detect new partitions
            udevadm settle
            ( # zfs
                pool='zroot'
                type='zfs'
                echo 'DEBUG preCreate: zfs'
                ZFSDEVICES_zroot="${ZFSDEVICES_zroot:-}/dev/sdy1 "

                echo 'DEBUG postCreate: zfs'
            )

            echo 'DEBUG postCreate: partition'
        )

        echo 'DEBUG postCreate: table'
    )

    echo 'DEBUG postCreate: disk'
)
( # zpool zroot   /
    mode='mirror'
    declare -a mountOptions=('defaults')
    mountpoint='/'
    name='zroot'
    declare -A options=()
    declare -A rootFsOptions=(['com.sun:auto-snapshot']='false' ['compression']='lz4')
    type='zpool'
    echo 'DEBUG preCreate: zpool'
    zpool create zroot \
        mirror \
        \
        -O com.sun:auto-snapshot=false -O compression=lz4 \
        ${ZFSDEVICES_zroot}
    ( # zfs_dataset encrypted
        declare -a mountOptions=('defaults')
        mountpoint=''
        name='encrypted'
        declare -A options=(['encryption']='aes-256-gcm' ['keyformat']='passphrase' ['keylocation']='file:///tmp/secret.key' ['mountpoint']='none')
        size='20M'
        type='zfs_dataset'
        zfs_type='filesystem'
        echo 'DEBUG preCreate: zfs_dataset'
        zfs create zroot/encrypted \
            -o encryption=aes-256-gcm -o keyformat=passphrase -o keylocation=file:///tmp/secret.key -o mountpoint=none

        echo 'DEBUG postCreate: zfs_dataset'
    )
    ( # zfs_dataset encrypted/test   /zfs_crypted
        declare -a mountOptions=('defaults')
        mountpoint='/zfs_crypted'
        name='encrypted/test'
        declare -A options=()
        size='2M'
        type='zfs_dataset'
        zfs_type='filesystem'
        echo 'DEBUG preCreate: zfs_dataset'
        zfs create zroot/encrypted/test

        echo 'DEBUG postCreate: zfs_dataset'
    )
    ( # zfs_dataset zfs_fs   /zfs_fs
        declare -a mountOptions=('defaults')
        mountpoint='/zfs_fs'
        name='zfs_fs'
        declare -A options=(['com.sun:auto-snapshot']='true')
        size=''
        type='zfs_dataset'
        zfs_type='filesystem'
        echo 'DEBUG preCreate: zfs_dataset'
        zfs create zroot/zfs_fs \
            -o com.sun:auto-snapshot=true

        echo 'DEBUG postCreate: zfs_dataset'
    )
    ( # zfs_dataset zfs_legacy_fs   /zfs_legacy_fs
        declare -a mountOptions=('defaults')
        mountpoint='/zfs_legacy_fs'
        name='zfs_legacy_fs'
        declare -A options=(['mountpoint']='legacy')
        size=''
        type='zfs_dataset'
        zfs_type='filesystem'
        echo 'DEBUG preCreate: zfs_dataset'
        zfs create zroot/zfs_legacy_fs \
            -o mountpoint=legacy

        echo 'DEBUG postCreate: zfs_dataset'
    )
    ( # zfs_dataset zfs_testvolume
        declare -a mountOptions=('defaults')
        mountpoint=''
        name='zfs_testvolume'
        declare -A options=()
        size='10M'
        type='zfs_dataset'
        zfs_type='volume'
        echo 'DEBUG preCreate: zfs_dataset'
        zfs create zroot/zfs_testvolume \
            \
            -V 10M
        udevadm settle
        ( # filesystem   ext4 /ext4onzfs
            extraArgs=''
            format='ext4'
            declare -a mountOptions=('defaults')
            mountpoint='/ext4onzfs'
            type='filesystem'
            echo 'DEBUG preCreate: filesystem'
            mkfs.ext4 \
                \
                /dev/zvol/zroot/zfs_testvolume

            echo 'DEBUG postCreate: filesystem'
        )

        echo 'DEBUG postCreate: zfs_dataset'
    )
    ( # zfs_dataset zfs_unmounted_fs
        declare -a mountOptions=('defaults')
        mountpoint=''
        name='zfs_unmounted_fs'
        declare -A options=(['mountpoint']='none')
        size=''
        type='zfs_dataset'
        zfs_type='filesystem'
        echo 'DEBUG preCreate: zfs_dataset'
        zfs create zroot/zfs_unmounted_fs \
            -o mountpoint=none

        echo 'DEBUG postCreate: zfs_dataset'
    )

    echo 'DEBUG postCreate: zpool'
)
```

</details>

Lassulus:
the generated shellscript look pretty nice. the mixins are kinda weird in the code, but I guess it's better than some alternatives. I will take a deeper look at some later point, but nice work so far!

phaer:
@Lassulus Thanks! I agree that code can be improved and I am not 100% sure yet how it will work with the mount logic, but I thought it's better to collect feedback on the principal approach before diving deeper :diving_mask: 

I also agree about the mixin. Maybe a `diskolib.subModule` similar to `diskolib.subType` would make sense? Or we could just replicate the hook declarations where needed.

> This currently means that i.e. postCreate for a zpool is currently called after all of its children/datasets have been created and that nested subshells are used.

And to elaborate a bit on this: this means it isn't really possible to run a hook e.g. in-between the creation of a zpool and the creation of its children atm. I think this could and should be improved, but require `_create` to return an attrset instead of a string, just like `_mount`, to get i.e. create commands for the zpool and its children split in the hook wrapper.

Thanks for your input so far, looking forward to further feedback!

phaer:
Ah, it turns out the zfs create script above can't work like this because it tries to append to `ZFSDEVICES_zroot` inside a subshell. That can't work, so I suppose we would need to use functions and local variables for namespacing :/

Lassulus:
maybe we need a tmpdir, but maybe that makes the script less portable

phaer:
Okay, semantics around when to run the hooks exactly and discussed code re-structuring are still open.

But we are now keeping device lists for zfs, lvm and mdadm in persistent files between subshells and all tests except the modules one are passing. Haven't looked into that last one too much yet and also haven't tested this branch in practice yet ;)

phaer:
Just deployed a real hcloud machine with this branch and hooks for both, my encryption prompt, and a blank snapshot for things like impermanence like @versedwildcat and @jackinloadup have asked for. I've used the following config for my zpool.

Note that i can already use `$name` in the zpool hook, and could to so for the "local/root" part in the dataset hook, but need to hardcode the pool name atm. this can be fixed later on.

<details>
<summary>zpool config with hooks</summary>

```nix
  zpool = {
    rpool = {
      type = "zpool";
      mode = "";
      options = {
        ashift = "12";
        autotrim = "on";
      };
      postCreateHook = ''
        zfs set keylocation="prompt" $name;
      '';
      rootFsOptions = {
        encryption = "on";
        keylocation = "file:///tmp/disk.key";
        keyformat = "passphrase";
        compression = "zstd";
        acltype = "posixacl";
        mountpoint = "none";
        canmount = "off";
        xattr = "sa";
        dnodesize = "auto";
        normalization = "formD";
        relatime = "on";
      };
      datasets = let
        unmountable = {
          zfs_type = "filesystem";
          mountpoint = null;
          options.canmount = "off";
        };
        filesystem = mountpoint: {
          zfs_type = "filesystem";
          inherit mountpoint;
        };
      in {
        "local" = unmountable;
        "safe" = unmountable;
        "local/root" = filesystem "/" // {
          postCreateHook = "zfs snapshot rpool/local/root@blank";
          options.mountpoint = "legacy";
        };
        "local/nix" = filesystem "/nix" // {options.mountpoint = "legacy";};
        "safe/home" = filesystem "/home";
      };
    };
  };
```

</details>

jackinloadup:
@phaer I'll play with it and get back. Looking forward to using it!

phaer:
@jackinloadup Glad it worked for you! Let's see what @Lassulus has to say about the code :)

Lassulus:
looks good, the code got a bit more complex, but I guess that is the price for more features.

I guess we could make the code a bit clearer with something like `mkDiskoType`, which we would then use to define the different types like disk, lvm, partition, ...

But this has not to happen here, We could merge this and refactor it later.

phaer:
@Lassulus Nice to read, thanks! I agree about `mkDiskoType` but would prefer to do that in another PR as this one is already big enough IMO.

So if okay, I'd like to merge this if tests pass; just rebased against master and removed debug output of the hooks.

PR #97, opened 2022-12-29
Add disks with bootable partition to boot.grub.devices

phaer:
We could also set it to "nodev" if we have an ESP partition but no bios_grub partition, but i am yet unsure whether that's a robust enough heuristic?

### DIFF
--- a/tests/gpt-bios-compat.nix
+++ b/tests/gpt-bios-compat.nix
@@ -7,5 +7,4 @@ makeDiskoTest {
     machine.succeed("mountpoint /");
   '';
   efi = false;
-  grub-devices = [ "/dev/vdb" ];
 }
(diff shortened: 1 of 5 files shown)


Lassulus:
I guess this should depend on the bios_grub partition, since the bootable flag has only an effect on dos partition tables afair

Lassulus:
we could also do this on dos partition tables in general. I'm not sure if they can even boot GPT? but there could be dos formatted disks which should not be bootable. not sure

phaer:
> I guess this should depend on the bios_grub partition, since the bootable flag has only an effect on dos partition tables afair

Oh you are fast! I had already force-pushed regarding the flag. And now adapted tests and some logic for nodev if there's no bios_grub partition.

Mic92:
What was the reason to close this?

phaer:
No real reason expect that it wasn't ready to merge yet and my priorities shifted to getting the hooks right first and potentially re-structure some of disko for that. Feel free to re-open/replace.

PR #106, opened 2023-01-10
feat: ShellCheck

l0b0:
Is this OK, @Mic92?

Ideally the scripts would have their respective `.sh` and `.bash` extensions, then we could `git ls-files '*.sh' '*.bash' | xargs shellcheck` instead.


.github/workflows/test.yml
```diff
@@ -0,0 +1,18 @@
...
+jobs:
+  lint:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3.2.0
+      - name: Install Nix
+        uses: cachix/install-nix-action@v18
+        with:
+          extra_nix_config: access-tokens = github.com=${{ secrets.GITHUB_TOKEN }}
```
l0b0, line 16: I don't think I need this, but I copied it from the other workflow since I'm not sure what it's being used for.

Mic92: This is needed because otherwise github actions often run into api request limits.

.github/workflows/test.yml
```diff
@@ -0,0 +1,18 @@
...
+      - name: ShellCheck
+        run: nix-shell --packages git shellcheck --pure --run 'shellcheck disk-deactivate/disk-deactivate disko'
```
Mic92, line 18: I would actually prefer this being a package in `checks` in our `flake.nix` so it runs also locally if you do `nix flake check`. This will be also likely faster to build in garnix than github actions.

Lassulus:
this does not shellcheck the generated code? would have been nice to check that also, since it's the part with the most error potential :)

Mic92:
@Lassulus I would avoid adding a runtime dependency on this, since haskell stuff might be quite heavy especially when you are cross compiling and also have to compile ghc. However one think we could do is shellchecking our tests for the same effect.

Mic92:
This seems like a bigger PR though, that we can do on top of this one.

lilyinstarlight:
Would we just have to add a few shellcheck steps to the testScript at https://github.com/nix-community/disko/blob/abcf1e18d5c98586204988c14fa99249db19583f/tests/lib.nix#L111 and make sure it's installed on the test VMs for that?

PR #108, opened 2023-01-11
wip: let's see what breaks with systemd.enable

Mic92:
don't merge


lilyinstarlight:
I know you're not looking to merge this, but as a quick note: systemd-initrd does not quite yet have correct initrd fs/disk dependencies, because systemd does not respect `/sysroot` in the fstab as needing the related root fs targets and nixos does not yet set `root=` on the kernel cmdline because it breaks other things (until the next systemd release where they have been fixed)

I don't know that the failures are really related to that, but it does make the occasional spooky problem with initrd ordering happen and so we should wait on at least NixOS/nixpkgs#208269 before turning it on for any real tests

(I'm sorry if you were already aware, since I know you're just trying to do a quick how-much-might-break test here and it's not the most relevant bit of info)

ElvishJerricco:
@lilyinstarlight while I broadly agree, I will also note that I don't see anything that *should* cause failures like we're seeing here. So we've definitely got some bugs to squash.

ElvishJerricco:
So, the output of the CI tests is a little unhelpful. In NixOS tests, the VM output was brought to quality parity with regular initrd in https://github.com/NixOS/nixpkgs/pull/207801. Should probably do similar here.

jackinloadup:
[NixOS/nixpkgs/pull/208269](https://github.com/NixOS/nixpkgs/pull/208269) was merged [](url)

lilyinstarlight:
> [NixOS/nixpkgs/pull/208269](https://github.com/NixOS/nixpkgs/pull/208269) was merged :-)

With minor modifications, it looks like most disko tests work with systemd-initrd, but the mdadm tests still do not because of NixOS/nixpkgs#210210. So we'll still need that fixed before we can do a test matrix or something with systemd-initrd both on and off for all tests. But we could enable it for a handful of tests now I suppose, if we really wanted to

Mic92:
rebased.

PR #111, opened 2023-01-12
allow referencing any path from /dev/disk/by-id

nazarewk:
for example: /dev/disk/by-id/usb-sage_3639S_2222222222222222005B-0:0

I'm setting up disko for the first time and encountered a validation error when trying to reference USB drives by their IDs

### DIFF
--- a/types.nix
+++ b/types.nix
@@ -205,8 +205,10 @@ rec {
           # trailing slashes.  It also removes leading slashes, thus we have to
           # check for "/" explicitly below.
           xs = filter (s: stringLength s > 0) (splitString "/" x);
+          # /dev/disk/by-id can have valid entries containing `:` for usb drives which do not pass filename validation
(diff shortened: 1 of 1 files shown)


Lassulus:
hmm, I guess we should edit the filename check to include more possible filenames (and remove the POSIX reference, since it's not correct anymore then)

nazarewk:
well in most other use cases it would indicate multi-value paths, this is why i stuck to whitelisting this prefix

Lassulus:
ah, but people can't mount their filesystems to /mnt/some:dir, we should allow that also

nazarewk:
sounds legit, will most likely continue on https://github.com/nix-community/disko/pull/112

Lassulus:
this is now included in nixpkgs via https://github.com/nix-community/disko/commit/6bbccd8edd06ff5c65e28c9be63f35224b7feaa3

PR #112, opened 2023-01-12
luks: add extraArgs{Format,Open}

nazarewk:
There are number of fixes I needed to make to support my use case, they're brewing up here.

### DIFF
--- a/types/luks.nix
+++ b/types/luks.nix
@@ -20,6 +20,16 @@
       default = [ ];
       description = "Extra arguments";
     };
+    extraArgsFormat = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = config.extraArgs;
+      description = "Extra arguments for `cryptsetup luksFormat`";
+    };
+    extraArgsOpen = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = [ ];
+      description = "Extra arguments for `cryptsetup luksOpen`";
+    };
     content = diskoLib.deviceType;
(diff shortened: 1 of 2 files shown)


types.nix
```diff
@@ -1,9 +1,9 @@
 { lib }:
 with lib;
 with builtins;
-
-rec {
-
+let
+  safeZpoolName = pool: builtins.replaceStrings [ "-" ] [ "_" ] pool;
+in rec {
```
Lassulus, line 6: hmm, I would prefer to abort on a invalid/unsafe zpool name rather than to replace it silently. replacing could lead to conflicts

nazarewk: it is for specific purpose of creating an environment variable name, could make it clearer

Lassulus: could they still conflict on similar zpool names? like my-pool and my_pool? maybe we don't care about that?

nazarewk: this should solve it once and for good https://github.com/nix-community/disko/pull/112/commits/9eeafb271583e4fbe7c05580cf9debcb7554aea0

Lassulus: I guess we use now tmpfiles for the pool names, not sure if this change is still relevant

nazarewk:
I've rebased the branch, but will test it and finish up the PR after setting up second laptop sometime next week

Lassulus:
I guess this is now partially obsolete due to https://github.com/nix-community/disko/pull/186

nazarewk:
this was addressed by #186 and #211

PR #103, opened 2023-01-03
feat: add the descriptions for the options

RossComputerGuy:
This PR fixes the missing descriptions in disko which is used by ExpidusOS by default. This will get rid of these errors.
```
options.json> error: option disko.devices has no description
options.json> error: option disko.devices.disk has no description
options.json> error: option disko.devices.disk.<name>.content has no description
options.json> error: option disko.devices.disk.<name>.device has no description
options.json> error: option disko.devices.disk.<name>.name has no description
options.json> error: option disko.devices.lvm_vg has no description
options.json> error: option disko.devices.lvm_vg.<name>.lvs has no description
options.json> error: option disko.devices.lvm_vg.<name>.lvs.<name>.content has no description
options.json> error: option disko.devices.lvm_vg.<name>.lvs.<name>.extraArgs has no description
options.json> error: option disko.devices.lvm_vg.<name>.lvs.<name>.lvm_type has no description
options.json> error: option disko.devices.lvm_vg.<name>.lvs.<name>.name has no description
options.json> error: option disko.devices.lvm_vg.<name>.lvs.<name>.size has no description
options.json> error: option disko.devices.lvm_vg.<name>.name has no description
options.json> error: option disko.devices.mdadm has no description
options.json> error: option disko.devices.mdadm.<name>.content has no description
options.json> error: option disko.devices.mdadm.<name>.level has no description
options.json> error: option disko.devices.mdadm.<name>.metadata has no description
options.json> error: option disko.devices.mdadm.<name>.name has no description
options.json> error: option disko.devices.nodev has no description
options.json> error: option disko.devices.nodev.<name>.device has no description
options.json> error: option disko.devices.nodev.<name>.fsType has no description
options.json> error: option disko.devices.nodev.<name>.mountOptions has no description
options.json> error: option disko.devices.nodev.<name>.mountpoint has no description
options.json> error: option disko.devices.zpool has no description
options.json> error: option disko.devices.zpool.<name>.datasets has no description
options.json> error: option disko.devices.zpool.<name>.datasets.<name>.content has no description
options.json> error: option disko.devices.zpool.<name>.datasets.<name>.mountOptions has no description
options.json> error: option disko.devices.zpool.<name>.datasets.<name>.mountpoint has no description
options.json> error: option disko.devices.zpool.<name>.datasets.<name>.name has no description
options.json> error: option disko.devices.zpool.<name>.datasets.<name>.options has no description
options.json> error: option disko.devices.zpool.<name>.datasets.<name>.size has no description
[...]

PR #115, opened 2023-01-15
flake: add documentation check

lilyinstarlight:
This doesn't generate a full manual, but it does ensure all options are correctly specified enough to be included in a manual

This would have caught the missing descriptions before #103 and implements the suggestion at https://github.com/nix-community/disko/pull/103#issuecomment-1370597259

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -28,17 +28,31 @@
     # Run checks: nix flake check -L
     checks = forAllSystems (system: let
       pkgs = nixpkgs.legacyPackages.${system};
+      lib = nixpkgs.lib;
       nixosTests = import ./tests {
         inherit pkgs;
         makeTest = import (pkgs.path + "/nixos/tests/make-test-python.nix");
         eval-config = import (pkgs.path + "/nixos/lib/eval-config.nix");
       };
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -28,17 +28,31 @@
     # Run checks: nix flake check -L
     checks = forAllSystems (system: let
       pkgs = nixpkgs.legacyPackages.${system};
+      lib = nixpkgs.lib;
       nixosTests = import ./tests {
         inherit pkgs;
         makeTest = import (pkgs.path + "/nixos/tests/make-test-python.nix");
         eval-config = import (pkgs.path + "/nixos/lib/eval-config.nix");
       };
+      documentation = (import (pkgs.path + "/nixos/doc/manual") rec {
```
Mic92, line 37: There is actual a proper api doc module documentation now. I had some wip branch here: https://github.com/nix-community/disko/pull/117
But also same issue because of those special subType the documentation cannot discover all options in disko.
To solve this issue one would need to add the equivalent of https://github.com/NixOS/nixpkgs/blob/master/lib/types.nix#L554-L555

lilyinstarlight: Yeah I'll close this one in favor of #117, since that one is clearly a better approach

PR #116, opened 2023-01-15
types,module,tests: run shellcheck on scripts before running them in NixOS tests

lilyinstarlight:
This is a draft because most generated scripts do not pass shellcheck and I want to make sure this approach is reasonable before fixing (or silencing) all of the little nits this finds

I had originally implemented it as a check run in the VM since that is a much simpler change, but I figured people may want to opt-in to having their scripts checked as they are generated (and also it is easier to get errors and fix that way when it's implemented as a checkPhase to the derivation rather than run in a VM way down the line)

(Also I may still change this approach -- I only wrote this in the last hour or two and tend to change my mind after a first implementation, but feedback is very welcome)

### DIFF
--- a/BREAKING_CHANGES.md
+++ b/BREAKING_CHANGES.md
@@ -1,0 +1,3 @@
+2023-02-14 6d630b8
+
+btrfs, btrfs_subvol filesystem and lvm_lv extraArgs are now lists
--- a/default.nix
+++ b/default.nix
@@ -1,5 +1,6 @@
 { lib ? import <nixpkgs/lib>
 , rootMountPoint ? "/mnt"
+, checked ? false
 }:
 let
   types = import ./types { inherit lib rootMountPoint; };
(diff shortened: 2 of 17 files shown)


lilyinstarlight:
Any thoughts on this @Lassulus and @Mic92?

I can start fixing (or silencing where appropriate) the reported shellcheck errors if both of you think this approach is reasonable for making it so that we have good shellcheck coverage even for generated scripts

Mic92:
Yes. We should have shellcheck in our CI.

Lassulus:
I will try to fix the remaining shellcheck errors and then we can merge this?

Lassulus:
we got a small backwards incompatiblity, but I wanted to get rid of it anyways: all extraArgs are lists now

lilyinstarlight:
> I will try to fix the remaining shellcheck errors and then we can merge this?

Awesome, thank you! :D

I'd intended to get to that at some point but this looks fabulous now

> we got a small backwards incompatiblity, but I wanted to get rid of it anyways: all extraArgs are lists now

That's probably for the best. Maybe we should start having a breaking changelog at some point?

Lassulus:
> That's probably for the best. Maybe we should start having a breaking changelog at some point?

sounds like a good idea, should we add a `BREAKING_CHANGES.md` to the root of the repo? or do you have another idea?

lilyinstarlight:
> sounds like a good idea, should we add a `BREAKING_CHANGES.md` to the root of the repo? or do you have another idea?

That sounds good enough to me. Or even just `CHANGES.md` would be fine, but I've no strong preference on name or method either way

Mic92:
I also think CHANGES.md would be fine but don't have releases yet for sectioning.

lilyinstarlight:
The breaking changes document looks fine enough for me, so is there anything else this PR is waiting on?

Mic92:
bors merge

PR #128, opened 2023-01-24
add support for another mountpoint than /mnt

Lassulus:


default.nix
```diff
@@ -1,6 +1,8 @@
-{ lib ? import <nixpkgs/lib> }:
+{ lib ? import <nixpkgs/lib>
+, rootMountPoint ? "/mnt"
```
Mic92, line 2: Should the cli not also expose this?

Lassulus: ah yeah, haven't used the cli in a while :D probably should also do some tests with it

Mic92:
bors merge

types.nix
```diff
@@ -223,7 +223,7 @@ rec {
     */
     zapCreateMount = devices: ''
       set -efux
-      umount -Rv /mnt || :
+      umount -Rv ${rootMountPoint} mnt || :
```
Mic92, line 226: Shoud this not be quoted?

Mic92, line 226:
```suggestion
      umount -Rv "${rootMountPoint}" mnt || :
```

Mic92:
bors merge

PR #131, opened 2023-01-28
Split disko type into multiple files

Mic92:


types/default.nix
```diff
@@ -0,0 +1,346 @@
...
+      nodev = lib.mkOption {
+        type = lib.types.attrsOf subTypes.nodev;
+        default = { };
+        description = "A non-block device";
+      };
+    };
+  };
+
+  subTypes = {
+    nodev = diskoLib.mkSubType [ ./nodev.nix ];
```
Lassulus, line 328: why does this take a list?

Mic92: modules

Mic92: https://github.com/nix-community/disko/pull/131/files#diff-63b2750ca88f0bb1b20a657859ddd5df357da7d87777bfd4054db88de1a6a598R148

Lassulus: but we don't need the list interface? it could be just a file?

Mic92: Done.

Lassulus (approved):

Lassulus:
bors merge

PR #143, opened 2023-02-06
Standardise formatting with nixpkgs-fmt and some patterns with statix 

cscutcher:
Perhaps a bit controversial, and a bit subjective, so please feel free to shoot down if it doesn't align with the maintainers point of view.

I'm a big fan of the [black] style of formatting; 

> _Black_ is the uncompromising Python code formatter. By using it, you agree to cede
> control over minutiae of hand-formatting. In return, _Black_ gives you speed,
> determinism, and freedom from `pycodestyle` nagging about formatting. You will save time
> and mental energy for more important matters.
> 
> Blackened code looks the same regardless of the project you're reading. Formatting
> becomes transparent after a while and you can focus on the content instead.
> 
> _Black_ makes code review faster by producing the smallest diffs possible.

`statix` goes a little further by normalising some standard patterns, but using it here with a similar philosophy, i.e. if two ways of doing things are, more or less the same, just pick which other one is most standard and leave me out of it.

There's nothing quite as universal as PEP8 or `black` in the Nix ecosystem (as far as I know anyway), so it doesn't quite provide the same level of "looks the same regardless of the project you're reading", but these both seem common enough to be part way there, and hopefully make it easier for potential contributors to read the existing code.

Even without the wider standardisation in the nix ecosystem, it still makes it easier to contribute to this project if one can just run a quick command against the code before submitting the PR and know that they're aligning with the conventions of the project. 

In this case I just ran;

```sh
nixpkgs-fmt **.nix && statix fix .
```

`nixpkgs-fmt` and `statix`  are my gotos just for [Ale](https://github.com/dense-analysis/ale)/vim integration, and due to the relative ease of autofixing on the command line, but I'd be happy with alternatives following the same philosophy. Or perhaps there's already a soft-convention in use in the repo that I've missed!

Anyway like I say, no biggie if it's an unpopular idea! I don't want to step on any toes, but just wanted to suggest it as someone thinking about contributing (if I find the time!). 

[black]: https://github.com/psf/black#the-uncompromising-code-formatter

### DIFF
--- a/doc.nix
+++ b/doc.nix
@@ -6,12 +6,12 @@ let
     rootMountPoint = "/mnt";
   };
   eval = lib.evalModules {
-    modules =  [
+    modules = [
       {
         options.disko = {
           devices = lib.mkOption {
             type = types.devices;
-            default = {};
+            default = { };
             description = "The devices to set up";
           };
         };
@@ -19,7 +19,7 @@ let
     ];
   };
   options = nixosOptionsDoc {
-    options = eval.options;
+    inherit (eval) options;
   };
   md = (runCommand "disko-options.md" { } ''
     cat >$out <<EOF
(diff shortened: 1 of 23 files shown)


doc.nix
```diff
@@ -6,20 +6,20 @@ let
...
+            default = { };
             description = "The devices to set up";
           };
         };
       }
     ];
   };
   options = nixosOptionsDoc {
-    options = eval.options;
+    inherit (eval) options;
```
Mic92, line 22: I think @Lassulus didn't liked those `inherit` constructs for readability reasons. If you could skip those in statix than it's good to merge.

Lassulus: We can merge them with the inherits and can try to configure the formater later

Mic92:
bors merge

cscutcher:
> We can merge them with the inherits and can try to configure the formater later

Awesome! Thanks for being flexible. I can be on the fence about the inherit fixes myself. Have added statix config, support for `nix fmt` and re-applied in draft PR #146. 

I wasn't sure if you guys would prefer a push to this branch given the bors appears to be struggling with the merge, or whether a follow up PR would be preferred.

Mic92:
Not so sure why the cli test was flaky.

PR #146, opened 2023-02-07
style: Add config for style normalisation and re-apply

cscutcher:
Following on from #143 

## Add config for `statix`

> We can merge them with the inherits and can try to configure the formater later

Awesome! Thanks for being flexible. I can be on the fence about the inherit fixes myself. Looks easy to disable, but it appears there are two variants to that check;

```
W03 manual_inherit
W04 manual_inherit_from
```

<details>
<summary>(Probably uninteresting detail on the difference here)</summary>

The former being cases like;

```
[W03] Warning: Assignment instead of inherit
    ╭─[<stdin>:18:3]
    │
 18 │   types = types;
    ·   ───────┬──────
    ·          ╰──────── This assignment is better written with inherit
────╯
```

the latter being;

```
[W04] Warning: Assignment instead of inherit from
    ╭─[./types/lvm_vg.nix:45:11]
    │
 45 │           fs = lvMounts.fs;
    ·           ────────┬────────
    ·                   ╰────────── This assignment is better written with inherit
────╯
```

obviously the difference being whether the `inherit` would require `inherit (foo) bar` rather than just `inherit bar`.

Just because the choice is there, I did wonder if W04 is the more objectionable of the two with;

```nix
inherit (eval) options;
```

seeming a bit of a pain compared to;

```nix
options = eval.options
```

when compared to `types = types;` ➡ `inherit types;` and while I enjoyed the brief diversionary contemplation, I'm not sure whether I'm that invested one way or the other!

</details>

So I've just disabled both for now.

## Add `nix fmt` support in flake

Just noticed this exists and figured it might be better to codify stuff in the flake, so it's easy to track and run.

As a side effect I realised that previously I hadn't applied `nixpkgs-fmt` recursively (I should have used `**/*.nix` not `**.nix`) so `nix fmt` does a better job than the manual command.

## Re-apply normalisation

As I said, I missed some files with original run of `nixpkgs-fmt` so this normalises those. Also undid then reapplied the `statix fix` with the new config which should remove the undesired addition of `inherit`.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -16,7 +16,7 @@ let
   };
 in
 {
-  inherit types;
+  types = types;
   create = cfg: types.diskoLib.create (eval cfg).config.devices;
   createScript = cfg: pkgs: pkgs.writeScript "disko-create" ''
     #!/usr/bin/env bash
--- a/doc.nix
+++ b/doc.nix
@@ -19,7 +19,7 @@ let
     ];
   };
   options = nixosOptionsDoc {
-    inherit (eval) options;
+    options = eval.options;
   };
   md = (runCommand "disko-options.md" { } ''
     cat >$out <<EOF
--- a/example/boot-raid1.nix
+++ b/example/boot-raid1.nix
@@ -13,7 +13,7 @@
             start = "0";
             end = "1M";
             part-type = "primary";
-            flags = ["bios_grub"];
+            flags = [ "bios_grub" ];
           }
           {
             type = "partition";
@@ -53,7 +53,7 @@
             start = "0";
             end = "1M";
             part-type = "primary";
-            flags = ["bios_grub"];
+            flags = [ "bios_grub" ];
           }
           {
             type = "partition";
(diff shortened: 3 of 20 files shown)


Lassulus:
I think the normal inherit are fine, the inherit_from is bad if it's a single thing we get from the namespace. if it's multiple things it can save a bit of code. But I guess there is no way to disable the inherit_from only if a single thing is inherited?

cscutcher:
> I think the normal inherit are fine, the inherit_from is bad if it's a single thing we get from the namespace. if it's multiple things it can save a bit of code. But I guess there is no way to disable the inherit_from only if a single thing is inherited?

I see what you mean. Yeah, I don't see a way to disable the check just for when it's a single item. 

The aim is just to normalise where stuff is uncontroversial and generally uninteresting, so I think it's fine to just disable the check generally (which is what I've done with this config).

Lassulus:
bors merge

PR #147, opened 2023-02-08
fix mdadm raid device numbering style

Tomaszal:
Currently creating an mdadm raid fails for me as it tries to create a filesystem on /dev/md/raidp1 device, which does not exist. This fixes that since /dev/md/raid1 exists instead.


Tomaszal:
Seeing that the checks fail, it's likely something on my setup that creates the different numbering style, but what could it be? I'm trying to run this on NixOS unstable on an ARM64 system, not sure what else might be useful to figure out the root cause.

Tomaszal:
I figured out the issue, from [mdadm manual](https://linux.die.net/man/8/mdadm):

> If the device name ends with a digit, the partition names add a 'p', and a number, e.g. /dev/md/home1p3. If there is no trailing digit, then the partition names just have a number added, e.g. /dev/md/scratch3.

PR #153, opened 2023-02-21
tests: add test names

lilyinstarlight:
Having to piece together which `vm-test-run-disko> [...]` is which test when running `nix flake check -vL` was getting a little annyoing 😄

So this makes them all have distinct names so when they are interleaved in Nix output, it's possible to tell them apart

Let me know if having names on the disko tests is undesired for any reason

### DIFF
--- a/tests/bcachefs.nix
+++ b/tests/bcachefs.nix
@@ -5,6 +5,7 @@ let
   linux-bcachefs = pkgs.callPackage ../linux-testing-bcachefs.nix { };
 in
 makeDiskoTest {
+  name = "bcachefs";
   disko-config = ../example/bcachefs.nix;
   extraTestScript = ''
     machine.succeed("mountpoint /");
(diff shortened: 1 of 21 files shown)


Mic92:
bors merge

Lassulus:
I guess I would have used the filename in the default.nix to set the name, but this is also fine

lilyinstarlight:
> I guess I would have used the filename in the default.nix to set the name, but this is also fine

You know, I guess I could have done that by passing in `name = test` when importing each of the files and let the files do `inherit name`. Would you like for me to change that @Lassulus?

Lassulus:
> > I guess I would have used the filename in the default.nix to set the name, but this is also fine
> 
> You know, I guess I could have done that by passing in `name = test` when importing each of the files and let the files do `inherit name`. Would you like for me to change that @Lassulus?

if you want to, I guess it's also fine like this

PR #154, opened 2023-02-21
tests: adjust wait_for_text calls to work for systemd-initrd or old stage-1

lilyinstarlight:
Related to #108, this fixes `wait_for_text` calls to match against both the old stage-1 and systemd-initrd console prompts for passphrases

It'll be needed later if we intend to enable systemd-initrd on some tests or do a test matrix with it both on+off


Mic92:
bors merge

PR #162, opened 2023-03-03
turn examples into nixos-config

Lassulus:


BREAKING_CHANGES.md
```diff
@@ -1,3 +1,7 @@
+2023-03-03 XXXXXXX
+
+the disko config needs now to be supplied in an `disko.config =` attrset
```
Mic92, line 3: Don't forget to also update the quickstart.md

BREAKING_CHANGES.md
```diff
@@ -1,3 +1,11 @@
+2023-03-22 2624af6
+
+disk config now needs to be inside a disko.devices attrset always
```
Mic92, line 3: It's not clear from this one line what the changes would need to be fore a user.
Also the tutorial needs an update: https://github.com/nix-community/disko/blob/master/docs/quickstart.md

default.nix
```diff
@@ -7,9 +7,9 @@ let
   eval = cfg: lib.evalModules {
     modules = lib.singleton {
       # _file = toString input;
-      imports = lib.singleton { devices = cfg; };
+      imports = lib.singleton { disko.devices = cfg.disko.devices; };
```
Mic92, line 10: There could be a warning here to help user upgrading there configuration i.e. if no `disko.devices` is set in the file.

BREAKING_CHANGES.md
Lassulus: https://github.com/nix-community/disko/pull/162/files#diff-988a5392cf907f7c5728b8ac90445facc7f835f3c00375764ec99769dcde2d5a

also people usually don't need to change anything if they haven't used the cli for installation. So it's unclear what kind of structure people have used before

default.nix
Lassulus: what would the warning say? right now we cat a warning, that disko.devices is not set correcty, which should be kinda clear enough?

PR #168, opened 2023-03-07
add support for Hybrid MBR

### LINKED ISSUE #166: Real hybrid MBR/GPT partitioning scheme
i notice that every _hybrid_ example is not an Hybrid partitioning scheme, only a  GPT BIOS compatible. Am I wrong?

I need a real Hybrid where `dos` table have a first `vfat` (0x0C) partition and then a full disk `gpt` (0xEE) type partition:

```
$ fdisk -t dos -l /dev/sda
(...)
Device                    Boot Start   End Sectors  Size Id Type
/dev/sda1                 *       34 65569   65536   32M  c W95 FAT32 (LBA)
/dev/sda2                          1    33      33 16.5K ee GPT
```

And `gpt` table have the _real_ partition info:

```
$ fdisk -t gpt -l /dev/sda
(...)
Device                    Start      End  Sectors Size Type
/dev/sda1                    34    65569    65536  32M EFI System
/dev/sda2                 65570 21037089 20971520  10G Linux root (ARM-64)
```

I guess a new interface similar to ones for lvm or mdraid will be needed to implement this.

oneingan:
Closes #166

### DIFF
--- a/example/hybrid-mbr.nix
+++ b/example/hybrid-mbr.nix
@@ -1,0 +1,62 @@
+{ disks ? [ "/dev/sda" ], ... }:
+{
+  disk = {
+    main = {
+      type = "disk";
+      device = builtins.elemAt disks 0;
+      content = {
+        type = "hybrid_table";
+        efiGptPartitionFirst = false;
+        hybrid_partitions = [
+          {
+            type = "hybrid_partition";
+            gptPartitionNumber = 1;
+            mbrPartitionType = "0x0c";
+            mbrBootableFlag = false;
+          }
+        ];
+        content = {
+          type = "table";
+          format = "gpt";
+          partitions = [
+            {
+              type = "partition";
+              name = "TOW-BOOT-FI";
+              start = "0%";
+              end = "32MiB";
+              fs-type = "fat32";
+              content = {
+                type = "filesystem";
+                format = "vfat";
+                mountpoint = "/boot/firmware";
+              };
+            }
+            {
+              type = "partition";
+              name = "ESP";
+              start = "32MiB";
+              end = "512MiB";
+              bootable = true;
+              content = {
+                type = "filesystem";
+                format = "vfat";
+                mountpoint = "/boot";
+              };
+            }
+            {
+              type = "partition";
+              name = "root";
+              start = "512MiB";
+              end = "100%";
+              content = {
+                type = "filesystem";
+                format = "ext4";
+                mountpoint = "/";
+              };
+            }
+          ];
+        };
+      };
+    };
+  };
+}
--- a/example/hybrid-mbr.nix
+++ b/example/hybrid-mbr.nix
@@ -28,7 +28,7 @@
               content = {
                 type = "filesystem";
                 format = "vfat";
-                mountpoint = "/boot/firmware";
+                mountpoint = null;
               };
             }
             {
--- a/types/default.nix
+++ b/types/default.nix
@@ -24,7 +24,7 @@ rec {
 
     # option for valid contents of devices
     deviceType = lib.mkOption {
-      type = lib.types.nullOr (diskoLib.subType { inherit (subTypes) table btrfs filesystem zfs mdraid luks lvm_pv swap; });
+      type = lib.types.nullOr (diskoLib.subType { inherit (subTypes) table hybrid_table btrfs filesystem zfs mdraid luks lvm_pv swap; });
       default = null;
       description = "The type of device";
     };
@@ -133,7 +133,7 @@ rec {
           || lib.hasSuffix "Hook" n
           || isAttrsOfSubmodule o
           # TODO don't hardcode diskoLib.subType options.
-          || n == "content" || n == "partitions"
+          || n == "content" || n == "partitions" || n == "hybrid_partitions"
         );
       in
       lib.toShellVars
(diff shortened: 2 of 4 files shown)


example/hybrid-mbr.nix
```diff
@@ -0,0 +1,62 @@
+{ disks ? [ "/dev/sda" ], ... }:
```
Mic92, line 1: Do you have a suggestion how we should name `hybrid.nix` in that case since it's still a bit confusing to have both.

uningan: As I comment here https://github.com/nix-community/disko/issues/166#issuecomment-1455683887 existent `gpt-bios-compat.nix` is enough. `hybrid.nix` or `hybrid-tmpfs-onroot.nix` looks confusing and redundant for me.

Lassulus: hmm, the gpt-bios-compat can't do EFI because it has no efi or boot partition

Mic92: Yes. I would also prefer something that can support both EFI/BIOS. We have installer that works with both EFI and BIOS and makes it unnecessary for people to have to figure out what mode there machines was booted from.

uningan: Sure, so maybe we could reorganize as:

remove `gpt-bios-compat.nix` (I'm not sure which is the use case of this without EFI partition)
remove `hybrid-tmpfs-onroot.nix` (or not, but for clarity on discussion)
`hybrid.nix` -> `efi-bios-compat.nix`

or, if we want to keep the hybrid keyword:

`hybrid_boot.nix` vs `hybrid_mbr.nix`

Also, being stricts, seems like the real name of hybrid boot using `bios_grub` flags is GPT-only protocol. As in https://wiki.archlinux.org/title/Arch_boot_process#Feature_comparison and https://repo.or.cz/syslinux.git/blob/HEAD:/doc/gpt.txt

uningan: Hi, maybe can we get this to a conclusion?

example/hybrid-mbr.nix
```diff
@@ -0,0 +1,62 @@
...
+      type = "disk";
+      device = builtins.elemAt disks 0;
+      content = {
+        type = "hybrid_table";
+        efiGptPartitionFirst = false;
+        hybrid_partitions = [
+          {
+            type = "hybrid_partition";
+            gptPartitionNumber = 1;
+            mbrPartitionType = "0x0c";
```
Lassulus, line 14: what are other valid mbrPartitionType options?

example/hybrid-mbr.nix
```diff
@@ -0,0 +1,62 @@
...
+    main = {
+      type = "disk";
+      device = builtins.elemAt disks 0;
+      content = {
+        type = "hybrid_table";
+        efiGptPartitionFirst = false;
+        hybrid_partitions = [
+          {
+            type = "hybrid_partition";
+            gptPartitionNumber = 1;
```
Lassulus, line 13: isn't the index in the list enough? why do we need to count here? what would happen if I set this to 3?

Lassulus (commented):
Not sure if we should support this, seems kinda clunky and I don't see many usecases? can't this be achieved with the existing table type and some extra flags?

example/hybrid-mbr.nix
Lassulus: maybe we can call it bios-efi-boot? since it does both?

example/hybrid-mbr.nix
uningan: whatever hexcode dos table partitions support, as in https://en.wikipedia.org/wiki/Partition_type#List_of_partition_IDs.

If `null` then it guess partition type based on GPT one.

example/hybrid-mbr.nix
uningan: You need to choose which GPT partition want to make available in DOS table. Maybe a sane default using index is ok.

uningan:
> Not sure if we should support this, seems kinda clunky and I don't see many usecases? can't this be achieved with the existing table type and some extra flags?

Sure not. Use cases are multiple, mine is to create a working SD Card for Rapsberry Pi. But is also extensible to boot multiple operating systems setups.

example/hybrid-mbr.nix
Lassulus: but would it still be a hybrid mbr if the partitionType is not "0x0c"?

uningan: Sure, set to `0x0c`have sense for RPi because RPi boot process look exactly for it, but other use cases could work with default value.

example/hybrid-mbr.nix
uningan: correct.

DavHau (commented):
I think the structure of the `hybrid_table` type should be simlified.
- remove the additional `hybrid_partitions` list and instead introduce a new partition type containing all fields
- this will get rid of the `gptPartitionNumber`, as there is only one list of partitions left
Modified example:
```nix
{ disks ? [ "/dev/sda" ], ... }:
{
  disk = {
    main = {
      type = "disk";
      device = builtins.elemAt disks 0;
      content = {
        type = "hybrid_table";
        efiGptPartitionFirst = false;
        partitions = [
          {
            # the `type` field is not needed and can be omitted
            type = "hybrid_partition";

            name = "TOW-BOOT-FI";
            start = "0%";
            end = "32MiB";
            fs-type = "fat32";
            content = {
              type = "filesystem";
              format = "vfat";
              mountpoint = null;
            };

            # mbr specific fields
            mbrPartitionType = "0x0c";
            bootable = true; # previously named `mbrBootableFlag`
          }
          {
            type = "partition";
            name = "ESP";
            start = "32MiB";
            end = "512MiB";
            bootable = true;
            content = {
              type = "filesystem";
              format = "vfat";
              mountpoint = "/boot";
            };
          }
          # ...
        ];
      };
    };
  };
}

```

Mic92:
Replaced by https://github.com/nix-community/disko/pull/508

PR #181, opened 2023-03-18
luks: add key file size & offset parameters

jkachmar:
Pretty self-explanatory: this adds config options for LUKS `--keyfile-size` and `--keyfile-offset` parameters, so that someone can configure a specific segment of the key file to be used when creating the container.

Manually tested that this works & verified that it's selecting a chunk at the specified offset from the key file.

i should go and add a test for this, since it's not too much work, but i figured i would open the PR to see if there are any objections to this.

### DIFF
--- a/types/luks.nix
+++ b/types/luks.nix
@@ -95,6 +95,9 @@
         [
           # TODO do we need this always in initrd and only there?
           { boot.initrd.luks.devices.${config.name}.device = dev; }
+          (lib.optional (config.keyFile != null) { boot.initrd.luks.devices.${config.name} = {inherit (config) keyFile; };})
+          (lib.optional (config.keyFileSize != null) { boot.initrd.luks.devices.${config.name} = {inherit (config) keyFileSize; };})
+          (lib.optional (config.keyFileOffset != null) { boot.initrd.luks.devices.${config.name} = { inherit (config) keyFileOffset; };})
         ] ++ (lib.optional (config.content != null) (config.content._config "/dev/mapper/${config.name}"));
       description = "NixOS configuration";
     };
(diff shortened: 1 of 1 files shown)


Lassulus:
looking good, have no usecase though, a test would be nice and now needs a rebase since https://github.com/nix-community/disko/pull/186 was merged

jkachmar:
rebased; will test locally & then try to make a representative test case for the suite.

shame about the `luksOpenArgs` not being passed into the NixOS module; i tried to fix that & then noticed https://github.com/nix-community/disko/pull/186#issuecomment-1478940710 :(

types/luks.nix
```diff
@@ -65,6 +95,9 @@
         [
           # TODO do we need this always in initrd and only there?
           { boot.initrd.luks.devices.${config.name}.device = dev; }
+          (lib.optional (config.keyFile != null) { boot.initrd.luks.devices.${config.name} = {inherit (config) keyFile; };})
+          (lib.optional (config.keyFileSize != null) { boot.initrd.luks.devices.${config.name} = {inherit (config) keyFileSize; };})
+          (lib.optional (config.keyFileOffset != null) { boot.initrd.luks.devices.${config.name} = { inherit (config) keyFileOffset; };})
```
jkachmar, line 100: hmm, i hadn't thought about this but i guess disko's `keyFile` option isn't meant to be the same as NixOS' right?

if you'd prefer to avoid making this assumption, i can just remove this commit & manually set these options in my NixOS config separately.

jkachmar:
i think #186 supersedes this; the only major difference is a result of https://github.com/nix-community/disko/pull/186#issuecomment-1478940710 which i can work around & will probably go away once upstream has been updated to allow these args to be propagated to NixOS

PR #186, opened 2023-03-21
luks: rename extraArgs to extraFormatArgs, add extraOpenArgs

lheckemann:


Lassulus:
ah, that was a breaking change, so we should comment it in BREAKING_CHANGES.md

Lassulus:
which I have now done in https://github.com/nix-community/disko/pull/162

Lassulus:
the luksOpenArgs are now only passed when first opening the device, not if booted later with nixos because they are not passed into the nixos config, is this something we want?

lheckemann:
Ouch, that was absolutely not the plan :( it seems the NixOS module doesn't actually support passing arbitrary flags...

PR #196, opened 2023-04-07
types: refactor partitions, lvm_lv and zfs_dataset

Lassulus:
inline some non polymorphic types so the code becomes more compact


types/zfs_fs.nix
```diff
@@ -7,15 +7,11 @@
...
-      default = "zfs_dataset";
+      type = lib.types.enum [ "zfs_fs" ];
+      default = "zfs_fs";
       internal = true;
       description = "Type";
     };
-    zfs_type = lib.mkOption {
-      type = lib.types.enum [ "filesystem" "volume" ];
-      description = "The type of the dataset";
-    };
```
Mic92, line 18: That might be useful:

```
  imports = [
    (mkRenamedOptionModule ["zfs_type" ] ["type" ])
  ];
```

Mic92: Than you would also avoid situations like https://github.com/nix-community/disko/issues/197

Lassulus: ah, that won't be useful, since the correct type cannot be found. but you get this error right now already:
```
error: No type option set in:
       {
         options = {
           encryption = "aes-256-gcm";
           keyformat = "passphrase";
           keylocation = "file:///tmp/secret.key";
           mountpoint = "none";
         };
         size = "20M";
         zfs_type = "filesystem";
       }
(use '--show-trace' to show detailed location information)
```

BREAKING_CHANGES.md
```diff
@@ -1,3 +1,15 @@
+2023-04-07 7d70009
+
+zfs_datasets are now split into zfs_fs and zfs_volume. zfs_type got removed. size is only available for zfs_volume
```
Mic92, line 3:
```suggestion
Date: 2023-04-07

Commit: 7d70009

Changes:

    ZFS datasets have been split into two types: zfs_fs and zfs_volume.
    The zfs_type attribute has been removed.
    The size attribute is now only available for zfs_volume.

Updated example/zfs.nix file:

```bash
datasets = {
  zfs_fs = {
    type = "zfs_fs";
    mountpoint = "/zfs_fs";
    options."com.sun:auto-snapshot" = "true";
  };
  zfs_unmounted_fs = {
    type = "zfs_fs";
    options.mountpoint = "none";
  };
  zfs_legacy_fs = {
    type = "zfs_fs";
    options.mountpoint = "legacy";
    mountpoint = "/zfs_legacy_fs";
  };
  zfs_testvolume = {
    type = "zfs_volume";
    size = "10M";
    content = {
      type = "filesystem";
      ...

Note: The zfs_type attribute has been replaced with a type attribute for each dataset, and the size attribute is only available for zfs_volume. These changes have been reflected in the example/zfs.nix file.
```

```

PR #204, opened 2023-04-16
Improve example code in README

nomeata:
It seems the type is not needed anymore.


Lassulus:
Ah always forget the README example. thank you!

PR #211, opened 2023-04-20
zpool: add option to mount zpool not at /

SuperSandro2000:
Otherwise creating and mounting the zpool in a live image removes the root disk of the live image.

Tested on physical hardware.

### DIFF
--- a/types/zpool.nix
+++ b/types/zpool.nix
@@ -37,6 +37,17 @@
       default = [ "defaults" ];
       description = "Options to pass to mount";
     };
+    mountRoot = lib.mkOption {
+      type = lib.types.str;
+      default = "/";
+      example = "/mnt";
+      description = ''
+        The root location where the zpool should be mounted.
+
+        Note:
+        Leaving this at the default "/" might break your live system.
+      '';
+    };
     datasets = lib.mkOption {
       type = lib.types.attrsOf (diskoLib.subType { inherit (subTypes) zfs_fs zfs_volume; });
       # type = lib.types.attrsOf subTypes.zfs_fs;
@@ -55,7 +66,7 @@
       default = _: ''
         readarray -t zfs_devices < <(cat "$disko_devices_dir"/zfs_${config.name})
         zpool create ${config.name} \
-          ${config.mode} \
+          -R ${config.mountRoot} ${config.mode} \
           ${lib.concatStringsSep " " (lib.mapAttrsToList (n: v: "-o ${n}=${v}") config.options)} \
           ${lib.concatStringsSep " " (lib.mapAttrsToList (n: v: "-O ${n}=${v}") config.rootFsOptions)} \
           "''${zfs_devices[@]}"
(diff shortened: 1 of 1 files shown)


types/zpool.nix
```diff
@@ -55,7 +66,7 @@
       default = _: ''
         readarray -t zfs_devices < <(cat "$disko_devices_dir"/zfs_${config.name})
         zpool create ${config.name} \
-          ${config.mode} \
+          -R ${config.mountRoot} ${config.mode} \
```
phaer, line 69: `man zpool-create` suggests that implicitly sets `cachefile=none` and `altroot` properties on the pool, even if no alt-root is used. Do we want this or should we use `lib.optionalString` here?

types/zpool.nix
```diff
@@ -70,7 +81,8 @@
         in
         {
           dev = ''
-            zpool list '${config.name}' >/dev/null 2>/dev/null || zpool import '${config.name}'
+            zpool list '${config.name}' >/dev/null 2>/dev/null || \
+              zpool import -R ${config.mountRoot} '${config.name}'
```
phaer, line 85: Given that you set the altroot here, during import, is it even necessary to do so above, during creation, as well? The remark about `cachefile=none` still applies here.

phaer (changes requested):

types/zpool.nix
SuperSandro2000: I just set ``/`` to keep backwards compatibility but it doesn't make much sense as it will most likely overwrite the system that the command is run from.

types/zpool.nix
SuperSandro2000: Yes, because zpool create mounts the pool already.

oddlama (approved):

types/zpool.nix
phaer: Ah yes, did a bit more docs reading, `cachefile=none` seems harmless.

I'd still prefer to only set `-R` if actually used, but don't care too much :)

types/zpool.nix
phaer: Maybe I am missing something, but why do we need to set altroot during creation? Shouldn't the import suffice to mount it in a live system? I'll try to do a practice test tomorrow

types/zpool.nix
oddlama: It's important to have -R on create too, otherwise any datasets that have `option.mountpoint` set will be mounted when they are created. If altroot is not set, this can shadow existing mountpoints like `/`, which would be fatal.

types/zpool.nix
oddlama: It's important to have -R on create too, otherwise any datasets that have `option.mountpoint` set will be mounted when they are created. If altroot is not set, this can shadow existing mountpoints like `/`, which would be fatal.

oddlama:
LGTM. I'm already relying on this fork to make disko work from a live environment. This is a very important change to support the non-legacy mountpoints.

SuperSandro2000:
Shouldn't this use disko.rootMountPoint? #238

Lassulus:
disko.rootMountPoint is the location where disko mounts the whole device tree with the mountScript (usually /mnt), this option tells where to mount the zpool to, which can be any folder, this gets appended to the rootMountPoint only when running the mountScript

SuperSandro2000:
This setting is basically duplicated with disko.rootMountPoint because I didn't know about it when I implemented this.

PR #213, opened 2023-04-22
Add `enableCryptodisk` to quickstart guide

ghostbuster91:
I was surprised to see some errors after following the quickstart guide. 
It turned out that it was because of that one missing setting. 

I think that it would be good to mention it somewhere, not sure if the suggested place it the best option.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -127,6 +127,8 @@ the existing lines about `systemd-boot`:
    boot.loader.grub.version = 2;
    boot.loader.grub.efiSupport = true;
    boot.loader.grub.efiInstallAsRemovable = true;
+   # uncomment below line if you use luks
+   #boot.loader.grub.enableCryptodisk = true; 
 # ...
 ```
 


phaer:
Thanks for your input!

 Happy to include that settings, but might I ask if you used one of the example configs with the quickstart guide? IIUC, your setup requires that GRUB itself is able to unlock LUKS. While this solution offers a different set of trade-offs, I think the more conventional way to solve this is to keep the Kernel and Initrd on an unencrypted /boot and leaving LUKS handling to the kernel.

So the current comment doesn't seem entirely true, you only need it if you use GRUB to load Kernel and Initrd from a LUKS encrypted volume, right?

ghostbuster91:
That makes sense. I've just started getting deeper into partitions & uefi vs legacy and in some places my understanding is still incomplete. 

So, it seems that you are right, that flag is redundant in case of example scenarios, and that I have just jumped into that conclusion to hastily. 

I was trying to run a modified example with luks on a legacy system and after some failures I discovered existence of `enableCryptodisk`, which seems at a first glance like something that could help. More or less at the same time, after learning more about uefi vs legacy, I switched virtualbox to use uefi and then the example with luks worked without any issues. At that moment I incorrectly assumed that it had also something to do with `enableCryptodisk`. 

I have just tested again the modified example of luks for legacy bios with that flag enabled and as expected, the presence of the flag didn't change anything. 

While, I won't need that I am still curious why my modified example of luks for legacy bios didn't work. 

*I am not opening issue yet, because maybe the problem is again with my incomplete understanding of the space*

Here is how it looks like:
```nix 
{ disks ? [ "/dev/vdb" ], ... }: {
  disko.devices = {
    disk = {
      vdb = {
        type = "disk";
        device = builtins.elemAt disks 0;
        content = {
          type = "table";
          format = "gpt";
          partitions = [
            {
              name = "boot";
              start = "0";
              end = "100M";
              part-type = "primary";
              flags = [ "bios_grub" ];
            }
            {
              name = "luks";
              start = "100M";
              end = "100%";
              content = {
                type = "luks";
                name = "crypted";
                content = {
                  type = "lvm_pv";
                  vg = "pool";
                };
              };
            }
          ];
        };
      };
    };
    lvm_vg = {
      pool = {
        type = "lvm_vg";
        lvs = {
          root = {
            size = "20G";
            content = {
              type = "filesystem";
              format = "ext4";
              mountpoint = "/";
              mountOptions = [
                "defaults"
              ];
            };
          };
          home = {
            size = "20G";
            content = {
              type = "filesystem";
              format = "ext4";
              mountpoint = "/home";
            };
          };
        };
      };
    };
  };
}

```


This is the error I am getting:

Screenshot from 2023-04-22 23-32-53

Screenshot from 2023-04-21 23-39-18

Screenshot from 2023-04-22 23-30-53

phaer:
Your example seems to be missing an unencrypted /boot. GRUB tries to load the kernel from a LVM volume inside LUKS, but can't find it as it hasn't unlocked LUKS yet. If you add an unecrypted /boot the flow is as follows: GRUB loads your Kernel and Initrd from /boot, the initrd unlocks LUKS, activates LVM and mounts /.

ghostbuster91:
Thanks for the response, maybe I don't understand something, but it for me it looks that I have unencrypted boot partition:
```
partitions = [
            {
              name = "boot";
              start = "0";
              end = "100M";
              part-type = "primary";
              flags = [ "bios_grub" ];
            }
            {
              name = "luks";
              start = "100M";
              end = "100%";
              content = {
                type = "luks";
                name = "crypted";
                content = {
                  type = "lvm_pv";
                  vg = "pool";
                };
              };
            }
          ];
```

phaer:
Thats a partition named "boot" to leave a bit of space at the beginnig of your disk to "embed" grub into. You need another, formatted and mounted "/boot" partition. Naming does not matter, but the mountpoint does.

Please refer to the existing examples and feel free to open another issue if things remain unclear, but at this point the discussion isn't about `enableCryptodisk` anymore :)

ghostbuster91:
Ok, I thought that the 
> another formatted and mounted "/boot" partition

is needed only when running in uefi mode. 

I will read more about that and check few scenarios, once again thank you for your help.

phaer:
> is needed only when running in uefi mode.

They are different: With UEFI you often require an ESP partition which afaiks needs to be FAT32, but can be re-used as your /boot.
You always need /boot with linux, but it can be a normal directory if you don't use LUKS.
With LUKS you either need to leave your kernel and initrd unencrypted (external /boot) or support to unlock in your botoloader (enableCyrptoDisk).

Issue #141, opened 2023-02-03
Support depends attribute


ktrinh-anduril:
Hello, is there plan to support the `depends` attribute that nixos filesystems option support? I think right now the mount order is only determined by parsing the name of the mount point for parent and child relationship but it would be nice to also have the option to explicitly specify the dependency relationship.

Mic92:
Can you add some example where this would be useful to have?

oddlama:
I'd also need this and I can provide an example. I am using the layout given below together with impermanence. The issue is that the datasets are created and mounted by disko in alphabetical order, which in the case of mounting is the wrong way around. Currently `local/nix` is mounted to `/nix` first, but is then shadowed by mounting `local/root` to `/`. So here I'd need an additional dependency (or a more complicated automatic resolution that inspects mountpoints like ZFS does internally).

Creation order should also be (partly) determined by this, since the dataset creation will implicitly mount them when using the non-legacy ZFS mountpoints (i.e. when setting `options.mountpoint`). To avoid issues with conflicting creation and mount order, one could also export the zpool after creating the datasets, since `zpool import` will then do the correct thing.

```nix
rpool = {
  datasets = {
    "local".type = "zfs_fs";
    "local/root" = {
      type = "zfs_fs";
      postCreateHook = "zfs snapshot rpool/local/root@blank";
      options.canmount = "on";
      mountpoint = "/";
    };
    "local/nix" = {
      type = "zfs_fs";
      options.canmount = "on";
      mountpoint = "/nix";
    };
    # ... some other persistent datasets
  };
};
```

Lassulus:
huh, maybe there is a bug. mounting should happen alphabetically sorted by mountpoint. but maybe that is not true for if mounting it via NixOS. for creation exporting and reimporting sounds like a good solution

oddlama:
You are right, this only seems to affect datasets mounted on creation, the mounting section seems to be sorted correctly.

I just noticed that the import statements are not using `-l` to load required encryption keys, which would be necessary to support encrypted datasets with `keylocation=prompt` when doing the export import workaround. Currently the datasets are still unlocked due to creation.

oddlama:
Turns out the export-import workaround doesn't work that easily since `zpool export` fails when there are shadowed datasets (it tries to unmount in the correct reversed order, which won't work). Instead we should create datasets with `-u` to prevent mounting them in the first place. Then the mounting logic can mount them as usual. I've created #218 that addresses that.

[...]

PR #218, opened 2023-04-29
zpool: don't mount created datasets directly; and fix encrypted import

oddlama:
This is necessary to support non-legacy mounts, since the creation order and the required mounting order can be different in order to not shadow any datasets (see #141).

Adding `-l` to the import statement is also necessary to allow encrypted datasets to be mounted later using disko-mount.

### DIFF
--- a/types/zfs_fs.nix
+++ b/types/zfs_fs.nix
@@ -38,8 +38,11 @@
     };
     _create = diskoLib.mkCreateOption {
       inherit config options;
+      # -u prevents mounting newly created datasets, which is
+      # important to prevent accidental shadowing of mount points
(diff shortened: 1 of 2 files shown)


phaer (approved):
Thanks, LGTM!

Lassulus (approved):

Lassulus:
thanks a lot for your contributions (everyone who contributed). disko feels like more and more of a community effort to get the hard topic of `disks and stuff` finally in a manageable format.

we now have a disko channel on matrix btw, you are invited to participate there also :) 
https://matrix.to/#/#disko:matrix.org

PR #222, opened 2023-05-01
add example zfs-luks-raidz

mrVanDalo:
It took a while to set up a 2 zpool, one with raidz2 and the other just "normal". Thought I might be a good idea to share.

### DIFF
--- a/example/zfs-raidz.nix
+++ b/example/zfs-raidz.nix
@@ -1,0 +1,123 @@
+# meant for NAS setups with raidzX approach.
+{ root_disk ? "/dev/vdb"
+, raid_disks ? {
+    # better use /dev/disk/by-id/
+    "vendor_id1" = "/dev/vdc";
+    "vendor_id2" = "/dev/vdd";
+    "vendor_id3" = "/dev/vde";
+  }
+, ...
+}:
+{
+  disko.devices = {
+    disk = {
+      root = {
+        type = "disk";
+        device = root_disk;
+        content = {
+          type = "table";
+          format = "gpt";
+          partitions = [
+            {
+              name = "ESP";
+              start = "0";
+              end = "500MiB";
+              bootable = true;
+              content = {
+                type = "filesystem";
+                format = "vfat";
+                mountpoint = "/boot";
+                mountOptions = [
+                  "defaults"
+                ];
+              };
+            }
+            {
+              name = "zfs";
+              start = "500MiB";
+              end = "100%";
+              content = {
+                type = "zfs";
+                pool = "zroot";
+              };
+            }
+          ];
+        };
+      };
+    } // builtins.mapAttrs
+      (_: device_path: {
+        type = "disk";
+        device = device_path;
+        content = {
+          type = "table";
+          format = "gpt";
+          partitions = [
+            {
+              name = "zfs";
+              start = "0";
+              end = "100%";
+              content = {
+                type = "zfs";
+                pool = "zraid";
+              };
+            }
+          ];
+        };
+      }
+      )
+      raid_disks;
+
+    zpool = {
+
+      zroot = {
+        type = "zpool";
+        rootFsOptions = {
+          mountpoint = "none";
+          canmount = "off";
+        };
+        datasets = {
+          "root" = {
+            type = "zfs_fs";
+            mountpoint = "/";
+            options = {
+              mountpoint = "legacy";
+              compression = "lz4";
+            };
+          };
+        };
+      };
+
+      # `zpool import -f zraid` once on the first boot and reboot as well everytime you change networking.hostname or networking.hostId
+      zraid = {
+        type = "zpool";
+        mode = "raidz1";
+        rootFsOptions = {
+          mountpoint = "none";
+          canmount = "off";
+        };
+        datasets = {
+          "media" = {
+            type = "zfs_fs";
+            mountpoint = "/media";
+            options = {
+              mountpoint = "legacy";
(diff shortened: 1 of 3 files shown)


Lassulus:
Cool. Can you also add a test?

mrVanDalo:
> Cool. Can you also add a test?

Yes I will do, tomorrow first thing in the morning.

example/zfs-luks-raidz.nix
```diff
@@ -0,0 +1,148 @@
+# meant for NAS setups with raidzX approach.
+{ ... }:
+let
+  raid_disks = {
+    "TOSHIBA_1360A003FVGG" = "/dev/disk/by-id/ata-TOSHIBA_MG08ACA16TE_1360A003FVGG";
+    "TOSHIBA_1360A00BFVGG" = "/dev/disk/by-id/ata-TOSHIBA_MG08ACA16TE_1360A00BFVGG";
+    "TOSHIBA_1360A00VFVGG" = "/dev/disk/by-id/ata-TOSHIBA_MG08ACA16TE_1360A00VFVGG";
+    "TOSHIBA_41R0A0EBF57H" = "/dev/disk/by-id/ata-TOSHIBA_MG08ACA16TE_41R0A0EBF57H";
+    "TOSHIBA_5120A03WF57H" = "/dev/disk/by-id/ata-TOSHIBA_MG08ACA16TE_5120A03WF57H";
+  };
```
Mic92, line 10: We usually make those over-writable for the test. The NixOS test framework allows to create additional disks.

example/zfs-luks-raidz.nix
```diff
@@ -0,0 +1,148 @@
...
+          };
+          "backup" = {
+            type = "zfs_fs";
+            mountpoint = "/backup";
+            options = {
+              mountpoint = "legacy";
+              compression = "lz4";
+              "com.sun:auto-snapshot" = "false";
+              "com.sun:auto-snapshot:daily" = "true,keep=32";
+            };
```
Mic92, line 140: I think people at this point get the idea:

```suggestion
```

example/zfs-luks-raidz.nix
mrVanDalo: Like this : 
```
{
  root_disk ? "/dev/vdb",
  raid_disks ? {
    # better use /dev/disk/by-id/
    "vendor_id02" = "/dev/vdc";
    "vendor_id03" = "/dev/vdd";
    "vendor_id04" = "/dev/vde";
    "vendor_id05" = "/dev/vdf";
    "vendor_id06" = "/dev/vdg";
  }, ... }:
```

mrVanDalo:
I'm stuck, ... like on my own set up, the first time I reboot I have to run `zpool import -f zraid` to make zfs import the `zraid-zpool` during boot time. 
It might be possible to use `boot.zfs.forceImportAll` or `boot.zfs.extraPools` in the `configuration.nix` But I don't know how to add lines to the `configuration.nix` of the test.

Lassulus:
https://github.com/nix-community/disko/blob/master/tests/zfs.nix#L7

mrVanDalo:
> I'm stuck, ... like on my own set up, the first time I reboot I have to run `zpool import -f zraid` to make zfs import the `zraid-zpool` during boot time. It might be possible to use `boot.zfs.forceImportAll` or `boot.zfs.extraPools` in the `configuration.nix` But I don't know how to add lines to the `configuration.nix` of the test.

Still not working :( 
I tryed `forceImportAll` and `extraPools`. I guess we have to make the test reboot 2 times and run tests after the second reboot.

example/zfs-luks-raidz.nix
Mic92: Yes. We can also make it just one list and assume that disks[0] is the root disk? No strong opinion here.

Mic92:
> > I'm stuck, ... like on my own set up, the first time I reboot I have to run `zpool import -f zraid` to make zfs import the `zraid-zpool` during boot time. It might be possible to use `boot.zfs.forceImportAll` or `boot.zfs.extraPools` in the `configuration.nix` But I don't know how to add lines to the `configuration.nix` of the test.
> 
> Still not working :( I tryed `forceImportAll` and `extraPools`. I guess we have to make the test reboot 2 times and run tests after the second reboot.
> 
> Alternatively I could only define 1 zpool which is the root and a raidz2. Not sure it the example makes any sense in that case.

Maybe something to look into on a Wednesday.

example/zfs-luks-raidz.nix
mrVanDalo: Hmm doing this kinda make the example even more useless, because naming your devices based on your vendor is part of a zfs-raid strategy to know which hard-drive needs to be changed if it failes. Of course you can always find your drive by investigating `/dev/disk/by-id` and `/dev/mapper`, but the idea here was to make this obsolete right from the start.

Lassulus: you can have the default value still be /dev/disk/by-id. the test runner passes the disks as disks to the example, so they would get override, root_disk has to be disks[0] though

mrVanDalo: Yeah for the root pool this might work, but not for the raid pool. This way I avoid to format the same partition twice.
For the raid pool I need a map. Hmm maybe I could use `builtins.basenameOf` to get rid of the map. Will think about it, ... at the moment it seems the test will never run anyway.

mrVanDalo:
@Mic92 I think we can close this pull request or skip the test.

PR #227, opened 2023-05-07
Add support for /dev/mapper LVM partitions in deviceNumbering

### LINKED ISSUE #226: deviceNumbering heuristic doesn't work for LVM partitions
I can't give `/dev/mapper/XXXX` to disko because deviceNumbering cannot detect it properly. I am not sure it is always correct to just use `${dev}${number}` when the prefix is `/dev/mapper`, so I can't just add it to the function as recommended.

It seems to me that it would be a lot more reliable to get the partition children either with sysfs or lsblk, instead of using this heuristic.

```
$ lsblk --json --paths -o NAME -n /dev/nvme0n1
{
   "blockdevices": [
      {"name":"/dev/nvme0n1",
         "children": [
            {"name":"/dev/nvme0n1p1",
               "children": [
                  {"name":"/dev/mapper/vg-lv",
                     "children": [
                        {"name":"/dev/mapper/vg-lv1"}
                     ]
                  },
               ]
            }
         ]
      }
   ]
}
```

seirl:
Fix #226

### DIFF
--- a/types/default.nix
+++ b/types/default.nix
@@ -59,6 +59,8 @@ rec {
         "${dev}p${toString index}" # /dev/nvme0n1p1 style
       else if match "/dev/md/.+" dev != null then
         "${dev}${toString index}" # /dev/md/raid1 style
+      else if match "/dev/mapper/.+" dev != null then
+        "${dev}${toString index}" # /dev/mapper/vg-lv1 style
       else
         abort ''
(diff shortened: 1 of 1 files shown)


Lassulus:
usually a test would be nice, but I guess this is not really a common usecase and we also don't test all the other deviceNumbering schemas right niw

Lassulus:
bors merge

Issue #192, opened 2023-03-29
neededForBoot flag


jsimonetti:
Hi,

Is there a way to make a fileSystem marked `neededForBoot`?
Right now I am having some issues because a filesystem is not mounted soon enough.

lilyinstarlight:
I currently set the nixos options for it just fine, since disko doesn't really need it for formatting or mounting

See my config:

https://github.com/lilyinstarlight/foosteros/blob/9dc10cbc68b56c02cd10642061a14b7fddf8adb5/hosts/bina/disks.nix#L117-L118

jnsgruk:
I'm interested in this too, for a slightly different reason.

I have a setup where I have a root disk that is just plain btrfs on a smaller SSD. I then have a data drive that is larger, and LUKS encrypted.

Because Disko automatically adds an `initrd.luks.devices` entry, I'm prompted for a passphrase on every boot. In reality, I don't need that drive to be unlocked at the initrd phase, and normally just have an `/etc/crypttab` file that specifies a key file. I'd like to be able to specify an encrypted drive that *doesn't* end up requiring unlocking during the initrd part of the boot.

Hopefully that makes sense? The inclusion of a `neededForBoot = false` type option could change the behaviour [here](https://github.com/nix-community/disko/blob/834b7b2a3310e19fca299df4b471f5750d485b11/types/luks.nix#L69) to not modify the initrd config?

yangm97:
Can't use @lilyinstarlight's strategy for zfs "native" mounts, gotta use a "legacy" mount instead.

jpds:
@yangm97 If you want to use ZFS native mounts, simply set: `boot.initrd.systemd.enable = true`.

iFreilicht:
@jnsgruk I agree with your reasoning. Because this flag would influence disko's behavior, it is valid to "duplicate" this option. An implementation of this feature would require the respective `fileSystems.<name>.neededForBoot` option to be set as well.

Note: other types might also have behavior that `neededForBoot = false;` should disable.

PRs are welcome.

PR #233, opened 2023-05-11
luks: add initrdUnlock option to luks type

jnsgruk:
I don't know whether this is a desired feature, so I've opened this PR mostly to stimulate a discussion. I'm happy to change the option name, or even just maintain a simple fork if this is undesirable.

Related to #192. Quoting my particular use case from that issue:

> I have a setup where I have a root disk that is just plain btrfs on a smaller SSD. I then have a data drive that is larger, and LUKS encrypted.
>
> Because Disko automatically adds an initrd.luks.devices entry, I'm prompted for a passphrase on every boot. In reality, I don't need that drive to be unlocked at the initrd phase, and normally just have an /etc/crypttab file that specifies a key file. I'd like to be able to specify an encrypted drive that doesn't end up requiring unlocking during the initrd part of the boot.

This PR introduces a new `configureInitrd` option to `luks` partitions, which when set explicitly to `false` prevents disko from modifying the initrd configuration, and therefore stops the system from trying to unlock the disk at early boot time.

This means that for non boot-critical drives, they can be lazily unlocked with a crypttab or other means later in the boot process.

### DIFF
--- a/types/luks.nix
+++ b/types/luks.nix
@@ -16,6 +16,11 @@
       description = "Path to the key for encryption";
       example = "/tmp/disk.key";
     };
+    initrdUnlock = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "Whether to add a boot.initrd.luks.devices entry for the specified disk.";
+    };
     extraFormatArgs = lib.mkOption {
       type = lib.types.listOf lib.types.str;
       default = [ ];
@@ -64,10 +69,13 @@
       internal = true;
       readOnly = true;
       default = dev:
-        [
-          # TODO do we need this always in initrd and only there?
-          { boot.initrd.luks.devices.${config.name}.device = dev; }
-        ] ++ (lib.optional (config.content != null) (config.content._config "/dev/mapper/${config.name}"));
+        [ ]
+        # If initrdUnlock is true, then add a device entry to the initrd.luks.devices config.
+        ++ (lib.optional (config.initrdUnlock)
+          [
+            { boot.initrd.luks.devices.${config.name}.device = dev; }
+          ])
+        ++ (lib.optional (config.content != null) (config.content._config "/dev/mapper/${config.name}"));
       description = "NixOS configuration";
     };
     _pkgs = lib.mkOption {
--- a/types/luks.nix
+++ b/types/luks.nix
@@ -68,13 +68,9 @@
     _config = lib.mkOption {
       internal = true;
       readOnly = true;
-      default = dev:
-        [ ]
+      default = dev: [ ]
         # If initrdUnlock is true, then add a device entry to the initrd.luks.devices config.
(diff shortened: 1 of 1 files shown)


jnsgruk:
(I should note - I've tested this on my machine, and it solves the problem for me at least!)

types/luks.nix
```diff
@@ -16,6 +16,11 @@
       description = "Path to the key for encryption";
       example = "/tmp/disk.key";
     };
+    configureInitrd = lib.mkOption {
```
Mic92, line 19: The option is not really self-describing. I wonder if 'unlockInInitrd' is semantically correct.

jnsgruk: Perhaps `enableInitrdUnlock`?

types/luks.nix
```diff
@@ -64,10 +69,13 @@
       internal = true;
       readOnly = true;
       default = dev:
-        [
-          # TODO do we need this always in initrd and only there?
-          { boot.initrd.luks.devices.${config.name}.device = dev; }
-        ] ++ (lib.optional (config.content != null) (config.content._config "/dev/mapper/${config.name}"));
+        [ ]
+        # If enableInitrdUnlock is true, then add a device entry to the initrd.luks.devices config.
+        ++ (lib.optional (config.enableInitrdUnlock == true)
```
Lassulus, line 74:
```suggestion
        ++ (lib.optional (config.enableInitrdUnlock)
```

types/luks.nix
```diff
@@ -16,6 +16,11 @@
       description = "Path to the key for encryption";
       example = "/tmp/disk.key";
     };
+    enableInitrdUnlock = lib.mkOption {
```
Lassulus, line 19: maybe we can call it just `initrdUnlock`?

jnsgruk: Done :)

types/luks.nix
jnsgruk: Done! :)

types/luks.nix
```diff
@@ -63,11 +68,10 @@
...
       internal = true;
       readOnly = true;
-      default = dev:
-        [
-          # TODO do we need this always in initrd and only there?
-          { boot.initrd.luks.devices.${config.name}.device = dev; }
-        ] ++ (lib.optional (config.content != null) (config.content._config "/dev/mapper/${config.name}"));
+      default = dev: [ ]
+        # If initrdUnlock is true, then add a device entry to the initrd.luks.devices config.
+        ++ (lib.optional (config.initrdUnlock) [{ boot.initrd.luks.devices.${config.name}.device = dev; }])
```
Lassulus, line 73:
```suggestion
        ++ (lib.optional config.initrdUnlock [{ boot.initrd.luks.devices.${config.name}.device = dev; }])
```

jnsgruk:
Committed suggestions directly on the assumption they'll be squashed on merge 😊

PR #238, opened 2023-05-18
zpool: deduplicate root mount option

SuperSandro2000:


types/zpool.nix
```diff
@@ -37,17 +37,6 @@
       default = [ "defaults" ];
       description = "Options to pass to mount";
     };
-    mountRoot = lib.mkOption {
```
Mic92, line 40: @Lassulus where we would add mkRemovedOptionModule if we wanted to add them for submodules?

Lassulus:
Do we want zfspools always to be at the root? What if people want to have tmpfs on /?

SuperSandro2000:
This should just be for mounting the zpool in a live iso where it must be mounted at /mnt otherwise it kills the running OS.

phaer:
> This should just be for relative mounting the zpool in a live iso where it must be mounted at /mnt otherwise it kills the running OS. It doesn't influence where the zpool is mounted in the resulting system.

But I couldn't build a system where I e.g. use `tmpfs` for `/mnt` while having a `zpool` at `/mnt/persist` with that?

Lassulus (approved):

Lassulus:
applied in https://github.com/nix-community/disko/pull/294, will be merged tomorrow or so. thanks!

PR #248, opened 2023-05-30
types btrfs: inline subvolumes

Lassulus:
inline the btrfs_subvolume type, since it's only used in btrfs
also remove the legacy index from the table type

the serializerLib can be used to get a json representation of the disko types. This can be used to render a webinterface. Output can be tested like this:
```
nix eval '.#lib.lib.jsonTypes' --json | jq .
```


Lassulus:
@dinvlad thanks for reporting the error, I created a fix PR in #252 . Could you test it please?

PR #249, opened 2023-05-30
Fix a potential error in the ZFS example

zah:
As far as I can tell, the example was wrong because it was copy/pasted from the first disk definition which also includes a `/boot` partition. Consuming the entire second disk is perfectly reasonable.

### DIFF
--- a/example/zfs.nix
+++ b/example/zfs.nix
@@ -41,7 +41,7 @@
           partitions = [
             {
               name = "zfs";
-              start = "128MiB";
+              start = "0";
               end = "100%";
               content = {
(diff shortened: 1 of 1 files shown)


phaer:
Not sure whether this should be regarded an "error". Best solution for a real-world setup would IMO be to have a copy of the ESP partition on both disks.

But even if not, you don't really gain anything by moving the offset as both disks are used for a mirror, which will only ever have the size of the smallest disk involved. So all other things being equal, I believe your change just moves 128MB of empty space from the beginning to (presumably) the end of the space.

PR #252, opened 2023-06-04
btrfs: create all subvolumes

Lassulus:
should fix https://github.com/nix-community/disko/pull/248#issuecomment-1575341938

I expected conatMapAttrs to merge deeply, sadly the test didn't catch that before since the test subvolume (because it's the last one) was accidentally created.


Lassulus:
bors merge

PR #256, opened 2023-06-07
introduce new table_gpt type

Lassulus:
we also add a new way to reference the parent node in the subtypes, this will be used in the future to simplify the code a bit (since we can pass device directly into an option instead of passing it to the _create, _mount and _config functions. this will also turn these into strings instead of functions, preparing the way for partial applications.

Also this new type uses partlabels instead of deviceNames, getting rid of the deviceNumbering logic we use in the other tables. This would also grant the ability to read the device during runtime from the environment.

### DIFF
--- a/example/simple-efi.nix
+++ b/example/simple-efi.nix
@@ -5,33 +5,27 @@
         device = builtins.elemAt disks 0;
         type = "disk";
         content = {
-          type = "table";
-          format = "gpt";
-          partitions = [
-            {
-              name = "ESP";
-              start = "1MiB";
-              end = "100MiB";
-              bootable = true;
+          type = "table_gpt";
+          partitions = {
+            ESP = {
+              type = "EF00";
+              start = "1M";
+              end = "+100M";
               content = {
                 type = "filesystem";
                 format = "vfat";
                 mountpoint = "/boot";
               };
-            }
-            {
-              name = "root";
-              start = "100MiB";
-              end = "100%";
-              part-type = "primary";
-              bootable = true;
+            };
+            root = {
+              end = "-0";
               content = {
                 type = "filesystem";
                 format = "ext4";
                 mountpoint = "/";
               };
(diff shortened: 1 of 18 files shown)


example/simple-efi.nix
```diff
@@ -5,33 +5,27 @@
...
         content = {
-          type = "table";
-          format = "gpt";
-          partitions = [
-            {
-              name = "ESP";
-              start = "1MiB";
-              end = "100MiB";
-              bootable = true;
+          type = "table_gpt";
```
Mic92, line 8:
```suggestion
          type = "gpt";
```

lib/types/table_gpt.nix
```diff
@@ -0,0 +1,122 @@
...
+        sgdisk --zap-all ${dev}
+        ${lib.concatStrings (lib.imap (index: partition: ''
+          sgdisk \
+            --new=${toString index}:${partition.start}:${partition.end} \
+            --change-name=${toString index}:${partition.label} \
+            --typecode=${toString index}:${partition.type} \
+            ${dev}
+          # ensure /dev/disk/by-path/..-partN exists before continuing
+          udevadm trigger --subsystem-match=block; udevadm settle
+          ls -la /dev/disk >&2
```
Mic92, line 78: Debug?

lib/types/table_gpt.nix
```diff
@@ -0,0 +1,122 @@
...
+          (index: partition:
+            lib.optionalAttrs (partition.content != null) (partition.content._meta dev)
+          )
+          (lib.attrValues config.partitions));
+      description = "Metadata";
+    };
+    _create = diskoLib.mkCreateOption {
+      inherit config options;
+      default = { dev }: ''
+        sgdisk --zap-all ${dev}
```
Mic92, line 69: Is this needed if we have disk-deativate?

PR #260, opened 2023-06-09
rely on partition.index

dguibert:
Without this patch, the index option of each partition set is ignored as the index from imap0 is used.

### DIFF
--- a/lib/types/table.nix
+++ b/lib/types/table.nix
@@ -84,14 +84,14 @@
           # ensure /dev/disk/by-path/..-partN exists before continuing
           udevadm trigger --subsystem-match=block; udevadm settle
           ${lib.optionalString partition.bootable ''
-            parted -s ${dev} -- set ${toString index} boot on
+            parted -s ${dev} -- set ${toString partition.index} boot on
           ''}
           ${lib.concatMapStringsSep "" (flag: ''
-            parted -s ${dev} -- set ${toString index} ${flag} on
+            parted -s ${dev} -- set ${toString partition.index} ${flag} on
           '') partition.flags}
           # ensure further operations can detect new partitions
           udevadm trigger --subsystem-match=block; udevadm settle
-          ${lib.optionalString (partition.content != null) (partition.content._create { dev = diskoLib.deviceNumbering dev index; })}
+          ${lib.optionalString (partition.content != null) (partition.content._create { dev = diskoLib.deviceNumbering dev partition.index; })}
         '') config.partitions)}
       '';
     };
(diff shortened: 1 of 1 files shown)


Mic92:
@Mergifyio rebase

Mic92:
@mergify unqueue

Mic92:
@mergify queue

Mic92:
@mergify rebase

Mic92:
@mergify queue

Mic92:
@mergify cancel

Mic92:
@mergify unqueue

Lassulus:
uh, the idea was to remove the index option from the table type and rely only on the order in which the partitions were defined in the list, since relying on the index can have weird implications (like creating partitions twice). The new gpt type has a priority field which gives more control if you want to define your partitions from multiple locations for example.

dguibert:
> uh, the idea was to remove the index option from the table type and rely only on the order in which the partitions were defined in the list, since relying on the index can have weird implications (like creating partitions twice). The new gpt type has a priority field which gives more control if you want to define your partitions from multiple locations for example.

Ok then 2 remarks:
- the index option for the partition type is defined, it can be set but its value is never used. This breaks the old behavior of disko and is also very usefull to import disk layout created without disko.
- If you remove the index option, the only way to configure disks with non contigous partition indices will be to reformat them?

Lassulus:
yeah I forgot to remove the index option. But I can reconsider, what is your usecase for non continuous partition numbers?

Lassulus:
also with the new gpt schema, you can use the index number as priority. If we want to restore the old behavior we also need to change the imap to a map.

dguibert:
> yeah I forgot to remove the index option. But I can reconsider, what is your usecase for non continuous partition numbers?

I've some quite old zfs configurations,  manually partitionned (mostly following https://openzfs.github.io/openzfs-docs/Getting%20Started/Ubuntu/Ubuntu%2018.04%20Root%20on%20ZFS.html#step-2-disk-formatting without the boot partition at index 3)

I've setup disko to handle fileSystems based on this setup and plan to update its config when a disk will fail.

But I've just learn that I could sort the partitions (https://superuser.com/a/429208), may give a try.

dguibert:
> yeah I forgot to remove the index option. But I can reconsider, what is your usecase for non continuous partition numbers?

Some old ZFS  configurations (following https://openzfs.github.io/openzfs-docs/Getting%20Started/Ubuntu/Ubuntu%2018.04%20Root%20on%20ZFS.html without a boot partition at index 2)

But I've resorted the GPT parititons with
- for all drives: $disk
````sh
sudo sgdisk -sp $disk
````
- remove partition indices
- rebuild system closure
- reboot

Everything is OK.
The index option could be removed.

PR #284, opened 2023-07-01
lib.types: turn _create, _mount and _config into values

Lassulus:
refactor some things, to make the code (hopefully) easier to understand.


Mic92:
@mergify queue

PR #285, opened 2023-07-03
lvm_vg: add lvm thinpool/thinlv support

ajs124:


lib/types/lvm_vg.nix
```diff
@@ -33,6 +38,16 @@
             default = [ ];
             description = "Extra arguments";
           };
+          pool = lib.mkOption {
```
Lassulus, line 41: isn't this the same as config.name? btw I refactored this mode a little bit in https://github.com/nix-community/disko/pull/284

ajs124: i don't think so. this is used for layered LVs. here I'm using it for a thin lv, on a thinpool (thinmeta+thindata LV). it could also be used for vdo, for example.

if you look at how it's used in the `lvcreate` invocation, this is passed to `--thinpool=`.

MatthewCash:
Any reason why this was closed? I've been running a rebased version with a few changes for a while and it works great.

ajs124:
I don't see me and/or @Lassulus finishing this in the foreseeable future (months/years), plus my interests and time investment in (projects like) this has declined since I opened this.

Lassulus:
I would be happy for a new PR and try to review it proper this time :D @MatthewCash

PR #287, opened 2023-07-07
Luks add additionalKeyFiles option

niki-on-github:
Currently disko allow only a single keyfile. I add an additional option to specify more key files. 

Hint: Separate option of the first key has the benefit to get the possible of passphrase + keyfile for an luks volume.

### DIFF
--- a/example/complex.nix
+++ b/example/complex.nix
@@ -32,6 +32,7 @@
                 type = "luks";
                 name = "crypted1";
                 settings.keyFile = "/tmp/secret.key";
+                additionalKeyFiles = ["/tmp/additionalSecret.key"];
                 extraFormatArgs = [
                   "--iter-time 1" # unsecure but fast for tests
                 ];
--- a/example/luks-lvm.nix
+++ b/example/luks-lvm.nix
@@ -33,6 +33,7 @@
                 # if you want to use the key for interactive login be sure there is no trailing newline
                 # for example use `echo -n "password" > /tmp/secret.key`
                 settings.keyFile = "/tmp/secret.key";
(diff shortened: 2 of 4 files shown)


Mic92 (approved):
Looks good maybe even our example could use this option to ensure it will not regress.

Lassulus:
Maybe we turn keyFile into a list?

niki-on-github:
> Maybe we turn keyFile into a list?

This would look nicer from the api point but doesn't allow to control if you want to store a passphrase manually as the first key. (You can of course write your passphrase into a temporary keyfile - but this require additional logic outside of disko to deploy the partition layout).

If desired I can look at how it would be possible to put everything in one list. Maybe by using `imap0` to handle index 0 different.

Lassulus:
lib.head would come to mind to handle the first key differently. But I don't want to stray to far from the nixpkgs config, so I guess this solution is actually fine

Lassulus:
@mergify queue

Mic92:
@mergify unqueue

Mic92:
@mergify queue

Mic92:
@mergifyio refresh

Mic92:
@mergify queue

Mic92:
@Mergifyio refresh

Lassulus:
@mergify queue

Lassulus:
this needs a rebase now after we merged the other luks related PRs

niki-on-github:
rebased now

Mic92:
@mergify requeue

Mic92:
@mergify queue

PR #291, opened 2023-07-10
feat(luks): Add settings submodule

rogarb:
This mirrors the NixOS boot.initrd.luks.devices.<name>.keyFileOffset and boot.initrd.luks.devices.<name>.keyFileSize options.

### DIFF
--- a/example/complex.nix
+++ b/example/complex.nix
@@ -31,9 +31,9 @@
               content = {
                 type = "luks";
                 name = "crypted1";
-                keyFile = "/tmp/secret.key";
+                settings.keyFile = "/tmp/secret.key";
                 extraFormatArgs = [
-                  "--iter-time 1"
+                  "--iter-time 1" # unsecure but fast for tests
                 ];
                 content = {
                   type = "lvm_pv";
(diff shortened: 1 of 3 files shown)


Lassulus:
How about we add a settings key to the luks type which would be passed to boot.initrd.luks.devices.${config.name} ? Not sure if we want to replicate all the options in disko.

Lassulus:
we could still get them from settings.keyFileSize in the create/mount script? we would get the type checks only when evaluating the config though. but that seems only important for people not using the config output of disko? 

So I'm not sure, I guess I would prefer having all the settings of nixpkgs inside one attribute?

rogarb:
Sorry I got confused about your proposal. I think that I have understood what you mean now, I think it is definitely a good idea. I will try to come up with an implementation.

rogarb:
@Lassulus it seems that I can't get to implement the `settings` option as a submodule properly, would you have any suggestion for a proper way to do it?

Lassulus:
ah yeah, I will push to your branch shortly. basically you can skip defining a type since we silently reuse the nixpkgs type.

Lassulus:
for backwards compatibility I suggest we keep keyFile in the luks type and trace a deprecation warning when it is used.

rogarb:
I added back the keyFile option with a warning message. I guess that the PR is now ready to be merged.

Lassulus:
@mergify queue

Issue #288, opened 2023-07-09
Support for BTRFS subvolumes to be created but not mounted


rogarb:
Hi,

I have a slightly more complex BTRFS layout than the classical "flat" one:
```
$ sudo btrfs subvolume list -t -a /mnt
ID	gen	top level	path	
--	---	---------	----	
256	1051	5		DATA
257	11	5		SYSTEM
258	15	5		VOLATILE
259	924	258		<FS_TREE>/VOLATILE/cache
260	1637	257		<FS_TREE>/SYSTEM/nix
261	1054	256		<FS_TREE>/DATA/nixos
262	14	256		<FS_TREE>/DATA/home
263	1901	262		<FS_TREE>/DATA/home/user
264	1635	259		<FS_TREE>/VOLATILE/cache/user
265	1836	256		<FS_TREE>/DATA/nixos-config
```
In this layout, the subvolumes `DATA`, `SYSTEM` and `VOLATILE` are not meant to be mounted but rather act as upper level "hidden folder" for other subvolumes.

I hoped to be able to reproduce this layout by declaring such "hidden suvolumes" with:
```
[...]
content = {
  type = "btrfs";
  [...]
  subvolumes = {
    "DATA" = { };
    "SYSTEM" = { };
    "VOLATILE" = { };
    [...]
  };
};
```
as I could read in the btrfs related example file, but this leads to having those filesystems mounted as `/<NAME>`, which I don't want.
They need to be declared though in order to create the subvolumes they contain, otherwise `disko` raises an error.

I came up with [a solution](https://github.com/rogarb/disko/tree/btrfs-fix) by adding a `dontMount` option for the subvolumes, and I modified `_mount` accordingly.

Is that a good solution? If so I shall open a PR.

cvoges12:
@rogarb Shouldn't this issue be closed as the solution has already been merged?

rogarb:
> @rogarb Shouldn't this issue be closed as the solution has already been merged?

Yes sorry for the delay.

PR #292, opened 2023-07-13
feat(btrfs): support unmounted subvolumes

rogarb:
This option allows the creation of subvolumes which won't be mounted, allowing to create more complex btrfs layouts.

This is a personal attempt to fix the issue #288, since it didn't seem to get much attention.

### DIFF
--- a/docs/upgrade-guide.md
+++ b/docs/upgrade-guide.md
@@ -1,3 +1,56 @@
+# 2023-07-09 121df48
+
+Changes:
+ - BTRFS subvolumes are mounted if and only their `mountpoint` is set.
+
+Especially, if you have specific mount options for a subvolume (through `mountOptions`), make sure to set `mountpoint` otherwise the subvolume will not be mounted.
+
+This change allows more flexibility when using BTRFS, giving access to its volume management functionality.
+
+It allows layouts as the following:
+```nix
+content = {
+  type = "btrfs";
+  # BTRFS partition is not mounted as it doesn't set a mountpoint explicitly
+  subvolumes = {
+    # This subvolume will not be mounted
+    "SYSTEM" = { };
+    # mounted as "/"
+    "SYSTEM/rootfs" = {
+      mountpoint = "/";
+    };
+    # mounted as "/nix"
+    "SYSTEM/nix" = {
+      mountOptions = [ "compress=zstd" "noatime" ];
+      mountpoint = "/nix";
+    };
+    # This subvolume will not be mounted
+    "DATA" = { };
+    # mounted as "/home"
+    "DATA/home" = {
+      mountOptions = [ "compress=zstd" ];
+      mountpoint = "/home";
+    };
+    # mounted as "/var/www"
+    "DATA/www" = {
+      mountpoint = "/var/www";
+    };
+  };
+};
+```
+corresponding to the following BTRFS layout:
+```
+BTRFS partition # not mounted
+ |
+ |-SYSTEM       # not mounted
+ |  |-rootfs    # mounted as "/"
+ |  |-nix       # mounted as "/nix"
+ |
+ |-DATA         # not mounted
+    |-home      # mounted as "/home"
+    |-www       # mounted as "/var/www"
+```
+
 # 2023-04-07 7d70009
 
 Changes:
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -91,16 +91,10 @@
         let
           subvolMounts = lib.concatMapAttrs
             (_: subvol:
-              let
-                mountpoint =
-                  if (subvol.mountpoint != null) then subvol.mountpoint
-                  else if (config.mountpoint == null) then subvol.name
-                  else null;
-              in
-              lib.optionalAttrs (mountpoint != null) {
-                ${mountpoint} = ''
-                  if ! findmnt ${config.device} "${rootMountPoint}${mountpoint}" > /dev/null 2>&1; then
-                    mount ${config.device} "${rootMountPoint}${mountpoint}" \
+              lib.optionalAttrs (subvol.mountpoint != null) {
+                ${subvol.mountpoint} = ''
+                  if ! findmnt ${config.device} "${rootMountPoint}${subvol.mountpoint}" > /dev/null 2>&1; then
+                    mount ${config.device} "${rootMountPoint}${subvol.mountpoint}" \
                     ${lib.concatMapStringsSep " " (opt: "-o ${opt}") (subvol.mountOptions ++ [ "subvol=${subvol.name}" ])} \
                     -o X-mount.mkdir
                   fi
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -91,6 +91,8 @@
         let
           subvolMounts = lib.concatMapAttrs
             (_: subvol:
+            lib.warnIf (subvol.mountOptions != (options.subvolumes.type.getSubOptions []).mountOptions.default && subvol.mountpoint == null)
+              "Subvolume ${subvol.name} has mountOptions but no mountpoint. See upgrade guide (2023-07-09 121df48)."
               lib.optionalAttrs (subvol.mountpoint != null) {
                 ${subvol.mountpoint} = ''
                   if ! findmnt ${config.device} "${rootMountPoint}${subvol.mountpoint}" > /dev/null 2>&1; then
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -123,14 +123,8 @@
       default = [
         (map
           (subvol:
-            let
-              mountpoint =
-                if (subvol.mountpoint != null) then subvol.mountpoint
-                else if (config.mountpoint == null) then subvol.name
-                else null;
-            in
-            lib.optional (mountpoint != null) {
-              fileSystems.${mountpoint} = {
+            lib.optional (subvol.mountpoint != null) {
+              fileSystems.${subvol.mountpoint} = {
                 device = config.device;
                 fsType = "btrfs";
                 options = subvol.mountOptions ++ [ "subvol=${subvol.name}" ];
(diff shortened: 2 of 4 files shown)


lilyinstarlight:
I think it might be better if we reconsider whether to automatically create mounts when the parent btrfs does not have a `mountpoint` and subvolumes also do not `mountpoint` rather than a one-off option on each subvol for this. I kinda regret a bit making it default to adding mountpoints in my initial btrfs subvol PR for that case 🙃

Maybe we can add a btrfs option like `mountSubvolumes` that defaults to `true` but allows disabling that behavior, and then do `config.mountpoint == null && config.mountSubvolumes` in the conditional? Or maybe it would be better to make a backwards incompatible change now rather than later and just force people to add `mountpoint` to each subvolume rather than letting it be implicit if the btrfs filesystem root itself isn't mounted

Do you have thoughts @Lassulus?

Lassulus:
how about we set the default of mountpoint to config.name? people can then set it to null if they don't want it mounted, that should simplify the code also in general

rogarb:
> how about we set the default of mountpoint to config.name? people can then set it to null if they don't want it mounted, that should simplify the code also in general

Great, I will have a look and try to implement that.

rogarb:
This is my attempt to solve the problem, I commented the code to explain assumptions, it is now waiting for review.

Lassulus:
I don't really understand the usecase why we want to automatically mount the subvolumes if the btrfs mainvolume has not mountpoint? 

wouldn't be a default in the subvolume like this better?:
```
 mountpoint = lib.mkOption {
  type = lib.types.nullOr diskoLib.optionTypes.absolute-pathname;
  default = if config.mountpoint == null then null else "${config.mountpoint}/${config.name}";
  description = "Location to mount the subvolume to (set to null if you don't want the subvolume to be mounted).";
};
```

we have to change the args in line 25 a bit, and change it basically to `{ name, ...}@subvol`

lilyinstarlight:
> I don't really understand the usecase why we want to automatically mount the subvolumes if the btrfs mainvolume has not mountpoint?

Yeah that's what I was saying, I regret doing that and would not mind if we just reversed that default entirely

> wouldn't be a default in the subvolume like this better?:

No because that subvolume wouldn't need to be mounted at all. Subvolumes only need to be mounted if their path would be different (e.g. mounting a `/` subvol if it is different from the filesystem's root subvol and then mounting each other subvolume where it goes within the `/` subvol, for my case)

Lassulus:
I'm not sure I understand fully. My idea was: if the subvolume.mountpoint has a value of null we skip mounting it, and the default value is just the root mountpoint + the name. if the root mountpoint has null as a value we skip mounting the subvolume by default.

lilyinstarlight:
> My idea was: if the subvolume.mountpoint has a value of null we skip mounting it, and the default value is just the root mountpoint + the name.

Yeah if you `btrfs subvolume create /mountpoint/subvolname` then you don't need to mount it at `/mountpoint/subvolname`. It already exists there

We should just skip mounting subvols by default entirely if we do not want to do it for the case that the root filesystem is not mounted

rogarb:
> > My idea was: if the subvolume.mountpoint has a value of null we skip mounting it, and the default value is just the root mountpoint + the name.
> 
> Yeah if you `btrfs subvolume create /mountpoint/subvolname` then you don't need to mount it at `/mountpoint/subvolname`. It already exists there
> 
> We should just skip mounting subvols by default entirely if we do not want to do it for the case that the root filesystem is not mounted

So basically, we should only mount {main,sub}volumes if and only they have a `mountpoint` value, right? Since any "sub-subvolume" will appear automatically inside their mounted parent. In that case, it is even better to have  a default value of `null` for the subvolume mountpoints.

I think that it's a bit different from the ZFS behaviour, which is what might have influenced my thoughts when considering use cases.

rogarb:
> I don't really understand the usecase why we want to automatically mount the subvolumes if the btrfs mainvolume has not mountpoint?

Well in my case, I don't mount the main volume at all and I mount different subvolumes to different places (e.g. `SYSTEM/nixos` to `/nix`, `DATA/nixos-config` to `/etc/nixos`, `DATA/home` to `/home`...). You can think about it as a LVM-like use case.

rogarb:
This simple last patch should behave as stated. 

The only case I can think of where you want to actually act on mounting and don't specify a different mountpoint is when you have specific mount options for a subvolume. Maybe we should check `subvol.mountpoint != null || subvol.mountOptions != []`?

Lassulus:
as far as I understood things, this seems correct. waiting for an approval from @lilyinstarlight and will merge then :)

lilyinstarlight (commented):
The change looks good to me and I just successfully tested it with my config by adding explicit mountpoints for my subvolumes

Can you update the `examples/btrfs-subvolumes.nix` file to have explicit mountpoints and add a note to `docs/upgrade-guide.md` that subvolume mounts when the parent filesystem has no mount points need to be explicit now?

Thank you for this contribution!

rogarb:
Le 2023-07-14 16:48, Lily Foster a écrit :
> @lilyinstarlight commented on this pull request.
> 
> The change looks good to me and I just successfully tested it with my
> config by adding explicit mountpoints for my subvolumes
> 
> Can you update the examples/btrfs-subvolumes.nix file to have explicit
> mountpoints and add a note to docs/upgrade-guide.md that subvolume
> mounts when the parent filesystem has no mount points need to be
> explicit now?
> 

I will be AFK for a few days, I will take care of all this when I come 
back.

rogarb:
> Can you update the `examples/btrfs-subvolumes.nix` file to have explicit mountpoints and add a note to `docs/upgrade-guide.md` that subvolume mounts when the parent filesystem has no mount points need to be explicit now?

I just added two commits addressing these 2 points. Please have a look and let me know if anything has to be changed.

I also thought about a corner case: if a subvolume has `mountOptions` and no `mountpoint`, **and** its parent is mounted, then it will appear in the filesystem hierarchy but it will not be mounted with the specified options.
In order to address that (if there is something to address which I am not sure), we could check for `mountOptions != []` in `_mount`, but in that case, we should also derive a mountpoint using the mounted parent's mountpoint.
My opinion is that for such case, `mountpoint` should be also specified, and thus no change in code would be needed. In any case, there could be somewhere a warning telling that a subvolume has `mountOptions` without a mountpoint, so the options will not be taken into account.

lilyinstarlight:
> My opinion is that for such a case, `mountpoint` should also be specified, and thus no change in code would be needed.
> 
> In any case, there could be somewhere a warning when a subvolume has `mountOptions` without a`mountpoint`, so the options will not be taken into account.

Yeah, I would agree that `mountpoint` should be specified in that case. Feel free to add a warning if you'd like

I'll review this one more time today or tomorrow, thank you!

rogarb:
> > My opinion is that for such a case, `mountpoint` should also be specified, and thus no change in code would be needed.
> > In any case, there could be somewhere a warning when a subvolume has `mountOptions` without a`mountpoint`, so the options will not be taken into account.
> 
> Yeah, I would agree that `mountpoint` should be specified in that case. Feel free to add a warning if you'd like

I added a new commit with a warning (and rebased to current main branch).

> I'll review this one more time today or tomorrow, thank you!

Great, thank you!

lib/types/btrfs.nix
```diff
@@ -91,6 +91,8 @@
         let
           subvolMounts = lib.concatMapAttrs
             (_: subvol:
+            lib.warnIf (subvol.mountOptions != [] && subvol.mountpoint == null)
```
rogarb, line 94: This actually triggers the warning all the time since mountOptions.default is [ "defaults" ] and not the empty list... 
Is there a way to get the corresponding Option type default value for the comparison here?

lilyinstarlight: `options.subvolumes.[name].default` should get it, I think. But it would probably be better to just check if it's set by `options.subvolumes.[name].highestPrio < (mkOptionDefault {}).priority`. That will tell if it's set at all aside from the option's default value, even if the value is set to what happens to be the default

Lassulus: Uh I think checking against the default value would be better. Since we reeval the config in the tests and then we could run in weird behavior

lilyinstarlight: What weird behavior?

I'm not sure I am able to come up with something weird that would happen (re-eval doesn't matter, this just checks what the highest option priority is set at all which only changes when someone sets it either without a priority or with a higher priority manually)

Fwiw this is how checking whether an option changed from the default is done in nixpkgs a few places

Lassulus: well I pass the whole config.disko.devices into config.disko.devices again (of the test target system) so everything would be set to the default value without it still being the default value.

lilyinstarlight: Yeah but that won't change the priority value, right? (from the special type `{ _type = "override"; priority = ...; content = ...; }`)

lilyinstarlight: Oh you mean the merged config values. Huh, yeah I guess that could do something weird. Yeah I guess comparing against the default is fine then

rogarb: I found a way to access the default value using `(options.subvolumes.type.getSubOptions []).mountOptions.default`, if you think there is a simpler way, please let me know.

lilyinstarlight (approved):
This diff looks good to me now!

It might be good for @Lassulus to do another sanity check on it, but I don't see any problems and my disko script actually comes out the same before and after this PR now that I added `mountpoint`s to the subvolumes

docs/upgrade-guide.md
```diff
@@ -1,3 +1,56 @@
...
+    # This subvolume will not be mounted
+    "DATA" = { };
+    # mounted as "/home"
+    "DATA/home" = {
+      mountOptions = [ "compress=zstd" ];
+      mountpoint = "/home";
+    };
+    # mounted as "/var/www"
+    "DATA/www" = {
+      mountpount = "/var/www";
```
tiredofit, line 36: I believe you will want `mountpoint` instead of `mountpount`

rogarb: Thank you for spotting this! I just pushed a fixed revision!

Lassulus (approved):

PR #294, opened 2023-07-13
lib: refactor actually a lot of stuff (sorry mic)

Lassulus:


Lassulus:
not sure if the cli is still working, since we have no tests for it :(

README.md
```diff
@@ -85,7 +85,7 @@ A simple disko configuration may look like this:
 If you'd saved this configuration in /tmp/disko-config.nix, and wanted to create a disk named /dev/nvme0n1, you would run the following command to partition, format and mount the disk.
 
 ```
-$ sudo nix run github:nix-community/disko -- --mode zap_create_mount /tmp/disko-config.nix --arg disks '[ "/dev/nvme0n1" ]'
+$ sudo nix run github:nix-community/disko -- --mode disko /tmp/disko-config.nix --arg disks '[ "/dev/nvme0n1" ]'
```
Mic92, line 88: Do we have some documentation what each of these modes do?
It's not exactly intuitive from the name.

default.nix
```diff
@@ -10,44 +10,35 @@ let
...
+
+  format = cfg: (eval cfg).config.disko.devices._create;
+  formatScript = cfg: pkgs: ((eval cfg).config.disko.devices._scripts { inherit pkgs checked; }).formatScript;
+  formatScriptNoDeps = cfg: pkgs: ((eval cfg).config.disko.devices._scripts { inherit pkgs checked; }).formatScriptNoDeps;
+
+  mount = cfg: (eval cfg).config.disko.devices._mount;
+  mountScript = cfg: pkgs: ((eval cfg).config.disko.devices._scripts { inherit pkgs checked; }).mountScript;
+  mountScriptNoDeps = cfg: pkgs: ((eval cfg).config.disko.devices._scripts { inherit pkgs checked; }).mountScriptNoDeps;
+
+  disko = cfg: (eval cfg).config.disko.devices._disko;
```
Mic92, line 35: so disko is create+format+mount?

disko
```diff
@@ -18,7 +18,7 @@ or $0 [options] --flake github:somebody/somewhere
 Options:
 
 * -m, --mode mode
-  set the mode, either create, mount, zap_create_mount or disko
+  set the mode, either format, mount or disko
```
Mic92, line 21: Here it also should say what disko mode does.

lib/default.nix
```diff
@@ -375,8 +351,149 @@ let
...
+            system.build.mountScriptNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "disko-mount" ''
+              ${cfg.config._mount}
+            '';
+
+            diskoScriptNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "disko" ''
+              ${cfg.config._disko}
+            '';
+          };
+        };
+        _umount = lib.mkOption {
```
Mic92, line 419: Would it not make sense to make this more scarier?

```suggestion
        _umountAndDestroy = lib.mkOption {
```

Otherwise I would actually call it _destroy instead of _umount.

lib/types/mdadm.nix
```diff
@@ -44,6 +44,8 @@
           "''${disk_devices[@]}"
         udevadm trigger --subsystem-match=block
         udevadm settle
+        # for some reason mdadm devices spwan with an existing partition table, so we need to wipe it
```
Mic92, line 47:
```suggestion
        # for some reason mdadm devices spawn with an existing partition table, so we need to wipe it
```

default.nix
Lassulus: create is an alias for format. disko is umount + format + mount

lib/default.nix
Lassulus: ah, I wanted to change it to just umount, the destroying is basically an undesired side effect :) but for now it would make sense yes

tests/zfs-over-legacy.nix
```diff
@@ -1,7 +1,8 @@
 { pkgs ? (import <nixpkgs> { })
-, makeDiskoTest ? (pkgs.callPackage ./lib.nix { }).makeDiskoTest
+, makeDiskoTest ? (pkgs.callPackage ../lib { }).testLib.makeDiskoTest
 }:
 makeDiskoTest {
+  inherit pkgs;
```
Mic92, line 5: We did we need passing pkgs here instead of passing it in callPackage already?

Lassulus: lib can't depend on pkgs, because of inifinite recursion

default.nix
Mic92: As I just learned it's umount + destroy + format + mount

lilyinstarlight (commented):
Do we intend to continue supporting `{ type = "table"; format = "gpt"; ... }` in addition to `{ type = "gpt"; ... }`?

This removes that former construction from all tests so it is now untested on this branch from what I can tell

Lassulus:
> Do we intend to continue supporting `{ type = "table"; format = "gpt"; ... }` in addition to `{ type = "gpt"; ... }`?
> 
> This removes that former construction from all tests so it is now untested on this branch from what I can tell

not really. But I'm not sure if type = "gpt" breaks in some scenarios

Mic92:
> Do we intend to continue supporting `{ type = "table"; format = "gpt"; ... }` in addition to `{ type = "gpt"; ... }`?
> 
> This removes that former construction from all tests so it is now untested on this branch from what I can tell

We will probably encourage using the new model but not remove it since peoples system already depend on it. But yeah, we should probably keep some regression test to make sure it won't stop working.

README.md
Lassulus: added some documentation to the help output of the disko cli

Lassulus:
added a legacy table test, renamed _umount to _destroy.

Lassulus:
ah forget the test, I also included some other stale PRs which I took a look at

Lassulus:
[`6068287` (#294)](https://github.com/nix-community/disko/pull/294/commits/6068287a8e072e3a06dd950ab09b2829d6653017) could be a breaking change for nixos-anywhere, hmm. not sure if we should include it here.

Lassulus:
https://github.com/nix-community/disko/pull/294/commits/4b9126c4dde5305d6b5950775ec06f33130e5b5e

Lassulus:
Basically what https://github.com/nix-community/disko/pull/78 does

Lassulus:
ok dropped the commit for now, we need better migrations for that one

Mic92:
@mergify queue

PR #299, opened 2023-07-20
allow using the disko test framework via the module

Lassulus:


Lassulus:
ah doesn't work in pure mode yet

lib/types/gpt.nix
```diff
@@ -129,11 +129,14 @@ in
...
       internal = true;
       readOnly = true;
-      default = map
+      default = (map
           (partition:
             lib.optional (partition.content != null) partition.content._config
           )
-          (lib.attrValues config.partitions);
+          (lib.attrValues config.partitions))
+          ++ (lib.optional (lib.any (part: part.type == "EF02") (lib.attrValues config.partitions)) {
```
Mic92, line 137: We should mention this in our documentation that one should use EF02 to create a partition for grub.

lib/tests.nix
```diff
@@ -100,7 +100,9 @@ let
             echo -n 'secretsecret' > /tmp/secret.key
           '';
           boot.consoleLogLevel = lib.mkForce 100;
+          boot.loader.systemd-boot.enable = lib.mkDefault efi;
```
Mic92, line 102: Do we use efi by default?

tests/boot-raid1.nix
```diff
@@ -9,4 +9,8 @@ makeDiskoTest {
     machine.succeed("test -b /dev/md/boot");
     machine.succeed("mountpoint /boot");
   '';
+  extraSystemConfig = {
+    # sadly systemd-boot fails to install to a raid /boot device
+    boot.loader.systemd-boot.enable = false;
```
Mic92, line 14: Yes. This requires a dedicated non-mirrored efi partition.

lib/tests.nix
Lassulus: Yes

tests/boot-raid1.nix
Lassulus: but this seems like a bug in systemd-boot. since it works fine in grub. maybe something with our install script. seems like systemd-boot does some unneccessary checks for the underlying FS of the /boot partition, so maybe we can skip those

lib/types/gpt.nix
Lassulus: but where?

Mic92: https://github.com/nix-community/disko/blob/master/docs/quickstart.md

Mic92 (approved):

Mic92:
@mergify queue

PR #301, opened 2023-07-21
fix(luks): type error with keyfile size/offset

rogarb:
Using `keyFileOffset`/`keyFileSize` leads to a `cannot coerce an integer to a string` error. Since they are used to generate the `cryptsetup` command in a script as text, they need to be converted to a string.


Lassulus:
Uh yeah. With a test we would have catched that earlier :)

rogarb:
Should I submit a separate PR with a test for this feature or add it to the example?

Lassulus:
probably a new test would be cool?

PR #302, opened 2023-07-23
docs(luks): add an example using keyFile options

rogarb:
This adds `keyFileOffset` and `keyFileSize` options in `example/complex.nix`. It follows the discussion in #301.

I thought that adding a new example file for these options was a lot of noise for little result. 

@Lassulus if you meant something else with "a new test", please put me on tracks, as I quickly checked the `tests` folder and couldn't think about something relevant I could add there.

### DIFF
--- a/example/complex.nix
+++ b/example/complex.nix
@@ -58,7 +58,11 @@
               content = {
                 type = "luks";
                 name = "crypted2";
-                settings.keyFile = "/tmp/secret.key";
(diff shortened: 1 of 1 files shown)


rogarb:
Mmm it seems that I asked too many bytes to read from the keyfile. I should have checked `lib/tests.nix` more thoroughly. It should work now.

PR #304, opened 2023-07-25
luks: utilize extraOpenArgs during partition mount; clarify how the extraOpenArgs option works

ElXreno:


iFreilicht:
Thank you for your contribution! However, the functional issue this PR addresses has been fixed in #488 by now. Sorry we didn't get around to this in time.

PR #307, opened 2023-07-27
add image builder based on disko

Lassulus:
like make-disk-image.nix from nixpkgs, but cooler :)
(multi image support, takes disko config, etc.)

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -10,6 +10,9 @@ let
 
   diskoLib = {
 
+    # like make-disk-image.nix from nixpkgs, but with disko config
+    makeDiskImage = args: import ./make-disk-image.nix ({ inherit diskoLib; } // args);
+
     testLib = import ./tests.nix { inherit lib makeTest eval-config; };
     # like lib.types.oneOf but instead of a list takes an attrset
     # uses the field "type" to find the correct type in the attrset
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -3,6 +3,7 @@
 , pkgs ? nixosConfig.pkgs
 , lib ? pkgs.lib
 , name ? "${nixosConfig.config.networking.hostName}-disko-images"
+, extraPostVM ? ""
 }:
 let
   cleanedConfig = diskoLib.testLib.prepareDiskoConfig nixosConfig.config diskoLib.testLib.devices;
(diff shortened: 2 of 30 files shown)


Lassulus:
it can be used like this: https://github.com/Lassulus/flakes-testing/blob/master/flake.nix#L11

lib/default.nix
```diff
@@ -10,6 +10,9 @@ let
 
   diskoLib = {
 
+    # like make-disk-image.nix from nixpkgs, but with disko config
+    makeDiskImage = args: import ./make-disk-image.nix ({ inherit diskoLib; } // args);
```
Mic92, line 14: Can we also add a test or example in our flake?

lib/make-disk-image.nix
```diff
@@ -0,0 +1,56 @@
...
+pkgs.vmTools.runInLinuxVM (pkgs.runCommand name {
+  buildInputs = with pkgs; [
+    systemdMinimal
+    nixos-install-tools
+    nix
+    utillinux
+  ];
+  preVM = ''
+    # TODO: get size either dynamically or from disko config
+    ${lib.concatMapStringsSep "\n" (disk: "truncate -s 15G ${disk.name}.raw") (lib.attrValues nixosConfig.config.disko.devices.disk)}
```
Mic92, line 25: Should we make this configurable at least?

lib/make-disk-image.nix
```diff
@@ -0,0 +1,56 @@
...
+  QEMU_OPTS = lib.concatMapStringsSep " " (disk: "-drive file=${disk.name}.raw,if=virtio,cache=unsafe,werror=report") (lib.attrValues nixosConfig.config.disko.devices.disk);
+
+  memSize = 1024;
+} ''
+  # running udev, stolen from stage-1.sh
+  echo "running udev..."
+  ln -sfn /proc/self/fd /dev/fd
+  ln -sfn /proc/self/fd/0 /dev/stdin
+  ln -sfn /proc/self/fd/1 /dev/stdout
+  ln -sfn /proc/self/fd/2 /dev/stderr
```
Mic92, line 40: Do we have other common devices such as /dev/null?

lib/make-disk-image.nix
```diff
@@ -0,0 +1,56 @@
...
+    }];
+  };
+in
+pkgs.vmTools.runInLinuxVM (pkgs.runCommand name {
+  buildInputs = with pkgs; [
+    systemdMinimal
+    nixos-install-tools
+    nix
+    utillinux
+  ];
```
Mic92, line 22: When people have common hooks, we probably should expose an extraPackages parameter?

Mic92: Maybe not needed.

lib/make-disk-image.nix
Lassulus: dunno, this is taken from nixos stage-1

Mic92:
@mergify queue

PR #313, opened 2023-08-06
disko: drop binPrefix when building destroyScript

Mic92:
This seems to cause eval errors and I also don't know where this is supposed to come from


Lassulus:
@mergify queue

PR #318, opened 2023-08-10
Added missing dependency to destroyScript

ktrinh-anduril:
This PR adds 2 missing deps for destroyScript:
- `jq`: Unlike other scripts, destroyScript wasn't pulling in `${lib.makeBinPath (cfg.config._packages pkgs)`
- `bash`: destroyScript already uses bash for its shebang but it looks like wipefs also need bash in the PATH

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -386,7 +386,8 @@ let
                 zfs
                 lvm2
                 bash
-              ])}:${lib.makeBinPath (cfg.config._packages pkgs)}:$PATH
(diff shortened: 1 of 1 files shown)


Lassulus:
destroyScript shouldn't rely on the generated packageset since it's not reliant on the disko config. better would be to put the jq binary directly there

ktrinh-anduril:
Gotcha, added jq directly instead

PR #319, opened 2023-08-11
silence make-disk-image tests

Mic92:


Lassulus (commented):
We cant drop that yet or we will crash on 23.05

lib/tests.nix
```diff
@@ -173,7 +173,7 @@ let
             })
           ];
 
-          boot.initrd.services.swraid.enable = true;
+          boot.swraid.enable = true;
```
Lassulus, line 176: This also needs to be wrapped like in mdadm.nix. since we export the rest function in the module

Mic92: Done.

Lassulus:
@mergify queue

PR #329, opened 2023-08-24
mdadm: insert grub module for mdraid

Mic92:
the condition is not perfect yet. I think one could also have no /boot and just / and grub might be also installed at a different mountpoint than /boot?


Mic92:
I woudn't use raid anymore on efi systems because it breaks randomly. mirrorBoots is better.

Issue #130, opened 2023-01-27
Spaces in GPT partition names not properly escaped and general question on escaping for generated scripts


cscutcher:
Noticed during some experimentation that spaces in GPT partition names cause problems. Irritating I've lost the exact error output but in the case of a GPT partition name `Basic Data Partition`, parted complains that `Data` isn't a valid fs-type i.e. parted interprets the partition name as `Basic` and then treats `Data` and `Partition` as the subsequent arguments. Even escaping the name with `strings.escapeShellArg` isn't sufficient.

I note that throughout the code, there appears to be little implicit escaping of strings (via `strings.escapeShellArg` or similar) which might be an intentional decision to keep things explicit, or more likely just something that's not done yet? 

In either case, arguments to parted `mkpart` are a bit more awkward since even if the user thinks to wrap name in `escapeShellArg` this isn't sufficient, since bash strips the quotes before the command is interpreted by `parted`. For example [this stackoverflow](https://superuser.com/a/1531018) shows an example of the required quoting;

```console
$ sudo parted /dev/sdb mkpart '"Name of the partition"' 0% 100%
```

I'm trying to work around this on the config side by wrapping the name arguments with this;

```nix
escape_parted_name = inp: l.strings.escapeShellArg "\"${inp}\"";
```

Which I was tempted to submit as a fix targeted specifically at the partition script generation

https://github.com/nix-community/disko/blob/d7e178126f05d9b1e5bfbf115c36ec9f75b85e8b/types.nix#L770

But then it occurred to me that would introduce an inconsistency, with this one argument being implicitly escaped but the majority of other arguments to parted, and in other generated scripts (names, options etc), being left raw. 

Any thoughts to how you guys would like to see this handled? Any existing plans to clean up some of the code around script generation and argument handling?

cscutcher:
Oh man, how have I never come across `toGNUCommandLineShell` before! Nice! 

> Maybe all those commands should just use `lib.toGNUCommandLineShell` for proper escaping.

Yeah, I think that's a good call, given the amount of script generation going on. Although I think it'd still need special handling for the case of `parted` to ensure that quotes remain when parted parses the inner command (i.e. they're not all stripped by bash).

cscutcher:
I'm up for having a swing at this myself, but might be a while since not much free time to code at the moment.

Lassulus:
not sure if thats still a problem? maybe PRing an example which causes the break is enough so I can fix it

cscutcher:
[...]

PR #330, opened 2023-08-25
tests: Add failing test to demonstrate issues with whitespace in part…

cscutcher:
… names

Reproduces #130.

For new style table the generated script has a few problems for example;

```sh
    sgdisk \
      --new=2:0:+100M \
      --change-name=2:disk-vdb-name with spaces \
      --typecode=2:EF00 \
      /dev/vdb
```

and

```sh
mkfs.vfat \
         \
        /dev/disk/by-partlabel/disk-vdb-name with spaces
```

Legacy table style generates slightly different problems e.g.;

```sh
parted -s /dev/vdb -- mkpart name with spaces  1MiB 100MiB
```

### DIFF
--- a/example/gpt-name-with-whitespace.nix
+++ b/example/gpt-name-with-whitespace.nix
@@ -1,0 +1,41 @@
+{
+  disko.devices = {
+    disk = {
+      vdb = {
+        device = "/dev/disk/by-id/some-disk-id";
+        type = "disk";
+        content = {
+          type = "gpt";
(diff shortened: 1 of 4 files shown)


iFreilicht:
Thank you so much for this repro! I created a PR to fix the issue, based on your tests: #841

In case you're still active, I would appreciate if you had a look at it :) I'll close this PR now because mine also merges your tests.

PR #338, opened 2023-08-31
make hybrid example more generic

fricklerhandwerk:
otherwise it requires editing the file before actually using it for the quick start guide

### DIFF
--- a/example/hybrid.nix
+++ b/example/hybrid.nix
@@ -3,7 +3,7 @@
     disk = {
       main = {
         type = "disk";
-        device = "/dev/disk/by-id/ata-Samsung_SSD_850_EVO_250GB_S21PNXAGB12345";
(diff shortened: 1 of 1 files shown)


Lassulus:
the gpt format automatically uses partlabels now, so the device name is only used for installing. usually We want people to check which disk they are formatting so I would like the default in the example to be non functional so people don't wipe the wrong disk :D (but thats not true for all examples)

fricklerhandwerk:
How about removing the value altogether or requiring it to be supplied as a parameter, as in the other examples? I saw the getting-started-guide being inconsistent about passing arguments, because many samples don't even take any.

Lassulus:
ah yeah thats historic burden, I removed the argument passing since the config is now evaluated via the nixos module system and that could be done with options and config. But I still have to write an example for that

PR #349, opened 2023-09-05
examples: Add example for Hetzner dedicated, NVMe, ZFS

serpent213:


phaer:
@serpent213 

What's specific to hetzner in this example? 

It might be useful to add the specific model, as there are lots of different dedicated machine models at Hetzner - i.e. with less than or more than two disks, no nvme, etc.

For mirrored boots, you could use https://search.nixos.org/options?channel=23.05&show=boot.loader.grub.mirroredBoots and a `/boot2` mount or so to be still able to boot after 1 of your disks failed

serpent213:
@phaer

It will surely work on many other systems, but as the servers on the cheaper side at Hetzner don't have a dedicated remote console, I find it beneficial to have a “known good” configuration, because debugging the boot process is rather painful.

Of course it would be fine to label it as generic configuration and maintain a list, like: known to work on Hetzner model x, NotherHoster model y and so on.

Added the specific model to the header.

Thanks for the pointer to `grub.mirroredBoots`, was looking for that bit.

example/hetzner-dedi-nvme-zfs.nix
```diff
@@ -0,0 +1,88 @@
...
+# Migrate your network config from the rescue system.
+#
+# Some settings might be not optimal or superfluous.
+
+{ lib, disks ? [ "/dev/nvme0n1" "/dev/nvme1n1" ], ... }: {
+  disk = lib.genAttrs disks (dev: {
+    device = dev;
+    type = "disk";
+    content = {
+      type = "table";
```
Mic92, line 27: We no longer use type = "table" for new things.

```suggestion
      type = "gpt";
```

Mic92:
I am currently wondering what the best way forward for this is. One the one hand I think it's nice to have examples that are specific for a provider because it helps users on the other the hand we have to run CI for every example so it's a bit expensive over time.

serpent213:
> I also have to use mirrored boot for efi on hetzner now. Also it seems hat all new hetzner machines now have EFI enabled by default.

I ordered mine two weeks ago and had them enable EFI manually for me. When experimenting with *nixos-anywhere* the server ended up in a deadlock, apparently, the rescue system did not come up anymore. Ordered a manual reset then and they put it back into legacy (without telling me).  🤷🏼‍♂️

serpent213:
> I am currently wondering what the best way forward for this is. One the one hand I think it's nice to have examples that are specific for a provider because it helps users on the other the hand we have to run CI for every example so it's a bit expensive over time to have all these examples build.

I would be happy with some kind of community repo containing “untested” configurations.

Mic92:
Usually I can recommend just to get the 3h kvm console that you can book at no extra charge to debug this.

Mic92:
Also avoid [EX101](https://www.hetzner.com/dedicated-rootserver/ex101) kexec doesn't work there and the CPU will crash the system if it builds up load to quickly (maybe it draws too much voltage?)

serpent213:
Sure, that's how I came up with this config. But still I feel better having a working rescue system.

Mic92:
> > I am currently wondering what the best way forward for this is. One the one hand I think it's nice to have examples that are specific for a provider because it helps users on the other the hand we have to run CI for every example so it's a bit expensive over time to have all these examples build.
> 
> I would be happy with some kind of community repo containing “untested” configurations.

Yes. I think some repository or directory that has more real-world examples rather than edge cases we want to test in disko, would be nice.

Mic92:
> > Usually I can recommend just to get the 3h kvm console that you can book at no extra charge to debug this. Than you can also switch back and forth between bios and EFI.
> 
> Sure, that's how I came up with this config. But still I feel better having a working rescue system.

I see. For me the rescue system is also working fine when configured as EFI.

phaer:
> I am currently wondering what the best way forward for this is. One the one hand I think it's nice to have examples that are specific for a provider because it helps users on the other the hand we have to run CI for every example so it's a bit expensive over time to have all these examples build.

Agreed. I think the idea of not testing *all* the examples for each commit does sound like an okay work-around here? We could full test runs upon nixos releases so twice a year; we won't be able to *fully* test those examples without renting a physical machine of each model anyway ;)

Can confirm that the rescue system works for me on multiple hetzner machines with UEFI.

Mic92:
We have created https://github.com/nix-community/disko-templates now for examples likes this.
This will help new users to quickly get good examples rather than our unittest configurations.
Would you mind open a pull request to this repository instead?

PR #355, opened 2023-09-10
module: add images to system.build

Lassulus:


module.nix
```diff
@@ -71,6 +71,13 @@ in
       disko = builtins.trace "the .disko output is deprecated, please use .diskoScript instead" (cfg.devices._scripts pkgs).diskoScript;
       diskoNoDeps = builtins.trace "the .diskoNoDeps output is deprecated, please use .diskoScriptNoDeps instead" (cfg.devices._scripts pkgs).diskoScriptNoDeps;
 
+      diskoImages = diskoLib.makeDiskImage {
```
phaer, line 74: Is it intentional that the binding uses plural form (image**s**) while the function uses singular form?

Lassulus: no :D but which one is better?

Lassulus: it returns a path to a directory with possibly multiple imnages

phaer: Is it too late yet to rename the function to plural form as well then? I don't image their being too many users yet, but what do i know 😅 an alias could be a good compromise if so?

Mic92: It doesn't had documentation so far. So I think it is rather safe.

phaer (approved):
I think this is fine? Just letting you decide when to press that button :)

Lassulus:
@mergify queue

PR #358, opened 2023-09-11
Fix problem with missing part. labels.

rudolfvesely:
After `sgdisk` I got:

```
Warning: The kernel is still using the old partition table.
The new table will be used at the next reboot or after you
run partprobe(8) or kpartx(8)
The operation has completed successfully.
```

And then:

```
mkswap: cannot open /dev/disk/by-partlabel/...: No such file or directory
```

This is a simple solution to this problem.

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -44,6 +44,7 @@ let
     ${pkgs.systemdMinimal}/lib/systemd/systemd-udevd --daemon
     udevadm trigger --action=add
     udevadm settle
+    partprobe
 
     # populate nix db, so nixos-install doesn't complain
     export NIX_STATE_DIR=$TMPDIR/state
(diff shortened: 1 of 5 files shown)


Mic92:
I would have expected that udev also would do this. Can you confirm that after this change, the problem goes away?

rudolfvesely:
I've just tested it and it solves the problem. I've encountered this issue after `mkswap` on top of `randomEncryption` after partition with `luks`. I didn't have any problems with `mkswap` on top of `randomEncryption` after non-encrypted partition.

phaer:
@mergify queue

phaer:
Thanks! I remember having similar experiences in the past, where udevadm --settle wasn't sufficient and partprobe seemingly needed. Also think it does not hurt and could easily be removed if it proves unecessary

rudolfvesely:
Thank you for merging it.

Have you considered to wait for a label to show up? Because they do after a second when all tasks are propagated.

``` sh
wait=0
until [[ ${wait} -eq 5 ]] || [[ -b /dev/disk/by-label/nixos ]]; do
    echo "wait loop #$(( wait ++ ))"
    sleep 1
done
[[ ${wait} -lt 5 ]]
```

or via

`inotifywait`

Maybe as a separate phase between `_create` and `_mount`.

PR #361, opened 2023-09-12
some small fixups

Lassulus:
tests with nix-build were broken, so adding parted where partprobe is used fixes that
we also add a test for raid0, which failed earlier today due to not being removed by the disk-deactvate script


Lassulus:
@mergify queue

PR #362, opened 2023-09-12
Specify alignment for sgdisk.

rudolfvesely:
The following command:

`sgdisk --new=4:0:-0 /dev/xda`

will work but unless the overall disk size is aligned to 4k, the new partition will not be aligned to 4k.

`--set-alignment=2048 --align-end` aligns new partitions to 1 MiB (2048 sectors). This enables for example `--sector-size=4096` for `cryptsetup luksFormat`.

### DIFF
--- a/lib/types/gpt.nix
+++ b/lib/types/gpt.nix
@@ -100,6 +100,8 @@ in
       default = ''
         ${lib.concatStrings (map (partition: ''
           sgdisk \
+            --set-alignment=2048 \
+            --align-end \
             --new=${toString partition._index}:${partition.start}:${partition.end} \
             --change-name=${toString partition._index}:${partition.label} \
             --typecode=${toString partition._index}:${partition.type} \


lib/types/gpt.nix
```diff
@@ -100,6 +100,8 @@ in
       default = ''
         ${lib.concatStrings (map (partition: ''
           sgdisk \
+            --set-alignment=2048 \
```
Mic92, line 103: Should we make this configurable? I know there are some arm boards that sometimes have weird requirements for embedded uboot loader.

rudolfvesely: The `--set-alignment` kicks in only when you specify relative size (100%) and the default behavior gives you maximum partition size. The only downside with this change is that you might possibly waste for example 0.5 MiB at the end. You can still specify `start = "1234";` for ARM boards and create unaligned partition if that's what you need.

Mic92: Sounds good.

Mic92:
@mergify queue

PR #363, opened 2023-09-12
Add swap config to btrfs types

### LINKED ISSUE #262: BTRFS Swapfile
I'd like to be able to designate a swap subvolume and have disko automatically create a swapfile.
Related: https://github.com/nix-community/disko/issues/179

Alxandr:
I tried adding swapfile support for btrfs. I've not been able to test if it works though, cause I'm not able to run any of the tests - I get a segfault from qemu-kvm. I will throw up a VM one of the comming days to see if it actually works however. I've generated and inspected the `_create` scripts however, and they seem correct, but I obviously need to test things. I'm leaving this as a draft until I've managed to do some tests at least.

Resolves #262

### DIFF
--- a/example/btrfs-subvolumes.nix
+++ b/example/btrfs-subvolumes.nix
@@ -49,9 +49,28 @@
                   "/swap" = {
                     mountpoint = "/.swapvol";
                     swap.enable = true;
-                    swap.size = "20M";
+                    swap.files = [
+                      {
+                        size = "20M";
+                      }
+                      {
+                        size = "20M";
+                        path = "rel-path";
+                      }
+                    ];
                   };
                 };
+
+                swap.enable = true;
+                swap.files = [
+                  {
+                    size = "20M";
+                  }
+                  {
+                    size = "20M";
+                    path = "rel-path";
+                  }
+                ];
               };
             };
           };
--- a/example/btrfs-subvolumes.nix
+++ b/example/btrfs-subvolumes.nix
@@ -45,6 +45,12 @@
                   };
                   # This subvolume will be created but not mounted
                   "/test" = { };
+                  # Subvolume for the swapfile
+                  "/swap" = {
+                    mountpoint = "/.swapvol";
+                    swap.enable = true;
+                    swap.size = "20M";
+                  };
                 };
               };
             };
--- a/example/btrfs-subvolumes.nix
+++ b/example/btrfs-subvolumes.nix
@@ -61,6 +61,7 @@
                   };
                 };
 
+                mountpoint = "/partition-root";
                 swap.enable = true;
                 swap.files = [
                   {
@@ -68,7 +69,6 @@
                   }
                   {
                     size = "20M";
-                    path = "rel-path";
                   }
                 ];
               };
--- a/example/btrfs-subvolumes.nix
+++ b/example/btrfs-subvolumes.nix
@@ -48,29 +48,25 @@
                   # Subvolume for the swapfile
                   "/swap" = {
                     mountpoint = "/.swapvol";
-                    swap.enable = true;
-                    swap.files = [
-                      {
-                        size = "20M";
-                      }
-                      {
-                        size = "20M";
-                        path = "rel-path";
-                      }
-                    ];
+                    swap.file = {
+                      swapfile.size = "20M";
+                      swapfile2.size = "20M";
+                      swapfile2.path = "rel-path";
+                      swapfile3.size = "20M";
+                      swapfile3.enable = false;
+                    };
                   };
                 };
 
                 mountpoint = "/partition-root";
-                swap.enable = true;
-                swap.files = [
-                  {
+                swap.file = {
+                  swapfile = {
                     size = "20M";
-                  }
-                  {
+                  };
+                  swapfile1 = {
                     size = "20M";
-                  }
-                ];
+                  };
+                };
               };
             };
           };
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -1,51 +1,83 @@
 { config, options, diskoLib, lib, rootMountPoint, parent, device, ... }:
 let
+  swapFileType = lib.types.submodule ({ config, ... }: {
+    options = {
+      size = lib.mkOption {
+        type = lib.types.strMatching "^([0-9]+[KMGTP])?$";
+        description = "Size of the swap file (e.g. 2G)";
+      };
+
+      pathPrefix = lib.mkOption {
+        type = lib.types.str;
+        default = "swapfile";
+        description = "Prefix for the swap file";
+      };
+
+      path = lib.mkOption {
+        type = lib.types.nullOr lib.types.str;
+        default = null;
+        description = "Path to the swap file (relative to the mountpoint) - gets the index as an argument, starting at 1";
+      };
+
+      _path = lib.mkOption {
+        internal = true;
+        type = lib.types.functionTo lib.types.str;
+        default = idx:
+          if config.path != null
+          then config.path
+          else "${config.pathPrefix}${lib.optionalString (idx != 1) (toString idx)}";
+      };
+    };
+  });
+
   swapType = lib.types.submodule ({ config, ... }: {
     options = {
       enable = lib.mkEnableOption "swap";
 
-      size = lib.mkOption {
-        type = lib.types.nullOr (lib.types.strMatching "^([0-9]+[KMGTP])?$");
-        default = null;
-        description = "Size of the swap file (e.g. 2G)";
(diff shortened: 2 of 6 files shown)


lib/types/btrfs.nix
```diff
@@ -1,4 +1,52 @@
...
+  partitionSwapCreate = device: swap:
+    if !swap.enable
+    then ""
+    else
+      assert lib.asserts.assertMsg (swap.size != null) "swap size must be set";
+      ''
+        (
+          MNTPOINT=$(mktemp -d)
+          mount ${device} "$MNTPOINT" -o subvol=/
+          trap 'umount $MNTPOINT; rm -rf $MNTPOINT' EXIT
```
Mic92, line 45: The trap is scoped here to the subshell right?

Alxandr: I'm not a bash expert - but I copied this from further down [in the same file](https://github.com/nix-community/disko/blob/b32f0e2/lib/types/btrfs.nix#L82). So if it's wrong, it's wrong for all of the subvolume creations as well.

Alxandr:
I've ran this on a VM now, checking that the swap file get's created as expected, so I'm taking this PR out of draft-status. I've not yet checked that the nixos config works (that the swap-file is used) cause my current configuration is in the middle of a major rewrite so I don't currently have a good test for that. I will hopefully get that done within next week though.

Alxandr:
Tested on a VM, and I indeed get swap setup:

image

lib/types/btrfs.nix
Mic92: Sounds about right.

lib/types/btrfs.nix
```diff
@@ -75,12 +133,14 @@
...
       default = ''
         mkfs.btrfs ${config.device} ${toString config.extraArgs}
+        ${partitionSwapCreate config.device config.swap}
         ${lib.concatMapStrings (subvol: ''
           (
             MNTPOINT=$(mktemp -d)
             mount ${config.device} "$MNTPOINT" -o subvol=/
             trap 'umount $MNTPOINT; rm -rf $MNTPOINT' EXIT
             btrfs subvolume create "$MNTPOINT"/${subvol.name} ${toString subvol.extraArgs}
+            ${subvolSwapCreate ''"$MNTPOINT"/${subvol.name}'' subvol.swap}
```
Lassulus, line 143:
```suggestion
            ${lib.optionalString subvol.swap.enable ''
              btrfs filesystem mkswapfile --size ${subvol.swap.size} "$MNTPOINT"/${subvol.swap.path}
            ''}
```

Lassulus: I guess it would be easier to inline this? would make the code more compact

lib/types/btrfs.nix
```diff
@@ -1,4 +1,52 @@
...
+let
+  swapType = lib.types.submodule ({ config, ... }: {
+    options = {
+      enable = lib.mkEnableOption "swap";
+
+      size = lib.mkOption {
+        type = lib.types.nullOr (lib.types.strMatching "^([0-9]+[KMGTP])?$");
+        default = null;
+        description = "Size of the swap file (e.g. 2G)";
+      };
```
Lassulus, line 11:
```suggestion
      size = lib.mkOption {
        type = lib.types.strMatching "^([0-9]+[KMGTP])?$";
        description = "Size of the swap file (e.g. 2G)";
      };
```

lib/types/btrfs.nix
```diff
@@ -1,4 +1,52 @@
...
+        "${mountpoint}/${swap.path}";
+    }];
+  };
+
+  subvolSwapCreate = mountpoint: swap:
+    if !swap.enable
+    then ""
+    else
+      assert lib.asserts.assertMsg (swap.size != null) "swap size must be set";
+      ''btrfs filesystem mkswapfile --size ${swap.size} ${mountpoint}/${swap.path}'';
```
Lassulus, line 34:
```suggestion

```

lib/types/btrfs.nix
```diff
@@ -1,4 +1,52 @@
...
+
+  partitionSwapCreate = device: swap:
+    if !swap.enable
+    then ""
+    else
+      assert lib.asserts.assertMsg (swap.size != null) "swap size must be set";
+      ''
```
Lassulus, line 41:
```suggestion
    lib.optionalString swap.enable ''
```

Lassulus (changes requested):
some code simplification.
Also this config supports only one swapfile per device/subvolume. are multiple swapfiles maybe a feature we want to support?

Alxandr:
> some code simplification. Also this config supports only one swapfile per device/subvolume. are multiple swapfiles maybe a feature we want to support?

Good question. I don't know too much about swap (beyond the fact that I've been told I should have one 😅), so I have no idea why you would have more than one. That isn't to say there aren't good reasons, that's just stating my lack of knowledge. Until I started looking at this feature I didn't even know you *could* have more than one swap.

I'm not sure what would be the best API for allowing multiple swaps though. Maybe instead of having:

```nix
swap = {
  enable = true;
  path = "some-path";
  size = "2G";
};
```

we could do

```nix
swap = {
  enable = true;
  files = [{
    path = "some-path";
    size = "2G";
  }];
};
```

Having a default path would be a bit trickier then - but just suffixing the array index would probably be fine?

lib/types/btrfs.nix
Alxandr: I personally don't like multiline and re-entrant string-interpolations, but that's just my personal preference. I can definitely make this change if that's what preferred.

lib/types/btrfs.nix
Alxandr: I *think* this might not work. I will test it, but when I was creating this code I just looked at outcomes in `nix repl`, and it would crash on me from `size` not being set on items that did not have `enable = true;`. It might be because I tried to print out values in the repl though. Now that I have a system where I can test this I will see if it works without `null` being a valid value.

lib/types/btrfs.nix
Alxandr: Thanks. Didn't know about `optionalString` :). Will update later today.

lib/types/btrfs.nix
Alxandr: Do you still feel it's better to inline after I've added support for multiple swapfiles?

lib/types/btrfs.nix
Alxandr: I've removed the ability to set this to `null` when moving to a list of files, cause the whole issue became moot.

lib/types/btrfs.nix
Alxandr: I have updated to using `optionalString`.

example/btrfs-subvolumes.nix
```diff
@@ -45,7 +45,32 @@
...
+                mountpoint = "/partition-root";
+                swap.enable = true;
+                swap.files = [
+                  {
+                    size = "20M";
+                  }
+                  {
+                    size = "20M";
+                  }
+                ];
```
Lassulus, line 73:
```suggestion
                swap.files = {
                  swapfile1 = {
                    size = "20M";
                  };
                  swapfile2 = {
                    size = "20M";
                  };
                };
```

Lassulus: how about this schema? lists are pretty bad since you cannot easy modify of reference them. the key could be the default value of the path?

Alxandr: Agreed. Also - maybe I should just move the `enable` switch to the individual swap files as well, such that separate modules can disable a specific swap file for instance? I'll try to update it today.

Lassulus: with attrsOf we don't need an enable flag. either it's there or it isn't

Alxandr: Right - but can you disable it in another module than where it was created? I personally have my disko config split up into multiple modules using things like `mkDefault`, and I don't know how to "remove" a already declared key in an attribute set.

Lassulus: Ah there are multiple ways. The easiest would be just to define it where needed. But you use swap.files = lib.mkForce {} if you want to empty it

Alxandr: Fair enough. I can remove the enable flag if you think that's better.

Mic92:
@Lassulus Is this pull request ready to go?

icyrockcom:
FWIW, I just moved my nixos from one disk to another and used disko with this PR merged for that. Swap seems to have been set up as expected.

Lassulus:
This was now implemented in a more minimal version in #440

PR #366, opened 2023-09-14
Add extra args into swap.

rudolfvesely:
The `swap` type doesn't have `extraArgs` like for example `btrfs` / `filesystem` type has. I tested it with `[ "-f" "--label=abc" ]` and it works well.


Mic92:
@mergify queue

PR #369, opened 2023-09-15
Changing how to open LUKS.

rudolfvesely:
Changing how to open encrypted partition from `luksOpen` to `open`.

`open` works same as `luksOpen` for LUKS1/2 but it is more flexible so it is possible to pass `--type=plain`.

I tested it with both LUKS2 and plain encrypted partitions in a single file.


Mic92:
@mergify queue

PR #376, opened 2023-09-20
types luks: add password prompt

Lassulus:


lib/types/luks.nix
```diff
@@ -3,6 +3,8 @@ let
   keyFile =
     if lib.hasAttr "keyFile" config.settings
     then config.settings.keyFile
+    else if config.askPassword
+    then ''<(echo -n "$password")''
```
Mic92, line 7: This will print the password again: https://github.com/nix-community/disko/issues/385

Lassulus:
@mergify queue

PR #384, opened 2023-09-23
make-disk-image: add features and fix stuff from #383

Janik-Haag:
This is a follow up to #383 which fixes some of the problems I described there.
I tested most possible combinations and also used --write-to-disk to write to a physical disk and booted it and it worked just fine

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -80,8 +80,9 @@ in
       copies the src to the dst on the VM, before disko is run
       This is useful to provide secrets like LUKS keys, or other files you need for formating
     * --post-format-files <src> <dst>
-      copies the src to the dst on the finished image.
+      copies the src to the dst on the finished image
       These end up in the images later and is useful if you want to add some extra stateful files
+      They will have the same permissions but will be owned by root:root
     USAGE
     }
 
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -83,6 +83,10 @@ in
       copies the src to the dst on the finished image
       These end up in the images later and is useful if you want to add some extra stateful files
       They will have the same permissions but will be owned by root:root
+    * --build-memory <amt>
+      specify the ammount of memory that gets allocated to the build vm (in mb)
+      This can be usefull if you want to build images with a more involed NixOS config
+      By default the vm will get 1024M/1GB
     USAGE
     }
 
@@ -106,6 +110,15 @@ in
         dst=$3
         cp --reflink=auto -r "$src" copy_after_disko/"$(echo "$dst" | base64)"
         shift 2
+        ;;
+      --build-memory)
+        regex="^[0-9]+$"
+        if ! [[ $2 =~ $regex ]]; then
+          echo "'$2' is not a number"
+          exit 1
+        fi
+        build_memory=$2
+        shift 1
         ;;
       *)
         showUsage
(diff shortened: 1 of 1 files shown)


Lassulus:
this needs a rebase first

Lassulus:
I guess we should move the --write-to-disk to another function/output since it doesn't really generate images anymore :)

maybe we should do this in another PR, since we also want to support multiple disks? (as there is no reasons why it should be limited to a single disk?)

lib/make-disk-image.nix
```diff
@@ -140,7 +161,27 @@ in
       done
       ${installer}
     ''}
-    export QEMU_OPTS=${lib.escapeShellArg "${QEMU_OPTS} -m 1024"}
+
+    if [[ -z $build_memory ]]; then
+      build_memory=1024
```
Lassulus, line 166: not sure this has any effect without export

lib/make-disk-image.nix
```diff
@@ -140,7 +161,27 @@ in
...
+    fi
+    if [[ -n $disk ]] && [ ${diskNumber} -eq 1 ]; then
+      export preVM=${diskoLib.writeCheckedBash { inherit pkgs checked; } "preVM.sh" ''
+        set -efu
+        mv copy_before_disko copy_after_disko xchg/
+      ''}
+      export postVM=${diskoLib.writeCheckedBash { inherit pkgs checked; } "postVM.sh" extraPostVM}
```
Lassulus, line 173: usually the extraPostVM is used to compress the image. so not sure we would need to pass it in this case

Janik-Haag: Ah I don't know, I didn't have a use case for it yet but in general it doesn't seem like it would hurt leaving it in here and would ensure that you have the same feature set.

Lassulus:
something like `./result --disk main /dev/sda --disk test /tmp/some-disk-file` 
output should be in the same file (make-disk-image.nix) but use another output. we should rename the impure output to something like writeDiskImages and the new one would be writeDiskImagesToFiles? For now I would let it run inside a VM, since there can be mixups with partlabels and dataloss without VMs. we can add an --no-vm option later which would skip the vm part. but that's a later feature/optimization

lib/make-disk-image.nix
Janik-Haag: I finally came around to update the pr, removed the `--write-to-file` option and will do another pr adding it back in with the suggestions from this one once I find time for that.

lib/make-disk-image.nix
```diff
@@ -140,7 +154,13 @@ in
       done
       ${installer}
     ''}
-    export QEMU_OPTS=${lib.escapeShellArg "${QEMU_OPTS} -m 1024"}
+
+    if [[ -z $build_memory ]]; then
+      build_memory=1024
+    fi
```
Lassulus, line 160: build_memory=''${build_memory:-1024}

Lassulus, line 160:
```suggestion
    build_memory=''${build_memory:-1024}
```

Janik-Haag: If I apply that locally the script fails because the var is just empty and qemu complaints about a missing value, but I don't know why since the syntax you suggested works fine if I do it in a test script.

Janik-Haag: any idea why that is?

Janik-Haag: fixed

Lassulus (approved):

Lassulus:
@mergify queue

Lassulus:
@mergify queue

Mic92:
@mergify rebase

PR #388, opened 2023-09-25
zfs_volume: fix /dev/zvol/ symlink race

samhug:
This fixes a race condition where the partprobe command tries to read the /dev/zvol symlink, potentially before it's created


Mic92:
@mergify queue

PR #390, opened 2023-09-25
docs: add make-disk-image

Janik-Haag:
Just adding some docs for `lib/make-disk-image`

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -44,46 +44,70 @@ Options:
   run with set -x
 ```
 
-## Generating Disk Images with secrets included
+# Generating Disk Images with Secrets Included using Disko
 
-If you have a system config that uses disko then you can run a script that generates a `.raw` VM image.
-This image can be used as a VM image, but you can also dd the `.raw` image on to a physical drive and just boot it.
-To execute the script replace the `mySystem` part in the snippet below and run it.
-```
-nix build .#nixosConfigurations.mySystem.config.system.build.diskoImagesScript
-```
-Now you will have a result file which will output something like this:
-```
-./result --help
-Usage: $script [options]
+Using Disko on NixOS allows you to efficiently create `.raw` VM images from a
+system configuration. The generated image can be used as a VM or directly
+written to a physical drive to create a bootable disk. Follow the steps below to
+generate disk images:
 
-Options:
-* --pre-format-files <src> <dst>
-  copies the src to the dst on the VM, before disko is run
-  This is useful to provide secrets like LUKS keys, or other files you need for formating
-* --post-format-files <src> <dst>
-  copies the src to the dst on the finished image
-  These end up in the images later and is useful if you want to add some extra stateful files
-  They will have the same permissions but will be owned by root:root
-* --build-memory
-  specify the ammount of memory that gets allocated to the build vm (in mb)
-  This can be usefull if you want to build images with a more involed NixOS config
-  By default the vm will get 1024M/1GB
-* --write-to-disk </dev/disk>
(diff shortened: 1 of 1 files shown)


docs/reference.md
```diff
@@ -44,4 +44,46 @@ Options:
...
+* --post-format-files <src> <dst>
+  copies the src to the dst on the finished image
+  These end up in the images later and is useful if you want to add some extra stateful files
+  They will have the same permissions but will be owned by root:root
+* --build-memory
+  specify the ammount of memory that gets allocated to the build vm (in mb)
+  This can be usefull if you want to build images with a more involed NixOS config
+  By default the vm will get 1024M/1GB
+* --write-to-disk </dev/disk>
+  use a actuall disk instead of writing to a file
```
Lassulus, line 73:
```suggestion
  use an actual disk instead of writing to a file
```

docs/reference.md
```diff
@@ -44,4 +44,70 @@ Options:
...
+# Generating Disk Images with Secrets Included in Disko
+
+Using Disko on NixOS allows you to efficiently create `.raw` VM images from a
+system configuration. The generated image can be used as a VM or directly
+written to a physical drive to create a bootable disk. Follow the steps below to
+generate disk images:
+
+## Generating the `.raw` VM Image
+
+1. **Modify the script name:** Replace `mySystem` in the command below with your
```
Janik-Haag, line 56: I feel like script name doesn't really fit here but don't know a better phrasing

docs/reference.md
```diff
@@ -44,4 +44,70 @@ Options:
...
+     By default the vm will get 1024M/1GB
+   * --write-to-disk </dev/disk>
+     use an actuall disk instead of writing to a file
+     This only works if your conifg has only one disk specified
+     There is no check if the specified path is actually a disk so you can also write to another file
+   ```
+
+## Additional Configuration
+
+- For virtual drive use, define the image size in your Disko configuration:
```
Janik-Haag, line 88:
```suggestion
- For virtual drives, define the image size in your Disko configuration:
```

docs/reference.md
```diff
@@ -44,4 +44,70 @@ Options:
...
+
+  ```bash
+  disko.devices.disk.<drive>.imageSize = "32G"; # Set your preferred size
+  ```
+
+- If the `.raw` image size is problematic, use `--write-to-disk` to write
+  directly to a drive. This bypasses the `.raw` file generation and is suitable
+  for single disk, simple setups.
```
Janik-Haag, line 96:
```suggestion
- If the `.raw` image size is not optimal, use `--write-to-disk` to write
  directly to a drive. This bypasses the `.raw` file generation, which saves on read/write operations
  and is suitable for single disk setups.
```

docs/reference.md
Mic92: technically the other one is also correct, but maybe more confusing.

docs/reference.md
```diff
@@ -44,4 +44,70 @@ Options:
   run with set -x
 ```
 
-## 
+# Generating Disk Images with Secrets Included in Disko
```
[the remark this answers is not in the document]
Mic92: I think both ways are fine here as well, but we can stick with your variant.

docs/reference.md
Mic92:
```suggestion
1. **Build the disko image script:** Replace `mySystem` in the command below with your
```

Mic92:
@mergify queue

Lassulus:
@mergify queue

PR #394, opened 2023-09-25
converted legacy-table example to new disko API

cvoges12:


Mic92:
Thanks but this test is called legacy table because we it tests the old table format.
We keep it since we still have users and therefore need to support it for longer.

cvoges12:
> Thanks but this test is called legacy table because we it tests the old table format. We keep it since we still have users and therefore need to support it for longer.

Are you sure? I thought this was for the old MBR format.

PR #406, opened 2023-09-27
make boot large enough in tests

Mic92:
100M is not enough and I see people just copying examples without putting in any extra thoughts https://github.com/TUM-DSE/doctor-cluster-config/pull/368


Mic92:
@mergify queue

PR #410, opened 2023-10-01
Fix example in README.md

kai-tub:
The minimal example was missing the wrapping `disko.devices` attributes which resulted in the following error message:

```
error: attribute 'disko' missing

       at /nix/store/1lvj5ry4y5qyis40qcch1csf8h86z8bk-disko/share/disko/default.nix:10:49:

            9|       # _file = toString input;
           10|       imports = lib.singleton { disko.devices = cfg.disko.devices; };
             |                                                 ^
           11|       options = {
       Did you mean disk?
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,32 +59,34 @@ A simple disko configuration may look like this:
 
 ```
 { disks ? [ "/dev/vdb" ], ... }: {
- disk = {
-  vdb = {
-   device = builtins.elemAt disks 0;
-   type = "disk";
-   content = {
-    type = "gpt";
-    partitions = {
-     ESP = {
-      size = "100M";
-      content = {
-       type = "filesystem";
-       format = "vfat";
-       mountpoint = "/boot";
-      };
-     };
-     root = {
-      size = "100%";
-      content = {
-       type = "filesystem";
-       format = "ext4";
-       mountpoint = "/";
-      };
-     };
-    };
-   };
-  };
+ disko.devices = {
+  disk = {
+   vdb = {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -59,32 +59,34 @@ A simple disko configuration may look like this:
...
+ disko.devices = {
+  disk = {
+   vdb = {
+    device = builtins.elemAt disks 0;
+    type = "disk";
+    content = {
+     type = "gpt";
+     partitions = {
+      ESP = {
+       size = "100M";
```
Mic92, line 71:
```suggestion
       size = "500M";
```

kai-tub:
I am not really sure why 500 instead of 100, but I have no issues with it :D
Also why not 512 looks weird but I have no clue about file systems.

README.md
```diff
@@ -59,32 +59,34 @@ A simple disko configuration may look like this:
...
+ disko.devices = {
+  disk = {
+   vdb = {
+    device = builtins.elemAt disks 0;
+    type = "disk";
+    content = {
+     type = "gpt";
+     partitions = {
+      ESP = {
+       size = "500M";
```
Mic92, line 71:
```suggestion
       size = "512M";
```

Mic92: Well. than it's consistent with the rest I suppose.

Mic92:
> I am not really sure why 500 instead of 100, but I have no issues with it :D Also why not 512 looks weird but I have no clue about file systems.

People tend to copy these examples and than end up with boot partitions that are too tiny for a regular nixos system.

PR #417, opened 2023-10-13
luks: fix interactive mount

Ma27:
Before that it was only attempted to unlock luks with the key from `$password`, however that variable didn't exist before.

cc @Lassulus


Lassulus:
Better do this inside the if. Otherwise we ask for the password for an already unlocked luks device

Ma27:
Whoops, good point, fixed! :)

Lassulus:
@mergify queue

PR #424, opened 2023-10-18
disk-deactivate: Remove MBR bootstrap code for disk

dongcarl:
Not sure if `coreutils` needs to be added to `pkgs` somewhere for `dd`, but this should work.


Lassulus:
@mergify queue

PR #425, opened 2023-10-19
boot.loader.grub.device not required

stfnx:
It is not necessary to manually set `boot.loader.grub.devices`, since Disko will automatically add the device to this option.

In fact setting the device manually will lead to the following issue, because in `mirroredBoots` the same device will end up twice.
```
error:
       Failed assertions:
       - You cannot have duplicated devices in mirroredBoots
```


Mic92:
Thanks

PR #427, opened 2023-10-22
swap: don't mount

### LINKED ISSUE #379: Encrypted swap fails to mount
encrypted swap fails to mount when using `disko -m mount` as `mkswap` is not called before mounting

Gerg-L:
Closes #379
Swap doesn't need to be mounted at all when calling `-m mount`

### DIFF
--- a/lib/types/swap.nix
+++ b/lib/types/swap.nix
@@ -47,13 +47,7 @@
     };
     _mount = diskoLib.mkMountOption {
       inherit config options;
-      default = {
-        fs.${config.device} = ''
-          if ! swapon --show | grep -q "^$(readlink -f ${config.device}) "; then
-            swapon ${config.device}
-          fi
-        '';
-      };
+      default = { };
     };
     _config = lib.mkOption {
       internal = true;


Lassulus:
hmm, not sure if this is the correct solution, would be better to fix the check? I guess the mount does not need to mount if it's run after the create? but this will probably cause issues if the mount is run after a reboot or so

Gerg-L:
> hmm, not sure if this is the correct solution, would be better to fix the check? I guess the mount does not need to mount if it's run after the create? but this will probably cause issues if the mount is run after a reboot or so

what do you think is the best solution
only mounting non-encrypted swap?
or just run `mkswap` on encrypted swap
or  be encrypted with the default values of the nixos module like this: <https://github.com/NixOS/nixpkgs/blob/e9363e559c31e7744c0e13d47245d1dd55919604/nixos/modules/config/swap.nix#L281-L290>

Mic92:
> hmm, not sure if this is the correct solution, would be better to fix the check? I guess the mount does not need to mount if it's run after the create? but this will probably cause issues if the mount is run after a reboot or so

Do you consider a typical use case that popl

Gerg-L:
> Do you consider a typical use case to have swap activated during installations?

I don't really see a reason for needing swap during installation

> Anyway. I changed it up here to not break random encryption: https://github.com/nix-community/disko/pull/447

That fixes my issue

Lassulus:
should be taken care of by https://github.com/nix-community/disko/pull/455

PR #428, opened 2023-10-22
Added example/zfs-impermanence.nix

jpds:
Added a base example for systems where impermanence with ZFS is desired.

### DIFF
--- a/example/zfs-impermanence.nix
+++ b/example/zfs-impermanence.nix
@@ -1,0 +1,69 @@
+{ disks ? [ "/dev/sda" ], ... }: {
+  disko.devices = {
+    disk = {
+      x = {
+        device = builtins.elemAt disks 0;
+        type = "disk";
+        content = {
+          type = "gpt";
+          partitions = {
+            ESP = {
+              size = "512M";
+              type = "EF00";
+              content = {
+                type = "filesystem";
+                format = "vfat";
+                mountpoint = "/boot";
+              };
+            };
+            zfs = {
+              size = "100%";
+              content = {
+                type = "zfs";
+                pool = "zroot";
+              };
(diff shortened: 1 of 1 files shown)


example/zfs-impermanence.nix
```diff
@@ -0,0 +1,69 @@
...
+
+        datasets = {
+          "local" = {
+            type = "zfs_fs";
+            options.mountpoint = "none";
+          };
+          "local/home" = {
+            type = "zfs_fs";
+            options.mountpoint = "/home";
+            options."com.sun:auto-snapshot" = "false";
```
Mic92, line 47: Wouldn't you want auto snaphots for home and persist at least?

example/zfs-impermanence.nix
```diff
@@ -0,0 +1,69 @@
...
+      };
+    };
+    zpool = {
+      zroot = {
+        type = "zpool";
+        rootFsOptions = {
+          compression = "zstd";
+          mountpoint = "none";
+        };
+        options.ashift = "12";
```
Mic92, line 37:
```suggestion
        # This is for 4k aligned block devices.
        # For NVMEs i.e. `nvme id-ns -H /dev/nvme0n1 | grep "Relative Performance"` as described here: https://wiki.archlinux.org/title/Advanced_Format#NVMe_solid_state_drives
        # to find out what the native sector size of your storage device
        options.ashift = "12";
```

Mic92:
@mergify rebase

Mic92:
We have created https://github.com/nix-community/disko-templates now for examples likes this.
This will help new users to quickly get good examples rather than our unittest configurations.
Would you mind open a pull request to this repository instead?

PR #431, opened 2023-10-22
luks: fix read ShellCheck warning SC2162

williamvds:
Use `-r` to avoid mangling backslashes, use `IFS=` to not discard leading and trailing whitespace.

While trying out disko I hit warning SC2162 which caused the build script derivation to fail, because it's ShellChecked. Fix the warning with the suggested approach of combining the `-r` flag and `IFS=`.


Lassulus:
@mergify queue

PR #439, opened 2023-11-04
luks: use allowDiscards option

Stunkymonkey:
i feel we should better use the nixos options, so when other people want to debug problem they really see what is defined.


Lassulus:
@mergify queue

PR #444, opened 2023-11-11
Implement preMountHook & postMountHook

amarshall:
These options previously existed but had no effect. Now they are implemented. They only affect mount during disko’s filesystem creation, not during any later mounts.

Adding test case to existing “complex” example and test mostly because couldn’t think of any better place to put it.

### DIFF
--- a/example/complex.nix
+++ b/example/complex.nix
@@ -110,6 +110,9 @@
               mountOptions = [
                 "defaults"
               ];
+              postMountHook = ''
+                touch /mnt/ext4_on_lvm/file-from-postMountHook
+              '';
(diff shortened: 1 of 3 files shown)


Lassulus:
@mergify rebase

Lassulus:
Seems ci got stuck and you need to rebase the branch manually

Mic92:
@mergify rebase

Mic92:
Done in https://github.com/nix-community/disko/pull/446

Mic92:
Merged in https://github.com/nix-community/disko/pull/446

PR #446, opened 2023-11-12
Implement preMountHook & postMountHook

Mic92:
These options previously existed but had no effect. Now they are implemented. They only affect mount during disko’s filesystem creation, not during any later mounts.

Adding test case to existing “complex” example and test mostly because couldn’t think of any better place to put it.


Mic92:
@mergify queue

PR #447, opened 2023-11-12
swap: don't activate random encrypted swap files

Mic92:
This requires more code and is not implemented just now.


Lassulus:
@mergify queue

PR #464, opened 2023-12-10
fix: create btrfs subvolume parent directories

### LINKED ISSUE #409: cannot create nested BTRFS subvolumes
Hi.

I'm trying to create a nested subvol in `home` like this.
```nix
...
           luks = {
              size = "100%";
              content = {
                type = "luks";
                name = "crypt-btrfs-part";
                passwordFile = secretKeyFile;
                content = {
                  type = "btrfs";
                  extraArgs = [ "-f" ];
                  subvolumes = {
                    "/root" = {
                      mountpoint = "/";
                      mountOptions = [ "compress=zstd" "noatime" ];
                    };
                    "/home" = {
                      mountpoint = "/home";
                      mountOptions = [ "compress=zstd" "noatime" ];
                    };
                    # "/home/besser" = { }; # does not help either
                    "/home/besser/Sync" = { }; # added
                    "/nix" = {
                      mountpoint = "/nix";
                      mountOptions = [ "compress=zstd" "noatime" ];
                    };
                  };
...
```

I get
```
...
+ btrfs subvolume create /tmp/tmp.d8z2eybzDV//home/besser/Sync
ERROR: cannot access '/tmp/tmp.d8z2eybzDV//home/besser': No such file or directory
...
```

What am I doing wrong? Are nested subvols supported by disko?

Thanks! :-)

ctem:
Fixes #409

Note: Because existing directories can't be converted to subvolumes directly, this implementation assumes hierarchical subvolume creation order (e.g., `/subvol1` is created before `/subvol1/subvol2`).

### DIFF
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -126,8 +126,10 @@ in
             MNTPOINT=$(mktemp -d)
             mount ${config.device} "$MNTPOINT" -o subvol=/
             trap 'umount $MNTPOINT; rm -rf $MNTPOINT' EXIT
-            btrfs subvolume create "$MNTPOINT"/${subvol.name} ${toString subvol.extraArgs}
-            ${swapCreate "$MNTPOINT/${subvol.name}" subvol.swap}
+            SUBVOL_ABS_PATH="$MNTPOINT/${subvol.name}"
+            mkdir -p "$(dirname "$SUBVOL_ABS_PATH")"
+            btrfs subvolume create "$SUBVOL_ABS_PATH" ${toString subvol.extraArgs}
+            ${swapCreate "$SUBVOL_ABS_PATH" subvol.swap}
           )
         '') (lib.attrValues config.subvolumes)}
       '';


lib/types/btrfs.nix
```diff
@@ -126,8 +126,10 @@ in
             MNTPOINT=$(mktemp -d)
             mount ${config.device} "$MNTPOINT" -o subvol=/
             trap 'umount $MNTPOINT; rm -rf $MNTPOINT' EXIT
-            btrfs subvolume create "$MNTPOINT"/${subvol.name} ${toString subvol.extraArgs}
-            ${swapCreate "$MNTPOINT/${subvol.name}" subvol.swap}
+            SUBVOL_ABS_PATH="$MNTPOINT/${subvol.name}"
+            mkdir -p "''${SUBVOL_ABS_PATH%/*}"
```
Lassulus, line 130: usually we try to avoid bash features, since people also use disko on non bash platforms (like openwrt)

ctem: Thanks for your review! Afaik, this particular parameter expansion is [required by POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/V3_chap02.html#tag_18_06_02). Based on my tests, recent versions of BusyBox's `ash` (used by OpenWrt) can also handle it.

 I'm happy to modify this to accommodate edge cases but would need to know what to target. We could replace it w/command substitution (e.g., using `dirname`, which is also [required by POSIX](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/dirname.html), or `awk`, `grep`, `sed`, ...), but given the choice, I tend to prefer parameter expansion to prevent invoking a separate process.

ctem: ...or the third option: Nix! 🤭

I've updated the PR to use the `lib.dirOf` function.

lib/types/btrfs.nix
```diff
@@ -126,8 +126,10 @@ in
             MNTPOINT=$(mktemp -d)
             mount ${config.device} "$MNTPOINT" -o subvol=/
             trap 'umount $MNTPOINT; rm -rf $MNTPOINT' EXIT
-            btrfs subvolume create "$MNTPOINT"/${subvol.name} ${toString subvol.extraArgs}
-            ${swapCreate "$MNTPOINT/${subvol.name}" subvol.swap}
+            SUBVOL_ABS_PATH="$MNTPOINT/${subvol.name}"
+            mkdir -p '${lib.dirOf "$SUBVOL_ABS_PATH"}'
```
Mic92, line 130: I don't think this can work nix. Cannot expand this bash variable here. Can we not just use `dirname` here?

lib/types/btrfs.nix
Mic92: I think we also should at what default system shells actually supports rather than what some out-dated standard says. Ash has a lot more features than your grandfathers sh.

lib/types/btrfs.nix
ctem: You're right. It appeared to work in my quick vm test, but I had forgotten to update the lock file prior to running it. Apologies.

The PR is now updated to use `dirname`.

Lassulus:
@mergify queue

PR #465, opened 2023-12-12
make-disk-image: use nix path-info to automatically set the imageSize

MatthewCroughan:
This may not be the best approach, but it is an approach.


lib/make-disk-image.nix
```diff
@@ -24,7 +24,14 @@ let
     util-linux
   ];
   preVM = ''
-    ${lib.concatMapStringsSep "\n" (disk: "truncate -s ${disk.imageSize} ${disk.name}.raw") (lib.attrValues nixosConfig.config.disko.devices.disk)}
+    export PATH=$PATH:${pkgs.nix}/bin:${pkgs.jq}/bin
```
Lassulus, line 27: lib.makeBinPath?

lib/make-disk-image.nix
```diff
@@ -24,7 +24,14 @@ let
...
   preVM = ''
-    ${lib.concatMapStringsSep "\n" (disk: "truncate -s ${disk.imageSize} ${disk.name}.raw") (lib.attrValues nixosConfig.config.disko.devices.disk)}
+    export PATH=$PATH:${lib.makeBinPath (with pkgs; [ nix jq ])}
+    export NIX_STATE_DIR=$TMPDIR/state
+    export NIX_CONFIG="experimental-features = nix-command"
+    nix-store --load-db < ${pkgs.closureInfo {
+      rootPaths = [ systemToInstall.config.system.build.toplevel ];
+    }}/registration
+
+    ${lib.concatMapStringsSep "\n" (disk: "truncate -s $(($(set -x; nix path-info --json ${systemToInstall.config.system.build.toplevel} | jq .[].closureSize) + ${toString disk.imageSizeExtraBytes})) ${disk.name}.raw") (lib.attrValues nixosConfig.config.disko.devices.disk)}
```
Lassulus, line 34: newlines could help to make that more readable. and probably using '' instead of "

Mic92: I would also assign variables so that in the event of a failure it's easier to spot which part failed.

iFreilicht:
This seems like a nice QoL improvement. Could you rebase it and address the comments?

MatthewCroughan:
@iFreilicht I don't have the time at the moment, but maybe you or @Enzime could

Mic92:
> This seems like a nice QoL improvement. Could you rebase it and address the comments?

I tried to come up with a more accurate algorithm, but it seems impossible. So I am not so sure if we should have this.

iFreilicht:
> I tried to come up with a more accurate algorithm, but it seems impossible.

What's the risk? Having the image size reduced automatically seems quite neat, even if it isn't perfect. I see two potential issues:

- If the result of the calculation is too small, the build will fail.
- If the calculation is automatic, an explosive increase in size wouldn't be noticed

I don't know too much about how these disk images are used in practice, so I can't tell how bad these are and how big of a pain this PR solves. What do you think?

michaelvanstraten:
Maybe as a reference, the nixpkgs `make-disk-image.nix` has a default option to automatically set the image size:

https://github.com/NixOS/nixpkgs/blob/4a7ecb93757b4b015362784e3ad2ada67291c4d4/nixos/lib/make-disk-image.nix#L534

PR #466, opened 2023-12-12
make-disk-image: make memSize configurable per disk

MatthewCroughan:
Sometimes the VM builder can run out of memory and fail to allocate memory when installing the system toplevel, this PR allows this to be adjusted by setting `disko.memSize`


Lassulus:
@mergify queue

PR #467, opened 2023-12-14
make-disk-image: inherit kernel from nixosConfiguration

MatthewCroughan:
`tests/bcachefs.nix` only works because bcachefs is baked into the kernel, without this PR, disko is not using the same kernel as the nixosConfiguration it is building an image for, which leads to edge cases, such as being able to create a bcachefs but not mount it.

In my testing, I observed that my nixosConfiguration was on kernel 6.7-rc5, whereas Disko's  `make-disk-image.nix` reported kernel 6.1 from `uname -a`, this PR synchronizes everything so that this will never happen.

This kind of thing happened in Nixpkgs with `make-single-disk-zfs-image.nix`, with the same solution
https://github.com/NixOS/nixpkgs/blob/master/nixos/lib/make-single-disk-zfs-image.nix#L224-L228

I've also added an `extraRootModules` to disko's toplevel which allows setting `disko.extraRootModules = [ "zram" ]` for example, in order to load zram in the disk builder like this:

```nix
...
                preCreateHook = ''
                  echo 0 > /sys/module/zswap/parameters/enabled
                  zramctl /dev/zram0 --algorithm zstd --size 32G
                  mkswap -U clear /dev/zram0
                  swapon --priority 100 /dev/zram0
                '';
...
```

This reduces the amount of ram required for the `nixos-install`, as well as the overhead of loading `extraRootModules` if any are provided.

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -7,6 +7,11 @@
 , checked ? false
 }:
 let
+  vmTools = pkgs.vmTools.override {
+    rootModules = ["9p" "9pnet_virtio" "virtio_pci" "virtio_blk"] ++ nixosConfig.config.disko.extraRootModules;
+    kernel = pkgs.aggregateModules
+    (with nixosConfig.config.boot.kernelPackages; [ kernel ]);
+  };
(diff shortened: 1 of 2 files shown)


Lassulus:
@mergify queue

PR #470, opened 2023-12-14
make-disk-image: allow extra dependencies to be passed into image builder

MatthewCroughan:
Without this, tools like `f2fs-tools` providing `mkfs.f2fs` for a disko config making an f2fs filesystem won't be possible to pass into the `vmTools` invocation.

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -29,6 +29,13 @@ in
       default = { };
       description = "The devices to set up";
     };
+    extraDependencies = lib.mkOption {
+      type = lib.types.listOf lib.types.package;
+      description = ''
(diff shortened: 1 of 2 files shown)


module.nix
```diff
@@ -22,6 +22,13 @@ in
       default = { };
       description = "The devices to set up";
     };
+    extraDependencies = lib.mkOption {
+      type = lib.types.functionTo (lib.types.listOf lib.types.package);
+      description = ''
+        list of extra packages to make available in the make-disk-image.nix VM builder, an example might be f2fs-tools
+      '';
+      default = [];
```
Lassulus, line 30:
```suggestion
      default = _: [];
```

Lassulus:
@mergify queue

Issue #261, opened 2023-06-10
top-level option for configuring btrfs multi-device volume


montchr:
### Context

With the following devices on a Hetzner AX52:

```
/dev/nvme0n1
/dev/nvme1n1
/dev/sda
/dev/sdb
```

I am looking to implement a layout similar to the following example:

<https://github.com/tfc/nixos-anywhere-example/blob/fd710be5255931a64a9efb9c15c9b36e6011e7b4/two-raids-on-two-disks.nix>

In contrast to that layout, I would like to use BTRFS' native support for multi-device volumes instead of using mdadm.

## Problem

I would expect to be able to configure subvolumes for a multi-device BTRFS volume at `disko.devices.btrfs` (or some more-fitting name) in the same way that other volume-like domains e.g. `mdadm`, `zpool` do:

https://github.com/nix-community/disko/blob/5d9f362aecd7a4c2e8a3bf2afddb49051988cab9/lib/default.nix#L306-L334

## Question

Is it currently possible to configure multi-device BTRFS subvolumes using Disko?

Mic92:
No this is not implemented just now, but feel free to implement it.
@Lassulus might have some guidelines on how to do this.

phaer:
Pretty much the same feature is also supported for zfs user space tools, but Disko doesn't implement it for zfs neither so far.

Mic92:
Well, you can great raids with zfs: https://github.com/nix-community/disko/blob/5d9f362aecd7a4c2e8a3bf2afddb49051988cab9/example/zfs.nix#L58

phaer:
> Well, you can great raids with zfs:

Yes, that's what I am talking about here. But with standard zfs tools you can i.e. create a mirror (or raidz) on N disks and add a cache on a faster SSD, all in the same command. I.e. 

```
zpool create my-pool raidz /dev/by-name/disk-1 /dev/by-name/disk-2 /dev/by-name/disk-3 [..] /dev/by-name/disk-N cache /dev/by-name/ssd-1
```
I believe this isn't supported by disko at the moment and seems similar to the btrfs feature in terms of nixos module code? Pretty sure there's a thread where @Lassulus and me've talked about this already but i can't find that atm.

stephenandary:
> No this is not implemented just now, but feel free to implement it. @Lassulus might have some guidelines on how to do this.

@Lassulus, after the holidays may I please have guidelines for this?

iFreilicht:
[...]

PR #482, opened 2023-12-24
btrfs multi-device raid capability support

stephenandary:
Pertaining to https://github.com/nix-community/disko/issues/261. 

Starting to think through how multi-device raid can be done easily for btrfs file systems. 

Raid 5/6 should not be supported, as they are unstable. Profiles can be seen [here](https://btrfs.readthedocs.io/en/latest/mkfs.btrfs.html#profiles)

I think the solution may be optimal if we can accept a list of mount devices, and pass in raid options through extraArgs to align with commands like:

`mkfs.btrfs -m raid10 -d raid10 /dev/sdb /dev/sdc /dev/sdd /dev/sde`

PR is not complete - just started so I can look at it after holiday.

I will revert formatting in PR eventually. Just don't feel like it at the moment.


stephenandary:
TODO: See `lib/default.nix`, there is a reference to a devices list which maps an attribute set based on device name and mount point. Can/should this be used in btrfs extraDevices option, or should I just update device option to use devices if possible?

Mic92:
It's also ok to introduce new types for btrfs raids, if it doesn't fit the current type.

stephenandary:
BTFRS raid is parity raid, and should be considered different than lvm raid, so that would make sense. 

I put a pin in this because I am not sure how disko partitions, formats, and works with fstab for uuid's yet, and it is easier to reason about in bash and rust for me

I can take a bit of time to look at it today though.

stephenandary:
@Mic92 

I think the best option here will be to close or rename this PR, and to add an example where one can pass [add](https://btrfs.readthedocs.io/en/latest/btrfs-device.html#add-new-device) and [balance](https://btrfs.readthedocs.io/en/latest/Balance.html) commands like `btrfs device add` and  `btrfs balance start -v -mconvert=raid1C4,soft -dconvert=raid10,soft mnt/` after the disks are formatted and mounted.

To implement this in btrfs.nix, you would need to ensure that swap was not defined within a btrfs partition for multidisk btrfs, because it is not supported yet. You would also need to warn or restrict about raid56 being unstable potentially. 

Is it possible for Disko to run btrfs `add` and `balance` commands after mounting? Because I think that would be lightest lift for refactor, and could potentially be done in an example with extra args or something.

stephenandary:
An example of why I say this may be challenging to refactor currently: 

```sh
# PARTITIONING

# primary disk
parted -s /dev/nvme0n1 -- mklabel gpt \
    mkpart ESP fat32 1MB 512MB \
    mkpart root btrfs 512MB -8GB \
    mkpart swap linux-swap -8GB 100% \
    set 3 esp on

# raid disks
parted --script /dev/nvme1n1 -- mklabel gpt \
    mkpart primary btrfs 0% 100% \

parted --script /dev/nvme2n1 -- mklabel gpt \
    mkpart primary btrfs 0% 100% \
    
# FORMATTING

# format boot
uuid_8=$(uuidgen | tr '[:lower:]' '[:upper:]' | head -c8) # lower case labels don't work on all systems
boot_uuid=(${uuid_8:0:4}-${uuid_8:4:9})
mkfs.fat -F 32 -n boot -i $uuid_8 /dev/nvme0n1p1

# format multidisk btrfs
btrfs_uuid=$(uuidgen)
mkfs.btrfs -f -L nixos -U $btrfs_uuid -d raid1c3 -m raid1c3 /dev/nvme0n1p2 /dev/nvme1n1p1 /dev/nvme2n1p1

# format swap - currently limitations to creating swap in btrfs for multi-disk so creating as seperate partition on single disk
swap_uuid=$(uuidgen)
mkswap -L swap -U $(uuidgen) /dev/nvme0n1p3

# MOUNTING 

# mount btrfs 
mount -o compress=zstd,subvol=root -U $btrfs_uuid /mnt
mkdir -p /mnt/{home,nix,persist,log}
mount -o compress=zstd,subvol=home -U $btrfs_uuid /mnt/home
mount -o compress=zstd,noatime,subvol=nix -U $btrfs_uuid /mnt/nix
mount -o compress=zstd,noatime,subvol=persist -U $btrfs_uuid /mnt/persist
mount -o compress=zstd,noatime,subvol=log -U $btrfs_uuid /mnt/log

# mount boot
mkdir -p /mnt/boot
mount -U $boot_uuid /mnt/boot

# mount swap
swapon -U $swap_uuid
```

stephenandary:
unless we could convert device to "devices" and pass a list - that may be easy. I am new to nix though, so I will wait for your opinion.

stephenandary:
Sorry about formatting, I forgot to turn it off again, but put rough thoughts into new commit.

stephenandary:
@Mic92 I did my best to find a simple solution today, but I get stuck at mounting. I use uuid to mount my multidisk btrfs, and I am not sure how to work around that without rewriting existing code.

stephenandary:
Closing PR as I am not using Disko right now so I won't be putting in more work, but if anything in this useful and needs discussion, please let me know.

Mic92:
Pull requests welcome from other people. I don't use btrfs myself though.

PR #487, opened 2023-12-30
zfs: respect `canmount`

### LINKED ISSUE #486: zfs support regression: cannot unmount X: not currently mounted
My [disko config](https://github.com/ghostbuster91/dot-files/blob/6ad7b34d732569dc103f4fdaf674fc692b1d9245/machines/focusM2/disko-config.nix#L1) doesn't work anymore with the latest version from master. 

when doing disks setup using:
```
sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko -- --mode disko /tmp/disko-config.nix --arg disks '["/dev/nvme0n1" "/dev/nvme1n1"]'
```

it fails midway with a following error:
```
+ zpool create -f rpool1 -R /mnt -o ashift=12 -o autotrim=on -O atime=off -O canmount=off /dev/mapper/nixos
+ zfs unmount rpool1
cannot unmount `rpool1`: not currently mounted
```

then calling `mount | grep /mnt` returns only standard overlays. 

The regression was introduced somewhere between https://github.com/nix-community/disko/compare/8759b5bf0af8c21685e1da7f20c40eeb79eaeec2...e526d4e6533d006e4386130df7729830770889b3
 
Later I can minimize the issue further but for now I just wanted to start with something.

Enzime:
-  tests: check `canmount` ZFS flag is respected
- zpool: only `unmount` when root dataset is mounted
Fixes #486
- zfs: respect `canmount`
Also refactor `zpool` type to use `zfs_fs` to construct the root dataset
for the zpool, which means we no longer need to duplicate the dataset
create and mount logic inside the `zpool` type.

cc @phaer

### DIFF
--- a/example/non-root-zfs.nix
+++ b/example/non-root-zfs.nix
@@ -59,6 +59,22 @@
           };
         };
       };
+      a = {
+        type = "disk";
+        device = "/dev/sda";
+        content = {
+          type = "gpt";
+          partitions = {
+            zfs = {
+              size = "100%";
+              content = {
+                type = "zfs";
+                pool = "storage2";
+              };
+            };
+          };
+        };
+      };
     };
     zpool = {
       storage = {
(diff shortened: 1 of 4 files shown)


Mic92:
@mergify queue

phaer (commented):
Late to the party due to a few days away for computers, but thanks for the quick fix! :)

example/non-root-zfs.nix
```diff
@@ -59,6 +59,22 @@
           };
         };
       };
+      a = {
+        type = "disk";
+        device = "/dev/sda";
+        content = {
+          type = "gpt";
```
T9ipa, line 66: 🪃👍🔥🎁image[[image]()]()

PR #488, opened 2023-12-31
luks: apply flags set in `config.settings`

0xAdk:
Along with adding flags to cryptsetup open I also unified the two calls to use a single binding, this also fixes an issue with one not using `config.extraFormatArgs` from #304

I added `--persistent` to the open call after format `_create` since that makes sense you'd want to keep those set.

The changes to `keyFileArgs` binding were due to it having a erroneous newline which caused issue when adding any flags after it was used; The `lib.forEach config.additionalKeyFiles` line was relying on that newline, so I put the `cryptsetup luksAddKey` in a string block to add that newline back in.


Mic92:
@mergify queue

PR #491, opened 2024-01-03
lib/types/gpt: accept GUID type to set the partition type

r-vdp:
See [the wikipedia article on GUID types][wiki_guid] and the sgdisk man page on `--typecode`.

[wiki_guid]: https://en.wikipedia.org/wiki/GUID_Partition_Table#Partition_type_GUIDs


Lassulus:
@mergify queue

PR #498, opened 2024-01-08
docs: mention EF00 partition type in README.md

raphaelr:
Otherwise systemd-boot will refuse to install the bootloader.


Lassulus:
@mergify queue

PR #499, opened 2024-01-08
bcachefs: add support for subvolumes

adamcstephens:
Picked up the work started by Raito.

Closes: #449


lib/types/bcachefs.nix
```diff
@@ -0,0 +1,108 @@
...
+      type = lib.types.functionTo diskoLib.jsonType;
+      default = _dev: { };
+      description = "Metadata";
+    };
+    _create = diskoLib.mkCreateOption {
+      inherit config options;
+      default = ''
+        modprobe bcachefs
+        mkfs.bcachefs ${toString config.extraArgs} ${config.device}
+        grep bcachefs /proc/filesystems
```
Mic92, line 58: Was this added for debugging?

Lassulus:
do we want to extend this also for multidisks? if yes then we need to split the bcachefs type into a type which marks the disk for including it in a bcachefs array and another toplevel type to assemble the array. The toplevel type would be mostly the same as the bcachefs type here, but with support for multiple disks.

Mic92:
I think even if we don't implement this feature fully just now, we should already prepare it in this direction. Otherwise we need to break compatibility with people soon again.

lib/types/bcachefs.nix
Mic92: I would rely on bcachefs commands to come up with an error except if that one is hard to understand.

adamcstephens:
The docs provide a complex example of how one formats a multi-disk filesystem, https://bcachefs-docs.readthedocs.io/en/latest/mgmt-formatting.html . Based on that, I would probably suggest we create our own construct, a multi-disk toplevel type named something like `bcachefs_multi` to clearly distinguish it from a simple filesystem. I would then suggest we duplicate the subvolume section into that new type instead of trying to overload a single `bcachefs` type in the two very different use cases. I'm also not excited about requiring users of bcachefs to build a toplevel config when they just want a simple filesystem, maybe with subvolumes, that the fs clearly targets as well.

In that case, I'm failing to see how what is done here would break compatibility for a separate type. Thoughts?

example/bcachefs-subvol.nix
```diff
@@ -0,0 +1,49 @@
...
+                type = "filesystem";
+                format = "vfat";
+                mountpoint = "/boot";
+              };
+            };
+            root = {
+              name = "root";
+              end = "-0";
+              content = {
+                type = "bcachefs";
```
Mic92, line 23: If we have two types, than we should rename this one to `bcachefs_single` or similar. Otherwise it creates confusion when people try to use `bcachefs` for multi disk setups.

tests/bcachefs-subvol.nix
```diff
@@ -0,0 +1,43 @@
...
+    machine.fail("bcachefs subvolume snapshot /srv /srv.snap"); 
+  '';
+
+  # so that the installer boots with a bcachefs enabled kernel
+  extraInstallerConfig = {
+    # disable zfs so we can support latest kernel
+    nixpkgs.overlays = [
+      (_final: super: { zfs = super.zfs.overrideAttrs (_: { meta.platforms = [ ]; }); })
+    ];
+    boot.kernelPackages = pkgs.lib.mkForce pkgs.linuxPackages_testing;
```
Sntx626, line 31: This can be changed to `pkgs.linuxPackages_latest` - testing is no longer required on unstable

tests/bcachefs-subvol.nix
```diff
@@ -0,0 +1,43 @@
...
+  };
+
+  extraSystemConfig = {
+    environment.systemPackages = [ pkgs.bcachefs-tools ];
+
+    # disable zfs so we can support latest kernel
+    nixpkgs.overlays = [
+      (_final: super: { zfs = super.zfs.overrideAttrs (_: { meta.platforms = [ ]; }); })
+    ];
+    boot.kernelPackages = pkgs.lib.mkForce pkgs.linuxPackages_testing;
```
Sntx626, line 41: This can be changed to pkgs.linuxPackages_latest - testing is no longer required on unstable

Sntx626 (changes requested):
Looks mostly good.

I agree that we should start separating between single- and multi-disk bcachefs types with this PR, even if it doesn't implement the functionality.

Also I'm not sure, but is there an assertion that allows us to check whether the user has defined a bcachefs enabled kernel (i.e. by comparing versions, even if inputs.nixpkgs is overwritten)?
I think that would be better than just overwriting whatever the user defines.

Lassulus:
We probably need a generic solution for mutli device support which integrates nicely with the single disk usecase.

PR #500, opened 2024-01-08
wipe device before creating partition table

phaer:
this erases not only pre-existing filesystems, but all magic strings recognized by libblkid(3), including raid superblocks

### DIFF
--- a/lib/types/table.nix
+++ b/lib/types/table.nix
@@ -82,6 +82,7 @@
     _create = diskoLib.mkCreateOption {
       inherit config options;
       default = ''
+        wipefs --all --force ${config.device}
         parted -s ${config.device} -- mklabel ${config.format}
         ${lib.concatStrings (map (partition: ''
           ${lib.optionalString (config.format == "gpt") ''


Lassulus:
this should probably happen in disk-deactivate

phaer:
Ah, in fact we should already do so in https://github.com/nix-community/disko/blob/f0a3425a7b173701922e7959d8bfb136ef53aa54/disk-deactivate/disk-deactivate.jq#L29

The thing is: we don't, that's why I didn't even check whether disko always runs deactivation (it does) and ignored that call.

But the real issue seems to be that the deactivation script seems to be a silent noop if one uses /dev/disk/by-id due to this line https://github.com/nix-community/disko/blob/f0a3425a7b173701922e7959d8bfb136ef53aa54/disk-deactivate/disk-deactivate.jq#L67

e.g. 
`lsblk --output-all --json | jq -r --arg disk_to_clear "/dev/disk/by-id/ata-Samsung_SSD_750_EVO_500GB_S3CJNWAHC85163B" -f $disko/disk-deactivate.jq
` produces empty output, wheres the same command with "/dev/sdb" produces a deactivation script as intended.

phaer:
Right & it does - i just can't read correctly; sorry for the noise

PR #502, opened 2024-01-09
deactivate: stop raid if disk is a member of it

phaer:
this is useful in cases where the target machine has a pre-configured RAID setup, as otherwise disko would wipe the partition tables and create new ones but would not be able to inform the kernel about it, as the old raid was still in use.

it's the spiritual successor to #500 ;)

### DIFF
--- a/disk-deactivate/disk-deactivate.jq
+++ b/disk-deactivate/disk-deactivate.jq
@@ -27,7 +27,7 @@ def deactivate:
   if .type == "disk" then
     [
       # If this disk is a member of raid, stop that raid
-      "mdadm --stop $(lsblk \(.path) -l -p -o type,name | awk 'match($1,\"raid.*\") {print $2}')",
+      "lsblk \(.path) -l -p -o type,name | awk 'match($1,\"raid.*\") {print $2}' | xargs -r mdadm --stop",
       # Remove all file-systems and other magic strings
       "wipefs --all -f \(.path)",
       # Remove the MBR bootstrap code
--- a/disk-deactivate/disk-deactivate.jq
+++ b/disk-deactivate/disk-deactivate.jq
@@ -27,7 +27,8 @@ def deactivate:
   if .type == "disk" then
     [
       # If this disk is a member of raid, stop that raid
-      "lsblk \(.path) -l -p -o type,name | awk 'match($1,\"raid.*\") {print $2}' | xargs -r mdadm --stop",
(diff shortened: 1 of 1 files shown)


disk-deactivate/disk-deactivate.jq
```diff
@@ -26,6 +26,9 @@ def remove:
 def deactivate:
   if .type == "disk" then
     [
+      # If this disk is a member of raid, stop that raid
+      "mdadm --stop $(lsblk \(.path) -l -p -o type,name | awk 'match($1,\"raid.*\") {print $2}')",
```
Mic92, line 30: This should print no error in case there is no device:

```suggestion
      "lsblk /dev/nvme0n1 -l -p -o type,name | awk 'match($1,"raid.*") {print $2}' | xargs -r mdadm --stop",
```

Mic92: What if the raid is still mounted? Can it be stopped?

Mic92: Also this might be less of a concern for the common installer use case where mdadm's are assembled but not mounted.

Mic92: Ah. We could also do an umount of the raid device prior to stopping? `umount /dev/md1` should work.

phaer: Do we want to do that? We don't umount normal filesystems here either? 

I mainly thought about the scenario where one uses disko with nixos-anywhere: after kexec, the raid is "running", but not mounted.

Mic92: We do unmount other filesystems in the same jq script that are used by the block device.

disk-deactivate/disk-deactivate.jq
```diff
@@ -26,6 +26,9 @@ def remove:
 def deactivate:
   if .type == "disk" then
     [
+      # If this disk is a member of raid, stop that raid
+      "lsblk /dev/nvme0n1 -l -p -o type,name | awk 'match($1,"raid.*") {print $2}' | xargs -r mdadm --stop",
```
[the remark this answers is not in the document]
phaer: haha, thanks! should probably not click "commit suggestion" on my phone while on the go.

disk-deactivate/disk-deactivate.jq
phaer: Ah, you are right. Thanks!

Mic92 (approved):
Looks ok, but I haven't tested it.

Lassulus:
@mergify queue

PR #506, opened 2024-01-12
make-disk-image: umount after install and explicitly set raw format

wfdewith:
This PR includes two minor changes to the `make-disk-image` script.

The `vmTools` in nixpkgs shuts the VM down with `poweroff -f`, which does not cleanly umount the filesystems. This may lead to filesystem corruption on the newly created image. To fix this, we add an explicit `umount` call after the installer is finished.

QEMU warns if `format=raw` is implied, and protects block 0 from getting overwritten. I did not encounter any issues because of that protection, but I think it is better to be explicit here. It also gets rid of the warning.


Lassulus:
@mergify queue

PR #508, opened 2024-01-13
Rework Hybrid MBR https://github.com/nix-community/disko/pull/168

danjujan:
Since https://github.com/nix-community/disko/pull/168 stalled I decided to rework it to the current status quo.
Changes compared to the previous approach:
- Moved the hybrid mbr code to the gpt module. That is way cleaner than introducing new types imho. The types just make sense when combined with gpt anyway.
- Removed the need for the `gptPartitionNumber` as we now take the index of the list entry.

@oneingan Hope you don't mind.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -194,7 +194,7 @@ let
           || lib.hasSuffix "Hook" n
           || isAttrsOfSubmodule o
           # TODO don't hardcode diskoLib.subType options.
-          || n == "content" || n == "partitions" || n == "datasets" || n == "swap" || n == "hybrid_partitions"
+          || n == "content" || n == "partitions" || n == "datasets" || n == "swap"
         );
       in
       lib.toShellVars
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -84,9 +84,11 @@ let
         "${dev}${toString index}" # /dev/md/raid1 style
       else if match "/dev/mapper/.+" dev != null then
         "${dev}${toString index}" # /dev/mapper/vg-lv1 style
+      else if match "/dev/loop[[:digit:]]+" dev != null
+        then "${dev}p${toString index}" # /dev/mapper/vg-lv1 style
       else
         abort ''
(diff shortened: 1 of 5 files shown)


lib/types/gpt.nix
```diff
@@ -108,6 +146,7 @@ in
     _create = diskoLib.mkCreateOption {
       inherit config options;
       default = ''
+        sgdisk -Z ${config.device}
```
Mic92, line 149: isn't our wipefs already doing something like that?

lib/types/gpt.nix
```diff
@@ -89,6 +89,44 @@ in
...
+          };
+          _index = lib.mkOption {
+            internal = true;
+            default = diskoLib.indexOf (x: x.mbrPartitionType == hp.config.mbrPartitionType && x.mbrBootableFlag == hp.config.mbrBootableFlag) (config.hybrid_partitions) 0;
+          };
+        };
+      }));
+      default = [ ];
+      description = "Entries to add to the Hybrid MBR table";
+    };
```
Lassulus, line 129: I think it would be better to include those flags in the parition type (maybe under a submodule .hybdrid.)
Lists are a bit unflexible in nix and overriding/updating them out of bounds is next to impossible.

lib/types/gpt.nix
Lassulus: we should probably call sgdisk -Z then in the disk_deactivate script instead of here

Lassulus (changes requested):

lib/default.nix
```diff
@@ -192,7 +194,7 @@ let
           || lib.hasSuffix "Hook" n
           || isAttrsOfSubmodule o
           # TODO don't hardcode diskoLib.subType options.
-          || n == "content" || n == "partitions" || n == "datasets" || n == "swap"
+          || n == "content" || n == "partitions" || n == "datasets" || n == "swap" || n == "hybrid_partitions"
```
Mic92, line 197: Is this case still used? I cannot see the string anywhere else.

Lassulus (approved):

Lassulus:
thanks a alot!

PR #509, opened 2024-01-14
lib/types/gpt: do not hard-code alignment to 2048 sectors

### LINKED ISSUE #497: Hardcoded 2048-sector alignment is excessive for 4Kn disks
The gpt partition table support in disko seems to hardcode a 2048-sector alignment:

https://github.com/nix-community/disko/blob/8a9e89d4669c8b6820ae7c7574f68908892fa5f3/lib/types/gpt.nix#L113

This only makes sense for 512-byte sector disks; for 4K-native disks this results in 8 MiB alignment which is excessive.

---

Does it even make sense to explicitly specify the alignment value, given that sgdisk already defaults to 1 MiB alignment using whatever sector size is reported by the disk?

intelfx:
On 4Kn disks, 2048-sector alignment is equivalent to 8 MiB which is
excessive. Moreover, there is no point in specifying _any_ alignment
at all because sgdisk defaults to 1 MiB already.

Fixes #497.

### DIFF
--- a/lib/types/gpt.nix
+++ b/lib/types/gpt.nix
@@ -110,7 +110,6 @@ in
       default = ''
         ${lib.concatStrings (map (partition: ''
           sgdisk \
-            --set-alignment=2048 \
             --align-end \
             --new=${toString partition._index}:${partition.start}:${partition.end} \
             --change-name=${toString partition._index}:${partition.label} \


Mic92:
@mergify queue

PR #535, opened 2024-02-15
Prevent `disk-deactivate` from failing when `disko` is run in a systemd service

Jayman2000:
See the individual commit messages for details. I ran into this issue while trying to create a custom NixOS installation ISO that can do unattended installs.


Mic92:
@mergify queue

PR #538, opened 2024-02-17
types table: add deprecation warning

Lassulus:


Lassulus:
@mergify rebase

nifoc:
Is there some kind of documentation/explanation of how to move from `table` to `gpt`?

I just got this error but it's very unclear to me how I should deal with it.

Lassulus:
Ah there pretty sure is. But I have not done it myself yet. If you don't want to reformat, it should be fairly straight forward and I will try to write something up in the next week. Sorry for the noise of the warning until then

kira-bruneau:
Is the table only deprecated for gpt partition tables? This warning still comes up if you're using mbr/msdos, but I can't find any other alternative.

Lassulus:
Yeah. Maybe hybrid gpt partitions can be used in that case. But I would declare legacy/msdos tables in general deprecated? What is your usecase for them?

kira-bruneau:
Mainly just using it as a declarative configuration for mounting an already formatted drive.

I tried using the gpt type, but msdos doesn't support partition labels, so it fails when trying to mount through `/dev/disk/by-partlabel`.

I could also just use the nixos filesystem module directly, but I've been trying to keep all disk-specific configuration in disko.

Lassulus:
hmm I guess in that case you can ignore the warning, or we could move it to the create argument, but then people only see it when creating new drives and there it could be already to late.

Issue #551, opened 2024-03-02
Random boot failues due to partition confusion because disko now relies on `by-partlabel`


oddlama:
I just updated disko and switched to `type = "gpt";` which a warning informed me to do. Now my device now randomly fails to boot. This is the case whenever I have another USB stick plugged into my system that coincidentally shares some partition labels (like `root`) with the partition labels of my actual system.

The new `gpt.nix` implementation seems to rely solely on `/dev/disk/by-partlabel` instead of `/dev/disk/by-id` which was used previously. As a consequence of this change, my system now sometimes confuses a USB stick partition and the system confusion which share the same name, which seemingly dependent on udev device initialization order. So sometimes my system uses the wrong one and it doesn't boot.

Is there any way to switch back to `/dev/disk/by-id` ? I'd prefer to have stable and unique ids to refer to my partitions that don't cause random clashes. If there is absolutely no way to use `by-id` now because we don't know the partition order anymore, would it be possible to require manually setting a unique id on the parent gpt layout that is prepended to all labels? This mostly mitigate the issue, although one could still easily craft a USB stick to disrupt a system.

Lassulus:
hmm, you would need to override the device of each partition. you can set it to /dev/disk/by-id/$my_id-part$x. Another alternative would be to override the disk name. so something like:
```
disko.devices.disk.name = builtins.substring 0 10 (builtins.hashString "sha256" config.disko.devices.disk.device);
```

oddlama:
> you can set it to /dev/disk/by-id/$my_id-part$x

That's what I did yesterday to work around the issue for the time being, but I was hoping there would be a better approach. The current partition numbers obviously only correspond to the order with legacy tables from before and if I were to resetup now, then the partitions would be in a different order because of alphabetical sorting and it would break again. Adding a partition would also break it if it isn't ordered last. Previously we could choose where to place the partitions so this would not happen.

I then tried to use the canonical device name via `name = "nvme-Samsung_SSD_980_PRO_2TB_SXXENFXXWXXXXX"` but noticed that this causes the partition labels to become very long because of the `disk-<diskname>-<partname>` format. And unfortunately those long labels seem to be unsupported somewhere (in the kernel ?).

>     disko.devices.disk.root.name = builtins.substring 0 10 (builtins.hashString "sha256" config.disko.devices.disk.root.device);
> that should ensure a unique id according to the device?

[...]

PR #553, opened 2024-03-04
types gpt: hash long device names

Lassulus:
this should mitigate some of the problems in https://github.com/nix-community/disko/issues/551

also we fix the test framework with recent versions of nixpkgs

### DIFF
--- a/lib/tests.nix
+++ b/lib/tests.nix
@@ -223,10 +223,15 @@ let
                   disk_flags += f' -drive file={oldmachine.state_dir}/empty{i}.qcow2,id=drive{i + 1},if=none,index={i + 1},werror=report'
                   disk_flags += f' -device virtio-blk-pci,drive=drive{i + 1}'
               return disk_flags
-          def create_test_machine(oldmachine=None, args={}): # taken from <nixpkgs/nixos/tests/installer.nix>
+          def create_test_machine(oldmachine, args={}): # taken from <nixpkgs/nixos/tests/installer.nix>
+              startCommand = "${pkgs.qemu_test}/bin/qemu-kvm"
+              startCommand += " -cpu max -m 1024 -virtfs local,path=/nix/store,security_model=none,mount_tag=nix-store"
+              startCommand += disks(oldmachine, ${toString num-disks})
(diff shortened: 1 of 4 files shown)


lib/types/gpt.nix
```diff
@@ -55,7 +55,13 @@ in
           };
           label = lib.mkOption {
             type = lib.types.str;
-            default = "${config._parent.type}-${config._parent.name}-${partition.config.name}";
+            default = let
+              limit = 36; # not sure why it's the limit but it seems to break with 37 characters
```
Mic92, line 59: For GPT it's actually 72 characters inside the label?

Mic92: Ah, found it: https://en.wikipedia.org/wiki/GUID_Partition_Table


56 (0x38) 	72 bytes 	Partition name (36 [UTF-16](https://en.wikipedia.org/wiki/UTF-16)LE code units)

Mic92: So fdisk encodes the name as UTF-16.

lib/tests.nix
```diff
@@ -223,10 +223,15 @@ let
                   disk_flags += f' -drive file={oldmachine.state_dir}/empty{i}.qcow2,id=drive{i + 1},if=none,index={i + 1},werror=report'
                   disk_flags += f' -device virtio-blk-pci,drive=drive{i + 1}'
               return disk_flags
-          def create_test_machine(oldmachine=None, args={}): # taken from <nixpkgs/nixos/tests/installer.nix>
+          def create_test_machine(oldmachine, args={}): # taken from <nixpkgs/nixos/tests/installer.nix>
+              startCommand = "${pkgs.qemu_test}/bin/qemu-kvm"
+              startCommand += " -cpu max -m 1024 -virtfs local,path=/nix/store,security_model=none,mount_tag=nix-store"
+              startCommand += disks(oldmachine, ${toString num-disks})
+              ${lib.optionalString efi ''
```
Mic92, line 230: is the efi branched used in the test?

Lassulus: Not sure if the new ones do it though

PR #566, opened 2024-03-11
disko-install: fix case where lib in NixOS is not defined

Mic92:
This seems like a recent addition that we don't have on all NixOS machines.


Lassulus (approved):

Mic92:
@mergify refresh

Mic92:
@mergify queue

PR #568, opened 2024-03-12
types *: make create idempotent

Lassulus:
rework of https://github.com/nix-community/disko/pull/435 with a smaller (and more logical) scope

### DIFF
--- a/example/zfs.nix
+++ b/example/zfs.nix
@@ -52,7 +52,7 @@
           "com.sun:auto-snapshot" = "false";
         };
         mountpoint = "/";
-        postCreateHook = "zfs snapshot zroot@blank";
+        postCreateHook = "zfs list -t snapshot -H -o name | grep -E '^zroot@blank$' || zfs snapshot zroot@blank";
 
         datasets = {
           zfs_fs = {
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -31,7 +31,11 @@ let
   swapCreate = mountpoint: swap:
     lib.concatMapStringsSep
       "\n"
-      (file: ''btrfs filesystem mkswapfile --size ${file.size} "${mountpoint}/${file.path}"'')
+      (file: ''
+      if ! test -e "${mountpoint}/${file.path}"; then
+        btrfs filesystem mkswapfile --size ${file.size} "${mountpoint}/${file.path}"
+      fi
+      '')
       (lib.attrValues swap);
 
 in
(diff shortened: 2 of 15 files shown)


example/zfs.nix
```diff
@@ -52,7 +52,7 @@
           "com.sun:auto-snapshot" = "false";
         };
         mountpoint = "/";
-        postCreateHook = "zfs snapshot zroot@blank";
+        postCreateHook = "zfs snapshot zroot@blank || :";
```
Mic92, line 55:
```suggestion
        postCreateHook = "zfs list -t snapshot -H -o name | grep -E '^zroot@blank$' || zfs snapshot zroot@blank";
```

lib/types/btrfs.nix
```diff
@@ -112,26 +116,33 @@ in
...
-            mkdir -p "$(dirname "$SUBVOL_ABS_PATH")"
-            btrfs subvolume create "$SUBVOL_ABS_PATH" ${toString subvol.extraArgs}
-            ${swapCreate "$SUBVOL_ABS_PATH" subvol.swap}
-          )
-        '') (lib.attrValues config.subvolumes)}
+        # create the filesystem only if the device seems empty
+        if ! (blkid '${config.device}' -o export | grep -q '^TYPE='); then
+          mkfs.btrfs "${config.device}" ${toString config.extraArgs}
+        fi
+        if (blkid "${config.device}" -o export | grep -q '^TYPE=btrfs$'); then
```
kira-bruneau, line 123: I noticed that this is causing ```syntax error near unexpected token `fi'``` when formatting a btrfs filesystem without any swap or subvolume configuration.

I guess this all should be gaurded by `config.swap` & `config.subvolumes` being set? It's a bit annoying bash doesn't support empty if statements.

gabevenberg:
Is there documentation on how to use this?

lib/types/btrfs.nix
iFreilicht: Simplest fix would be `echo -n ""` in the first line.

iFreilicht:
@gabevenberg Basically, run this:

```
disko --mode format --dry-run disko.nix
```

This outputs a path to the script that will be run once you remove `--dry-run` **Review the that script first!** As this is a new and relatively untested feature, there might be bugs that cause it to change things you don't want it to change. Though when I did it, everything looked good 👍 

Then, run the same command without `--dry-run`.

Be aware that this is only additive. When you add drives and partitions, they will be formatted as desired, but when you remove them from the config, they just become unmanaged. Same for zfs datasets and their options. Adding options will set them, but removing options will not remove them from the dataset.

adam248:
@gabevenberg 
Just to confirm, when you say "only additive" that means changing an existing disk from ext4 to btrfs will not work either.
I just did a test on a VM and it appears that it did not change the format.

So to be 100% clear... of the possible CRUD operations it will only do a Create but not the Update (modify) or Delete existing partitions when using `--mode format`. Thanks

adam248:
@iFreilicht i just rebooted the VM and now it doesn't boot at all. it appears that it may have reformated the root drive even though i didn't change that part of the disk-config.nix.

Mic92:
No we don't support ext4 to btrfs conversion for now.

iFreilicht:
> So to be 100% clear... of the possible CRUD operations it will only do a Create but not the Update (modify) or Delete existing partitions when using `--mode format`. Thanks

Not quite. Update is partially supported for some things like updating the mountpoint of a zfs dataset. The cases where this works are limited, though and there is no comprehensive list of this. That's why I specifically wrote in my instructions that you should use `--dry-run` and read the resulting script before running anything. This way you will know exactly what changes are made to your system.

> This is strange as the only thing I did after booting was try the `disko --mode format disko-config.nix` after changing one of the data drives from 'ext4' to 'btrfs' then tried a reboot.

Are you sure that's all you did? Maybe you also did a `nixos-rebuild` after updating the `disko-config.nix`? That updates the fstab, which specifies which filesystem driver a partition should be mounted with, which will fail if there's a mismatch.

TheRealGramdalf:
> Be aware that this is only additive.

For future improvements, it might be possible to use custom ZFS properties, zfs history, and/or some other form of examining zfsprops to check what the delta changes are for additional error handling (i.e. dataset has been modified since last run, refusing to overwrite), viewing changes (i.e. datasets exist on device that don't exist in disko config) etc. Just a random thought I had.

PR #579, opened 2024-03-22
Update quickstart.md

3Quatorze:
Fix `nixos-install` command invocation, it must use the root dir defined on `nixos-generate-config`


Lassulus:
isn't that the default if you use nixos-install?

3Quatorze:
Yes you're right, last time I install Nixos I had to define it but it was a long time ago.
Closing the merge request

PR #592, opened 2024-04-05
Add missing dependency gnused

jmbaur:
gnused is used in the disk-deactivate.jq script, but was not defined in the list of dependencies.


Lassulus (approved):

Lassulus:
@mergify queue

PR #593, opened 2024-04-05
Use existing bash instance when running destroyScript

jmbaur:
The destroy script already has a bash interpreter in use, but when the `disk-deactivate` script is ran, it calls out to /usr/bin/env to query for a bash interpreter. When running the destroyScript in unique places (such as the stage-1 initrd), /usr/bin/env might not exist, so we can make destroyScript more self-contained by reusing the same bash.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -449,7 +449,7 @@ let
 
               # shellcheck disable=SC2043
               for dev in ${toString (lib.catAttrs "device" (lib.attrValues devices.disk))}; do
-                ${../disk-deactivate}/disk-deactivate "$dev"
+                $BASH ${../disk-deactivate}/disk-deactivate "$dev"
               done
             '';
           };


Lassulus:
hmm, not sure how portable that is. but it works in the tests, so let's hope it doesn't break anyones usecase

Lassulus:
@mergify queue

lib/default.nix
```diff
@@ -449,7 +449,7 @@ let
 
               # shellcheck disable=SC2043
               for dev in ${toString (lib.catAttrs "device" (lib.attrValues devices.disk))}; do
-                ${../disk-deactivate}/disk-deactivate "$dev"
+                $BASH ${../disk-deactivate}/disk-deactivate "$dev"
```
Mic92, line 452: Do we need this?

```suggestion
                ${BASH:-} ${../disk-deactivate}/disk-deactivate "$dev"
```

jmbaur: Applied this change

Lassulus:
@mergify queue

PR #597, opened 2024-04-08
types btrfs: fix format without swap or subvolumes

kira-bruneau:
Hi! Recently when trying to format a btrfs drive I ran into ```syntax error near unexpected token `fi'```.

It looks like #568 introduced a change that generates an empty if statement if neither the swap config or subvolumes are set and this just adds a guard to avoid adding that empty if block if neither are set.

### DIFF
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -120,29 +120,31 @@ in
         if ! (blkid '${config.device}' -o export | grep -q '^TYPE='); then
           mkfs.btrfs "${config.device}" ${toString config.extraArgs}
         fi
-        if (blkid "${config.device}" -o export | grep -q '^TYPE=btrfs$'); then
-          ${lib.optionalString (config.swap != {}) ''
-            (
-              MNTPOINT=$(mktemp -d)
-              mount ${device} "$MNTPOINT" -o subvol=/
-              trap 'umount $MNTPOINT; rm -rf $MNTPOINT' EXIT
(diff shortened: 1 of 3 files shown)


lib/types/btrfs.nix
```diff
@@ -120,29 +120,31 @@ in
...
-              SUBVOL_ABS_PATH="$MNTPOINT/${subvol.name}"
-              mkdir -p "$(dirname "$SUBVOL_ABS_PATH")"
-              if ! btrfs subvolume show "$SUBVOL_ABS_PATH" > /dev/null 2>&1; then
-                btrfs subvolume create "$SUBVOL_ABS_PATH" ${toString subvol.extraArgs}
-              fi
-              ${swapCreate "$SUBVOL_ABS_PATH" subvol.swap}
-            )
-          '') (lib.attrValues config.subvolumes)}
-        fi
+        ${lib.optionalString (config.swap != {} && config.subvolumes != []) ''
```
Lassulus, line 123:
```suggestion
        ${lib.optionalString (config.swap != {} && config.subvolumes != {}) ''
```

Lassulus: subvolume is an attrset

kira-bruneau: Oops right thanks!

Lassulus:
it would also be nice to have an example config and a test which triggers that error. so we can detect regressions in the future :)

Lassulus:
@mergify queue

PR #608, opened 2024-04-20
disko-install: Add --system-option flag. Fix EF02 priority everywhere…

Qubasa:
…. Nix fmt.

I added the --system-option flag that enables a user to dynamically add system nix options. This is useful for authorized keys for example. I also added the priority flag on every EF02 partition, as else sticks wouldn't boot. Because of nix fmt more files changed then I expected sorry ^^

### DIFF
--- a/disko-install
+++ b/disko-install
@@ -19,10 +19,7 @@ Usage: $0 [OPTIONS]
   --write-efi-boot-entries    Write EFI boot entries to the NVRAM of the system for the installed system.
                               Specify this option if you plan to boot from this disk on the current machine,
                               but not if you plan to move the disk to another machine.
-  --system-option KEY VALUE   Pass the specified system option to the NixOS configuration.
-                              This option can be specified multiple times.
-                              For example, to set the authorizedKeys, use
-                              --system-option users.users.root.openssh.authorizedKeys.keys '[ "ssh-rsa ..." ]'
+  --system-config JSON        Merges the specified JSON object into the NixOS configuration.
 EOF
 }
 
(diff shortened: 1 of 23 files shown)


disko-install
```diff
@@ -19,6 +19,10 @@ Usage: $0 [OPTIONS]
   --write-efi-boot-entries    Write EFI boot entries to the NVRAM of the system for the installed system.
                               Specify this option if you plan to boot from this disk on the current machine,
                               but not if you plan to move the disk to another machine.
+  --system-option KEY VALUE   Pass the specified system option to the NixOS configuration.
+                              This option can be specified multiple times.
+                              For example, to set the hostname, use
+                              --system-option users.users.root.openssh.authorizedKeys.keys '[ "ssh-rsa ..." ];'
```
phaer, line 25: Is there a reason to require the semicolons here? Why don't we either add them in the serialization step below or to just let user pass a full nix attrset as a string here?

phaer (commented):
Could you split that into one commit per change, i.e. 3 in this case? Makes it easier to review code & style changes separately. :)

disko-install
Qubasa: @phaer Hmm you are right I should add the semicolon in the serialization step. I want to make the option human readable so that's why I don't want a full attrset.

Qubasa:
@phaer Done

disko-install
```diff
@@ -35,6 +39,19 @@ serialiaseArrayToNix() {
   echo "$nixExpr"
 }
 
+serialiaseArrayToNixUnquoted() {
```
Lassulus, line 42: this seems a bit hacky, maybe we should use json instead?

Qubasa: Ahh yes, json is a better way. Will do

Qubasa:
@Lassulus Done

install-cli.nix
```diff
@@ -49,7 +52,7 @@ let
         {
           boot.loader.efi.canTouchEfiVariables = lib.mkVMOverride writeEfiBootEntries;
           boot.loader.grub.devices = lib.mkVMOverride diskoSystem.config.boot.loader.grub.devices;
-        }
+        } // (builtins.fromJSON extraSystemConfig)
```
Mic92, line 55: This makes conflict error messages nicer:

```suggestion
          imports = [
             { _file = "disko-install --system-config"} // (builtins.fromJSON extraSystemConfig)
           ];
        }
```

Lassulus:
@mergify queue

Mic92:
@mergify queue

PR #609, opened 2024-04-20
Allow more ways to detect disk to deactivate

immae:
This MR adds more ways to detect disk to deactivate, by using /dev/disk elements. This allows to specify disk paths by /dev/disk/by-id in the config and allow disko to find them when it tries to deactivate them


Mic92:
@mergify queue

PR #625, opened 2024-05-06
Fix building images when non-binary reproducible builds are present                 

Mic92:
Closure-info has one flaw, it's possible that this file contains stale              
information when build are not reproducible and the local checksum                  
doesn't match what the remote build created.                                        
The work-around here is that we disregard the hashes and re-compute with            
what nar's we actually have locally instead.


lib/make-disk-image.nix
```diff
@@ -55,7 +57,7 @@ let
 
     # populate nix db, so nixos-install doesn't complain
     export NIX_STATE_DIR=$TMPDIR/state
-    nix-store --load-db < ${pkgs.closureInfo {
+    ${pkgs.fakeroot}/bin/fakeroot nix-store --register-validity --reregister < ${closureInfo {
```
Mic92, line 60: This is a bit slower but solves a real issue as it needs to recompute all checksums.
If the performance overhead is not acceptable, I can also hide this behind a conditional.

brianmcgillion:
Verified by dropping [this](https://github.com/tiiuae/ghaf/commit/b7465674daf81ac0c9de9a4e83011e3a7dfcfb85) patch against Ghaf. The patch was created because of an apparent mem leak in nix copy during image building.

With this change applied, the Image builds successfully again without the large memSize requirements. Thanks!

lib/make-disk-image.nix
Mic92: I decided to drop this approach again.

PR #632, opened 2024-05-13
swap: forward discardPolicy option to randomEncryption.allowDiscards

muggenhor:
I.e. instead of effectively ignoring `discardPolicy` when combined with `randomEncryption`, or adding a second option requiring the user to repeat themselves (DRY violation), just enable both when requested via `discardPolicy`.

Without this `discardPolicy` is a NOP when `randomEncryption` is enabled.

Follow-up: #623

### DIFF
--- a/lib/types/swap.nix
+++ b/lib/types/swap.nix
@@ -92,7 +92,12 @@
       default = [{
         swapDevices = [{
           device = config.device;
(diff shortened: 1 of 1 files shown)


Mic92:
@mergify queue

PR #634, opened 2024-05-14
disko-install: only set NIX_STATE_DIR for load-db

flokli:
We don't need to keep this env var around for the xargs invocation, only when running `nix-store --load-db`.


Mic92:
@mergify queue

PR #635, opened 2024-05-15
Add missing dependencies

jmbaur:
This allows for the disko scripts to be ran in more foreign environments where we can't assume the basic utilities coreutils provides are available or aren't 1-for-1 compatible (e.g. busybox utilities).

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -385,8 +385,9 @@ let
             internal = true;
             description = ''
               packages required by the disko configuration
+              coreutils is always included
             '';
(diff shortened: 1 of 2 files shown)


Lassulus:
what is the difference between coreutils and coreutils-full?

jmbaur:
Honestly not sure of the semantic difference between `coreutils-full` and `coreutils`, but nixos uses `coreutils-full` by default (https://github.com/NixOS/nixpkgs/blob/2057814051972fa1453ddfb0d98badbea9b83c06/nixos/modules/config/system-path.nix#L15), so I figured it would be safer to specify that here, as that's what variant of coreutils is most likely being implicitly used by default by disko

Issue #622, opened 2024-05-05
diskoImagesScript creates host's user home dir with some garbage leftovers


pedohorse:
after running `nix build .#nixosConfigurations.myhost.config.system.build.diskoImagesScript` the resulting image has additional home dir created for host's user that ran the command (the user itself is not created, and that dir is owned by root). 

That dir contains single `.nix-defexpr` dir with `channels`->`/tmp/state/profiles/per-user/root/channels` broken link.

I assume that's some leftovers from building process that should not be there.

Mic92:
Weird. No idea where these are coming from. We don't have this string in any of our code.

Mic92:
Ok. Can you make a pull request?

pedohorse:
oh, I can try, but i've just recently discovered this repo and don't know much about all features and commands disco provides, so can you point me towards where I should look to add that cleanup?

from a brief look it seems that cleanup should be added somewhere here, no?
https://github.com/nix-community/disko/blob/cb1d6fba694ab3887600d606106f5a044ba1712c/lib/make-disk-image.nix#L73-L74

Mic92:
We should probably clear out as much as possible environment variables. Does overriding home has any other side effects?

pedohorse:
I've just cleaned up everything, except origBuilder though preVM script

but it relies on env being passed through `xchg/saved-env` file, and I'm not sure if this is a spec, or implementation detail for `vmRunCommand`

here's an example: https://github.com/nix-community/disko/pull/639

It does seem to work fine, but i only tested a few configs for `config.system.build.diskoImagesScript`, nothing else

PR #639, opened 2024-05-17
cleanup host environment passed to VM for image building

pedohorse:
keeping only `origBuilder` env from host

should resolve https://github.com/nix-community/disko/issues/622

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -146,29 +146,30 @@ in
     export preVM=${diskoLib.writeCheckedBash { inherit pkgs checked; } "preVM.sh" ''
       set -efu
       mv copy_before_disko copy_after_disko xchg/
+      origBuilder=${pkgs.writeScript "disko-builder" ''
+        set -eu
+        export PATH=${lib.makeBinPath dependencies}
+        for src in /tmp/xchg/copy_before_disko/*; do
+          [ -e "$src" ] || continue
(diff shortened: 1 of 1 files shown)


Mic92:
Looks good. Does it work?

pedohorse:
i've mentioned in the issue https://github.com/nix-community/disko/issues/622 that it does seem to work fine, but i only manually tested a few configs for `config.system.build.diskoImagesScript`, and visually confirmed that result boots and seems good, nothing else.

the change _should_ affect only VM environment, created early in stage-2 script, the only change in environment in which the script that starts VM is executed is the absence of `origBuilder` var there, as it's set through `preVM` instead

i basically know two features in this project so far :)
are there any automated tests to ensure nothing else broke from this change?

Lassulus:
there are tests in the tests folder, they should be nix-build able or alternatively with nix build .#tests.make-disk-image

pedohorse:
i've ran some tests so far (I assume - the process of building IS running the test), it seem to work fine.
interestingly, the tests run with `nix build` will run in clean environment, so they won't catch issues as with running the result of image building script in tainted user enviornment. (unless there is such special test case that imitates that problem somewhere, but I doubt that since it seems that this was not noticed to be something noticeable before)

Oh wait, (i'm blind) I've just noticed that there are github actions that run all the tests, and all the tests pass

Mic92:
@mergify queue

PR #640, opened 2024-05-19
make-disk-image: make extraPostVM configurable from module.nix

MatthewCroughan:
My use case was compressing disk images, like this:

```nix
  disko.extraPostVM = ''
    ${pkgs.zstd}/bin/zstd --compress $out/*raw
    rm $out/*raw
  '';
```

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -17,6 +17,17 @@ in
       '';
       default = [ ];
     };
+    extraPostVM = lib.mkOption {
+      type = lib.types.str;
+      description = ''
+        extra shell code to execute once the disk image(s) have been succesfully created and moved to $out
+      '';
+      default = "";
+      example = pkgs.literalExpression ''
+        ''${pkgs.zstd}/bin/zstd --compress $out/*raw
+        rm $out/*raw
+      '';
+    };
     memSize = lib.mkOption {
       type = lib.types.int;
       description = ''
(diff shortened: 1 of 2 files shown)


module.nix
```diff
@@ -17,6 +17,13 @@ in
       '';
       default = [ ];
     };
+    extraPostVM = lib.mkOption {
+      type = lib.types.str;
+      description = ''
+        extra shell code to execute once the disk image(s) have been succesfully created and moved to $out
+      '';
+      default = "";
+    };
```
Lassulus, line 26: an example for the compression would be nice here? :)

MatthewCroughan: Done

module.nix
```diff
@@ -17,6 +17,17 @@ in
       '';
       default = [ ];
     };
+    extraPostVM = lib.mkOption {
+      type = lib.types.str;
+      description = ''
+        extra shell code to execute once the disk image(s) have been succesfully created and moved to $out
+      '';
+      default = "";
+      example = pkgs.literalExpression ''
```
dasJ, line 26: This doesn't evaluate - `literalExpression` is in `lib`, not in `pkgs`

MatthewCroughan: Thank you! Fixed in https://github.com/nix-community/disko/pull/646

PR #654, opened 2024-05-30
fix partprobe erroring sometimes with mdadm

Lassulus:


lib/types/mdadm.nix
```diff
@@ -43,11 +43,11 @@
             --force \
             --homehost=any \
             "''${disk_devices[@]}"
-          partprobe /dev/md/${config.name}
           udevadm trigger --subsystem-match=block
           udevadm settle
           # for some reason mdadm devices spawn with an existing partition table, so we need to wipe it
           sgdisk --zap-all /dev/md/${config.name}
+          partprobe /dev/md/${config.name}
```
Mic92, line 50: Makes sense to me. Afaik there udev rules that will assemble raids.

Lassulus: https://github.com/nix-community/disko/pull/358
https://github.com/nix-community/disko/pull/370

Mic92: If we zap the device, wouldn't partprobe then not always fail?

PR #655, opened 2024-05-30
add table-to-gpt migration guide

Lassulus:


Lassulus:
@makefu here for review

PR #685, opened 2024-06-24
document extraRootModules better

Lassulus:


MatthewCroughan (approved):
Precisely, thank you!

PR #686, opened 2024-06-25
lvm_vg: add lvm thinpool/thinlv support 

MatthewCash:
This is #285 rebased with some changes to priority.

- Device mapper kernel modules are loaded based on the LV types used. This applies to all types of LVs (mirror, raid, etc.) that have loadable modules.
- New types for the containing "thin-pool" and nested "thinlv".
- Thin pools have a lower priority than other LVs so that they are created first.
- Example and tests for a setup with thin pool and thin LVs

### DIFF
--- a/lib/types/lvm_vg.nix
+++ b/lib/types/lvm_vg.nix
@@ -1,4 +1,10 @@
 { config, options, lib, diskoLib, ... }:
+let
+  # Load kernel modules to ensure device mapper types are available
+  kernelModules = lib.filter (x: x != "") (map
+    (lv: lib.optionalString (lv.lvm_type != null && lv.lvm_type != "thinlv") "dm-${lv.lvm_type}")
+  (lib.attrValues config.lvs));
+in
 {
   options = {
     name = lib.mkOption {
@@ -21,7 +27,7 @@
           };
           priority = lib.mkOption {
             type = lib.types.int;
-            default = if (lib.hasInfix "100%" lv.config.size) then 9001 else 1000;
+            default = (if lv.config.lvm_type == "thin-pool" then 501 else 1000) + (if lib.hasInfix "100%" lv.config.size then 251 else 0);
             defaultText = lib.literalExpression ''
               if (lib.hasInfix "100%" lv.config.size) then 9001 else 1000
             '';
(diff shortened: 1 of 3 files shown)


lib/types/lvm_vg.nix
```diff
@@ -76,9 +88,12 @@
             if ! lvdisplay '${config.name}/${lv.name}'; then
               lvcreate \
                 --yes \
-                ${if lib.hasInfix "%" lv.size then "-l" else "-L"} ${lv.size} \
+                ${if (lv.lvm_type != "thinlv") then
+                  (if lib.hasInfix "%" lv.size then "-l" else "-L")
+                  else "-V"} ${lv.size} \
                 -n ${lv.name} \
-                ${lib.optionalString (lv.lvm_type != null) "--type=${lv.lvm_type}"} \
+                ${lib.optionalString (lv.lvm_type == "thinlv") "--thinpool=${lv.pool}"} \
```
Mic92, line 95: Can you add an example to examples for thinlv or extend an existing one so that we can have an integration test for it?

lib/types/lvm_vg.nix
```diff
@@ -76,9 +88,12 @@
             if ! lvdisplay '${config.name}/${lv.name}'; then
               lvcreate \
                 --yes \
-                ${if lib.hasInfix "%" lv.size then "-l" else "-L"} ${lv.size} \
+                ${if (lv.lvm_type != "thinlv") then
+                  (if lib.hasInfix "%" lv.size then "-l" else "-L")
+                  else "-V"} ${lv.size} \
```
Mic92, line 93:
```suggestion
                ${if (lv.lvm_type == "thinlv") then "-V" 
                  else if lib.hasInfix "%" lv.size then "-l" else "-L"} 
                  ${lv.size} \
```

lib/types/lvm_vg.nix
MatthewCash: I added a fairly simple example and test based on one of the other lvm ones.

lib/types/lvm_vg.nix
MatthewCash: Done

Mic92 (approved):
@Lassulus any comments from your side?

lib/types/lvm_vg.nix
Lassulus: does the thinpool need to exist before? maybe we need to do some ordering in that case? or is the thinpool always `size = "100%FREE` ?

MatthewCash: The thinpool does need to exist before any of its thin LVs which is taken care of by the priority option's default value.
https://github.com/nix-community/disko/blob/de29130286bbad9e17ea71f6bd99a272a1ec5c75/lib/types/lvm_vg.nix#L30

Mic92:
@mergify queue

PR #688, opened 2024-06-28
gpt: move boot partitions to the front of disko

Mic92:
Than we no longer need to do this in our examples.

### DIFF
--- a/lib/types/gpt.nix
+++ b/lib/types/gpt.nix
@@ -45,7 +45,19 @@ in
           };
           priority = lib.mkOption {
             type = lib.types.int;
-            default = if (partition.config.size or "" == "100%") then 9001 else 1000;
+            default = if partition.config.size or "" == "100%" then
+              9001
+            else if partition.config.type == "EF02" then
+              # Boot partition should be created first, because some BIOS implementations require it.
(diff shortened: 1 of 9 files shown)


lib/types/gpt.nix
```diff
@@ -45,7 +45,20 @@ in
           };
           priority = lib.mkOption {
             type = lib.types.int;
-            default = if (partition.config.size or "" == "100%") then 9001 else 1000;
+            default = if partition.config.size or "" == "100%" then
+              9001
+            else if partition.config.type == "EF02" then
+              # Boot partition should be created first,
+              # because some BIOS implementations require it.
+              # We leave some numbers before in case we ever hat to insert something before
```
polarathene, line 53: Typo:

```suggestion
              # We leave some numbers before in case we ever have to insert something before
```

Alternatively, this might be phrased a little better? (_"We" reasoning technically applies to users too, not just `disko`?_):

```suggestion
              # Priority defaults to 100 here to support any potential use-case for placing partitions prior to EF02
```

Mic92: Thanks!

Mic92:
@mergify queue

Mic92:
@mergifyio refresh

Issue #690, opened 2024-06-28
Incremental formatting will wipe a pool if the zpool pool is not imported


Mic92:
`zpool list` will return false if the pool exists but is not imported:

https://github.com/nix-community/disko/blob/8767dbf5d723b1b6834f4d09b217da7c31580d58/lib/types/zpool.nix#L65

PR #691, opened 2024-06-28
zfs: try to import zpool in incremental mode

Mic92:
if the pool is not imported, zfs list will return a value. So to check that the pool does not exist, we also import it.

fixes https://github.com/nix-community/disko/issues/690


Mic92:
@mergify refresh

PR #692, opened 2024-06-28
disko-install: don't copy channel

Mic92:
Since we use flakes, channels have no buisness.
This is consistent with nixos-anywhere.


Mic92:
@mergify queue

Issue #668, opened 2024-06-07
Use Disko with "nixos-rebuild build-vm"


arjan-s:
My system configs are all using Disko for partitioning and mounting, which is fantastic. However, sometimes I would like to test stuff without rebooting, so I would like to use `nixos-rebuild build-vm` for that. That doesn't seem to be supported currently, because booting the VM results in the boot process waiting indefinitely for the root partition to be found.

I've found the same question [at least here](https://discourse.nixos.org/t/use-disko-with-nixos-rebuild-build-vm-with-bootloader-or-something-similar/44895) and [possibly here](https://discourse.nixos.org/t/how-to-create-a-virtual-machine-with-a-filesystem-defined-by-nixos-virtualisation-module/43396), which sadly didn't receive any answers.

My question is how hard this would be to achieve, and additionally whether it can be added to the documentation somewhere.

brianmcgee:
I was thinking of something similar as I was playing around with the `vmVariant` config and the `config.system.build.vm` target to try and simulate my desktop machine, disk setup and all. @Lassulus pointed out there's the `config.system.build.installTest` which will spin up a VM and check your `disko` config can install correctly. 

It would be nice of that could be re-purposed to mix in the necessary config to `vmVariant` so you can fire up an interactive VM instead.

Mic92:
If you manually import this module in your nixos configuration: https://github.com/NixOS/nixpkgs/blob/717c708df2ed4488bff130f9c34c16de7b419f4a/nixos/modules/virtualisation/build-vm.nix#L9
you should be able to debug what's going on with the `config.fileSystems` in when loading your nixos configuration into a nix repl: https://nixos-and-flakes.thiscute.world/best-practices/debugging

Mic92:
Interestingly build-vm now works on my machine? Does someone has a broken configuration for me?

Enzime:
Closing this as it should be working as of #713

If you're running flakes, you can run:

```
nix run .#nixosConfigurations.<hostname>.config.system.build.vmWithDisko
```

There's also documentation available here: https://github.com/nix-community/disko/blob/master/docs/interactive-vm.md

PR #713, opened 2024-07-21
add interactive VMs

Lassulus:


lib/types/luks.nix
```diff
@@ -114,19 +114,23 @@ in
...
           ${lib.optionalString config.askPassword ''
-            set +x
             askPassword() {
-              echo "Enter password for ${config.device}: "
-              IFS= read -r -s password
-              echo "Enter password for ${config.device} again to be safe: "
-              IFS= read -r -s password_check
-              export password
-              [ "$password" = "$password_check" ]
+              if [ -z ''${IN_DISKO_TEST+x} ]; then
```
phaer, line 118: Why is this needed? To an interactive password code path as opposed to just key files?

I think it can be a bit smelly to let code check whether it's in a test here 😅

docs/interactive-vm.md
```diff
@@ -0,0 +1,17 @@
+# Running Interactive VMs with disko
+
+disko now exports it's own flavor of interactive VMs (similiar to config.system.build.vm).
```
Mic92, line 3: Since we normal build-vm stuff is broken, should we maybe mention this here?

docs/interactive-vm.md
```diff
@@ -0,0 +1,17 @@
...
+Simply import the disko module and build the vm runner with:
+```
+nix build -L '.#nixosConfigurations.mymachine.config.system.build.vmWithDisko'
+```
+
+afterwards you can run the interactive VM with:
+
+```
+result/bin/disko-vm
```
Mic92, line 12: Could potentially become part of the disko cli in another PR.

lib/interactive-vm.nix
```diff
@@ -0,0 +1,88 @@
...
+in
+{
+  pure = pkgs.writeDashBin "disko-vm" ''
+    set -efux
+    export tmp=$(mktemp -d)
+    trap 'rm -rf "$tmp"' EXIT
+    ${lib.concatMapStringsSep "\n" (disk: ''
+      ${pkgs.qemu}/bin/qemu-img create -f qcow2 \
+      -b ${diskoImages}/${disk.name}.raw \
+      -F raw "$tmp"/${disk.name}.qcow2
```
Mic92, line 83: How does this cope with different disk sizes? Also why are not using qcow here?

lib/types/luks.nix
Mic92: Can we not just override passwordFile in disko?

lib/make-disk-image.nix
```diff
@@ -15,10 +18,14 @@ let
...
   cleanedConfig = diskoLib.testLib.prepareDiskoConfig nixosConfig.config diskoLib.testLib.devices;
   systemToInstall = nixosConfig.extendModules {
-    modules = [{
-      disko.devices = lib.mkForce cleanedConfig.disko.devices;
-      boot.loader.grub.devices = lib.mkForce cleanedConfig.boot.loader.grub.devices;
-    }];
+    modules = [
+      extraConfig
+      {
+        disko.testMode = true;
```
Mic92, line 24: We don't need testMode if we can override the password file in luks.

lib/types/luks.nix
Lassulus: no, that's not possible since disko doesn't merge the devices correctly

lib/make-disk-image.nix
Lassulus: password cannot be overwritten, so testMode is the easiest way to fix that

lib/interactive-vm.nix
Lassulus: this creates a COW overlay image, with the backing one the raw image generated by the imagebuilder

lib/types/luks.nix
Mic92: Can we set a "default" with higher than normal priority through "options" instead?

lib/interactive-vm.nix
Mic92: It would be good if we also could have qcow on the disko images side later. Because than we can make these images sparse and not explode in size.

Lassulus: thats possible with the extraPostVM hook already, but yeah maybe it would be better to use that here?

lib/types/luks.nix
Lassulus: no, since we can't modify the the options after we evaluate the vm attribute, because the config is already evaluated

docs/interactive-vm.md
Lassulus: whats broken?

Mic92: See the details here: https://github.com/nix-community/disko/issues/668

PR #716, opened 2024-07-22
Add --check option to formatter in flake

beviu:
I noticed that master is not formatted properly. Would it be interesting to add a CI check for formatting? This PR adds an option to `nix fmt` to check formatting without modifying files which could be added to buildbot.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,7 +40,8 @@ recursive layouts.
 Disko doesn't require installation: it can be run directly from nix-community
 repository. The [Quickstart Guide](./docs/quickstart.md) documents how to run
 Disko in its simplest form when installing NixOS. Alternatively, you can also
-use the new [disko-install](./docs/disko-install.md) tool, which combines `disko` and `nixos-install` into one step.
+use the new [disko-install](./docs/disko-install.md) tool, which combines
+`disko` and `nixos-install` into one step.
 
 For information on other use cases, including upgrading from an older version of
 **disko**, using **disko** without NixOS and downloading the module, see the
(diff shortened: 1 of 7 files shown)


iFreilicht:
Very cool! Yes, checking the formatting in PRs automatically would be very nice. Could you rebase the PR on current master?

Enzime:
I think ideally we would replace the  makeshift formatter script with one generated by a framework:

https://github.com/nix-community/nixos-anywhere/blob/34e1c624ec04d4aaddc22b07ff9de6186bd6ad0c/treefmt/flake-module.nix#L1-L17

or

https://github.com/Enzime/dotfiles-nix/blob/%F0%9F%94%A5/flake.nix#L359-L385

Mic92:
Since this repository needs to be imported into many nixos configurations, we tried so far the keep the flake inputs at a minimum.
However there is also a pattern to hide flake inputs that we only need for `nix flake check`: https://github.com/numtide/nixos-facter-modules/blob/6eb13404bbe40b9a5ef9984792b27933658571fa/flake.nix#L7

iFreilicht:
> I think ideally we would replace the makeshift formatter script with one generated by a framework:

Sure, that's nicer. But this PR would be good enough in my book to get the ball rolling on this. A working solution now is better than a perfect one later.

> Since this repository needs to be imported into many nixos configurations, we tried so far the keep the flake inputs at a minimum.

That's a good point to consider. I didn't know about this pattern, that would be very useful to incorporate Enzime's suggestion without any downsides.

iFreilicht (changes requested):
Forgot to formally make this a review; please rebase, apart from that the changes look good to me.

iFreilicht:
Alright I rebased and actually ran the formatter now. This already illustrates why we need PR check for this. I would be interested in adding that as well, but I'm not sure where to start. I'm not even sure how buildbot is triggered.

Are you alright with me merging this? Don't want to make that decision on my own.

iFreilicht (approved):

PR #717, opened 2024-07-22
Add installMountpoint option to filesystem

beviu:
Add an option to specify a different mountpoint for a filesystem during installation.

This can be useful when using the impermanence NixOS module with /nix on the persistent filesystem. During installation, we want to mount the filesystem to / and write to /nix under it, whereas in the NixOS configuration we want to mount the filesystem to /persistent and create a bind mount from /nix to /persistent/nix.

### DIFF
--- a/lib/types/filesystem.nix
+++ b/lib/types/filesystem.nix
@@ -24,7 +24,19 @@
     mountpoint = lib.mkOption {
       type = lib.types.nullOr diskoLib.optionTypes.absolute-pathname;
       default = null;
-      description = "Path to mount the filesystem to";
+      description = ''
+        Path to mount the filesystem to during normal operation, or null to
+        disable. The NixOS configuration module will add an entry under
+        `fileSystems` to mount the filesystem at this path.
+      '';
+    };
+    installMountpoint = lib.mkOption {
+      type = lib.types.nullOr diskoLib.optionTypes.absolute-pathname;
+      default = config.mountpoint;
+      description = ''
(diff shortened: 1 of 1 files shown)


lib/types/filesystem.nix
```diff
@@ -24,7 +24,19 @@
...
       type = lib.types.nullOr diskoLib.optionTypes.absolute-pathname;
       default = null;
-      description = "Path to mount the filesystem to";
+      description = ''
+        Path to mount the filesystem to during normal operation, or null to
+        disable. The NixOS configuration module will add an entry under
+        `fileSystems` to mount the filesystem at this path.
+      '';
+    };
+    installMountpoint = lib.mkOption {
```
Mic92, line 33: Would be nice to also have an impermanence example using this. It's also often what people want to see anyway.

Mic92:
What makes this a bit complex is that disko doesn't really distinguish between a normal mount and installation. I.e. if someone after installation wants to use the `disko --mode mount` command to just mount the system i.e. for repairing or chrooting into it.

beviu:
> What makes this a bit complex is that disko doesn't really distinguish between a normal mount and installation. I.e. if someone after installation wants to use the `disko --mode mount` command to just mount the system i.e. for repairing or chrooting into it

Oh no, I didn't see the mount command. I'm not sure what would be a good solution then.

beviu:
I am going to close this PR because there is a more elegant solution in #718. I can reopen it if needed.

PR #723, opened 2024-08-01
zfs: add ZFS "topology" features like explicit vdevs, cache, and special

### LINKED ISSUE #298: ZFS features support
As suggested in https://github.com/nix-community/disko/issues/267#issuecomment-1636789807, here's an issue to summarize ZFS features that disko does not currently support:

- [x] #267 complex vdex configurations, such as:
  - separate devices for L2ARC (2nd-layer cache) or ZIL (intent log) ;
  - multiple vdevs, each comprising of multiple drives (mirrored, in draid or raidz, etc.)
  such configurations can be nested recursively, see [`zpoolconcepts(7)`](https://openzfs.github.io/openzfs-docs/man/7/zpoolconcepts.7.html)

- [ ] “delegations,” which allow otherwise-unprivileged users to create children of a given dataset, snapshot it, set properties, etc. see [`zfs-allow(8)`](https://openzfs.github.io/openzfs-docs/man/8/zfs-allow.8.html)

dmadisetti:
Addresses vdev definitions, partially addressing #298 

I don't think this is the most elegant solution, so happy to churn over this a few times. Happy to receive feedback.
**TODOs** as I see it:

 - [x] More comprehensive tests, currently just a stub
 - [x] Clean up. I'm not exceptionally happy with this and would love some feedback, but I tried to stick the spec in #298

I think it might be worth having a `lib/types/zfs` folder. The vdev submodule could then be moved to be a standalone file without polluting the diskoTypes. Might be worth doing in another PR.

stakeholders: @victorklos @nbraud @luishfonseca @Mic92 @Lassulus


lib/types/zpool.nix
```diff
@@ -50,6 +53,55 @@
...
+                # zfs disk types
+                vdev = lib.mkOption {
+                  type = lib.types.listOf vdev;
+                  default = [ ];
+                  description = "A list of storage vdevs";
+                };
+                special = lib.mkOption {
+                  type = lib.types.nullOr vdev;
+                  default = null;
+                  description = "A list of devices for the special vdev";
```
Mic92, line 93: Can we link to some upstream documentation what special means? It's new to me.

lib/types/zpool.nix
```diff
@@ -50,6 +53,55 @@
...
+                  type = lib.types.enum [ "zfs_topology" ];
+                  default = "zfs_topology";
+                  internal = true;
+                  description = "Type";
+                };
+                # zfs disk types
+                vdev = lib.mkOption {
+                  type = lib.types.listOf vdev;
+                  default = [ ];
+                  description = "A list of storage vdevs";
```
Mic92, line 88: some example would be nice here:

```suggestion
                  description = "A list of storage vdevs";
                  example = [{
                   mode = "mirror";
                   members = [ "x" "y" ];
                  }];
```

lib/types/zpool.nix
```diff
@@ -60,42 +112,93 @@
...
+              elif (blkid "$dev" -o export | grep '^TYPE='); then
+                echo "device $dev already has a partition, skipping creating zpool ${config.name}" >&2
+                continue=0
+              fi
+            done
+            if [ $continue -eq 1 ]; then
+              topology=""
+              # For shell check
+              mode="${config.mode}"
+              if [ "$mode" != "prescribed" ]; then
```
Mic92, line 153: I think this should be rather a nix eval error than a runtime error.

lib/types/zpool.nix
```diff
@@ -13,13 +22,7 @@
...
     };
     mode = lib.mkOption {
-      type = lib.types.enum [
-        ""
-        "mirror"
-        "raidz"
-        "raidz2"
-        "raidz3"
-      ];
+      type = lib.types.enum (modeOptions ++ [ "prescribed" ]);
```
Mic92, line 25: If a user provides a `topology`, can we than not default the value to `prescribed`?
I also wonder if we should just call it `topology` here as well for consistency.

example/zfs-with-vdevs.nix
```diff
@@ -0,0 +1,113 @@
...
+        type = "zpool";
+        mode = "prescribed";
+        rootFsOptions = {
+          compression = "zstd";
+          "com.sun:auto-snapshot" = "false";
+        };
+        mountpoint = "/";
+
+        topology = {
+          type = "zfs_topology";
```
Mic92, line 89: Do we need to support any form of nesting here? Maybe we don't need a zfs_topology type here.

lib/types/zpool.nix
Mic92: Different idea. Could we accept the topology as an argument here? I don't know if this works with type checking though.

lib/types/zpool.nix
```diff
@@ -60,42 +112,93 @@
...
-              continue=0
-            elif (blkid "$dev" -o export | grep '^TYPE=zfs_member'); then
-              # zfs_member is a zfs partition, so we try to add the device to the pool
-              :
-            elif (blkid "$dev" -o export | grep '^TYPE='); then
-              echo "device $dev already has a partition, skipping creating zpool ${config.name}" >&2
-              continue=0
+      default =
+        let
+          format_output = (mode: members: ''
```
Mic92, line 117:
```suggestion
          formatOutput = (mode: members: ''
```

lib/types/zpool.nix
```diff
@@ -60,42 +112,93 @@
...
+            entries+=("${mode}=${
+              lib.concatMapStringsSep " "
+              (d: "/dev/disk/by-partlabel/disk-${d}-zfs") members
+            }")
+          '');
+          format_vdev = (vdev: format_output vdev.mode vdev.members);
```
Mic92, line 123:
```suggestion
          formatVdev = (vdev: format_output vdev.mode vdev.members);
```

lib/types/zpool.nix
```diff
@@ -60,42 +112,93 @@
...
+                  ''
+                }
+              else
+                entries=()
+                ${lib.optionalString (hasTopology && config.topology.vdev != null)
+                    (lib.concatMapStrings format_vdev config.topology.vdev)}
+                ${lib.optionalString (hasTopology && config.topology.special != null)
+                    (format_output "special ${config.topology.special.mode}" config.topology.special.members)}
+                ${lib.optionalString (hasTopology && config.topology.cache != null)
+                    (format_output "cache" [config.topology.cache])}
```
Mic92, line 169:
```suggestion
                ${lib.optionalString (hasTopology && config.topology.vdev != null)
                    (lib.concatMapStrings formatVdev config.topology.vdev)}
                ${lib.optionalString (hasTopology && config.topology.special != null)
                    (formatOutput "special ${config.topology.special.mode}" config.topology.special.members)}
                ${lib.optionalString (hasTopology && config.topology.cache != null)
                    (formatOutput "cache" [config.topology.cache])}
```

example/zfs-with-vdevs.nix
Mic92: Maybe we do need that because of the way serialization works?

victorklos:
How very nice that you are working on this! Glad to confirm your code is working and allowed me to use disko on a previously unsupported ZFS layout.

A couple of things remain, of which one is critical: the members of a vdev can either be whole disks or partitions.

Are you interested in feedback on a broader usability/logical level, or is your goal to produce something that is usable? Just asking before I get carried away ;)

dmadisetti:
Thanks @victorklos ! This was more so that I could utilize Disko myself in a similar "hack it together" context.

I would like a potential refactor and before upstreaming though.

I would love some feedback, maybe in the context of the comments that @ Mic92  has already left. I think they are good suggestions, I just haven't gotten around to it

0x450x6c:
LUKS support would be nice.

lib/types/zpool.nix
dmadisetti: Expanded description with relevant headings in https://openzfs.github.io/openzfs-docs/man/master/7/zpoolconcepts.7.html

example/zfs-with-vdevs.nix
dmadisetti: Type must be set, which is a little redundant, but fine. I did move this as per the suggestion

dmadisetti (commented):
Addressed comments, and pulled in commit from @0x450x6c

I can clean up the commit history, but seems like everything gets merged anyway ¯\_(ツ)_/¯

lib/types/zpool.nix
```diff
@@ -60,42 +133,88 @@
...
+        let
+          formatOutput = (mode: members: ''
+            entries+=("${mode}=${
+              lib.concatMapStringsSep " "
+              (d: if lib.strings.hasPrefix "/" d then d else "/dev/disk/by-partlabel/disk-${d}-zfs") members
+            }")
+          '');
+          formatVdev = (vdev: formatOutput vdev.mode vdev.members);
+          hasTopology = !(builtins.isString config.mode);
+          mode = if !hasTopology then config.mode else "prescribed";
```
Mic92, line 146:
```suggestion
          mode = if hasTopology then "prescribed" else config.mode;
```

lib/types/zpool.nix
```diff
@@ -60,42 +133,88 @@
...
+          topology = if hasTopology then config.mode.topology else { };
```
Mic92, line 147:
```suggestion
          topology = lib.optionalAttrs hasTopology config.mode.topology;
```

lib/types/zpool.nix
```diff
@@ -60,42 +133,88 @@
...
-              continue=0
+      default =
+        let
+          formatOutput = (mode: members: ''
+            entries+=("${mode}=${
+              lib.concatMapStringsSep " "
+              (d: if lib.strings.hasPrefix "/" d then d else "/dev/disk/by-partlabel/disk-${d}-zfs") members
+            }")
+          '');
+          formatVdev = (vdev: formatOutput vdev.mode vdev.members);
```
Mic92, line 144:
```suggestion
          formatVdev = vdev: formatOutput vdev.mode vdev.members;
```

lib/types/zpool.nix
```diff
@@ -60,42 +133,88 @@
...
-              echo "device $dev already has a partition, skipping creating zpool ${config.name}" >&2
-              continue=0
+      default =
+        let
+          formatOutput = (mode: members: ''
+            entries+=("${mode}=${
+              lib.concatMapStringsSep " "
+              (d: if lib.strings.hasPrefix "/" d then d else "/dev/disk/by-partlabel/disk-${d}-zfs") members
+            }")
+          '');
```
Mic92, line 143:
```suggestion
          formatOutput = mode: members: ''
            entries+=("${mode}=${
              lib.concatMapStringsSep " "
              (d: if lib.strings.hasPrefix "/" d then d else "/dev/disk/by-partlabel/disk-${d}-zfs") members
            }")
          '';
```

Mic92 (approved):
Looks overall pretty good.

Issue #678, opened 2024-06-17
NixOS modules with disko config don't merge


jmbaur:
I would expect setting disko config in two different NixOS modules would properly merge together. A minimal reproducer is the config below, where I would expect evaluating it to fail with conflicting definitions for `mountpoint`:

```nix
let
  nixpkgs = builtins.getFlake "github:nixos/nixpkgs/nixos-unstable";
  disko = builtins.getFlake "github:nix-community/disko/master";
in
(nixpkgs.lib.nixosSystem {
  modules = [
    disko.nixosModules.default
    { nixpkgs.hostPlatform = "x86_64-linux"; }
    {
      disko.devices.disk.disk0 = {
        device = "/dev/foobar";
        type = "disk";
        content = {
          type = "gpt";
          partitions.foo = {
            type = "EF00";
            content = {
              type = "filesystem";
              format = "vfat";
              mountpoint = "/boot";
            };
          };
        };
      };
    }
    {
      disko.devices.disk.disk0 = {
        type = "disk";
        content = {
          type = "gpt";
          partitions.foo = {
            type = "EF00";
            content = {
              type = "filesystem";
              mountpoint = null;
            };
          };
        };
      };
    }
  ];
}).config.disko.devices.disk.disk0.content.partitions.foo.content.mountpoint
```

Lassulus:
but you have conflicting definitions? not sure whats the error here? what should happen otherwise?

jmbaur:
That's the thing, I would expect that evaluating the above expression would error out due to the fact that there are conflicting definitions, but it does not.

Result from evaluation (where test.nix contains the contents as posted above):
```
~ > nix eval -f test.nix
"/boot"
```

jmbaur:
Seems related: https://github.com/NixOS/nixpkgs/pull/254790#discussion_r1358073458

Lassulus:
Ah this should be fixed now

Enzime:
Fixed by #1094

PR #724, opened 2024-08-02
refac(examples): Denest the source

goyalyashpal:
Hi!

I was just trying to understand the `disko`'s examples, and i did this refactoring.

i found this interesting and thought to share with yous and ask if you'd be interested to get a merge request  refactoring others in a similar fashion too :smiley:?

image

----
[nixel reference]: https://nix.dev/tutorials/nix-language.html#evaluating-nix-files "Evaluating Nix files"

i tried to run the `nix-instantiate --eval --strict` ([nixel reference]) and yes, these evaluated to absolutely same thing :smiley:  

```console
$ nix-instantiate --eval --strict legacy-table.nix
{ disko = { devices = { disk = { main = { content = { format = "gpt"; partitions = [ { bootable = true; content = { format = "vfat"; mountpoint = "/boot"; type = "filesystem"; }; end = "500M"; name = "ESP"; start = "1M"; } { bootable = true; content = { format = "ext4"; mountpoint = "/"; type = "filesystem"; }; end = "100%"; name = "root"; part-type = "primary"; start = "500M"; } ]; type = "table"; }; device = "/dev/sda"; type = "disk"; }; }; }; }; }

$ nix-instantiate --eval --strict legacy-table-refactored.nix 
{ disko = { devices = { disk = { main = { content = { format = "gpt"; partitions = [ { bootable = true; content = { format = "vfat"; mountpoint = "/boot"; type = "filesystem"; }; end = "500M"; name = "ESP"; start = "1M"; } { bootable = true; content = { format = "ext4"; mountpoint = "/"; type = "filesystem"; }; end = "100%"; name = "root"; part-type = "primary"; start = "500M"; } ]; type = "table"; }; device = "/dev/sda"; type = "disk"; }; }; }; }; }

```

----
a more elaborate setup to compare automated-ly, rather than eyeballing :eyes: :

```bash
eval-diff() {
  # Show diff of strict-eval of 2 nix files

  # Get the parameters
  local oldfile=$1
  local newfile=$2

  # Function: Abstract out the flags in invocation
  sevalnix() { nix-instantiate --eval --strict $1; }

  # Strict-evaluate both the files
  local oldval="$(sevalnix ${oldfile})"
  local newval="$(sevalnix ${newfile})"

  # Use process substitution to make stdout act as files to `diff` utility
  diff -u <(echo "$oldval") <(echo "$newval") | 

  # Pipe to `bat` with syntax-highlighting for diff
  bat -l diff;
}
```

```console
$ eval-diff legacy-table{,-refactored}.nix 
───────┬──────────────────────────────────────────────────────────────
       │ STDIN   <EMPTY>
───────┴──────────────────────────────────────────────────────────────


```

### DIFF
--- a/example/legacy-table.nix
+++ b/example/legacy-table.nix
@@ -1,40 +1,38 @@
 {
-  disko.devices = {
-    disk = {
-      main = {
-        device = "/dev/sda";
-        type = "disk";
-        content = {
-          type = "table";
-          format = "gpt";
-          partitions = [
-            {
-              name = "ESP";
-              start = "1M";
-              end = "500M";
-              bootable = true;
-              content = {
-                type = "filesystem";
-                format = "vfat";
-                mountpoint = "/boot";
-              };
-            }
-            {
-              name = "root";
-              start = "500M";
-              end = "100%";
-              part-type = "primary";
-              bootable = true;
-              content = {
-                type = "filesystem";
-                format = "ext4";
-                mountpoint = "/";
-              };
-            }
-          ];
-        };
-      };
-    };
+  disko.devices.disk.main = {
+    device = "/dev/sda";
+    type = "disk";
   };
+
+  disko.devices.disk.main.content = {
+    type = "table";
+    format = "gpt";
+  };
+
+  disko.devices.disk.main.content.partitions = [
+    {
+      name = "ESP";
+      start = "1M";
+      end = "500M";
+      bootable = true;
+      content = {
+        type = "filesystem";
+        format = "vfat";
+        mountpoint = "/boot";
+      };
+    }
+    {
+      name = "root";
+      start = "500M";
+      end = "100%";
+      part-type = "primary";
+      bootable = true;
+      content = {
+        type = "filesystem";
+        format = "ext4";
+        mountpoint = "/";
+      };
+    }
+  ];
 }
 


Mic92 (approved):
Looks easier to read to me.

Lassulus:
not sure, this touches the legacy partitions which we should migrate from anyway. Also by splitting it up people could be inclined to split the code over multiple files and this will trigger weird errors.

Mic92:
> not sure, this touches the legacy partitions which we should migrate from anyway. 

I think this was just meant as an example to show what this would look like for all examples.
Also as long as we still have uboot firmware that requires fat partitioning, we cannot get rid of it until there is a proper replacement for MBR.

> Also by splitting it up people could be inclined to split the code over multiple files and this will trigger weird errors.

What parts don't merge correctly? There are ways to allow only one instance to be defined.
I think this is not a very strong argument to make things less readable for everyone.

Lassulus:
> What parts don't merge correctly? There are ways to allow only one instance to be defined. I think this is not a very strong argument to make things less readable for everyone.

https://github.com/nix-community/disko/issues/678

goyalyashpal:
> Also by splitting it up people could be inclined to split the code over multiple files and this will trigger weird errors.
> https://github.com/nix-community/disko/issues/678
> \- @ Lassulus at https://github.com/nix-community/disko/pull/724#issuecomment-2283195200

how about some note like any of following or smth similar; maybe in the example files themselves?

* "Do not specify same key twice" 
* "Unexpected behaviour if same key specified more than once"
* ...


'cz in the current form, the examples are unreadable and don't really fulfill their purpose to the full potent.

goyalyashpal:
hi all.  so, did you reach on any conclusion?

should this be kept open, discussed, closed or merged?

goyalyashpal:
maybe there's some way in nix-language to restrict (say error out and early exit) in such cases?

PR #727, opened 2024-08-07
Lib: vmtools use default linux pkgs.

qbisi:
When building diskoImage in Qemu with a custom kernel (nixosConfig.config.boot.kernelPackages) which does not have kernel modules in ["9p" "9pnet_virtio" "virtio_pci" "virtio_blk"]. 
The build process will raise a module not found error.
Just use nixpkgs default kernel instead.

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -10,7 +10,7 @@ let
   vmTools = pkgs.vmTools.override {
     rootModules = [ "9p" "9pnet_virtio" "virtio_pci" "virtio_blk" ] ++ nixosConfig.config.disko.extraRootModules;
     kernel = pkgs.aggregateModules
-      (with nixosConfig.config.boot.kernelPackages; [ kernel ]
+      (with pkgs.linuxPackages; [ kernel ]
         ++ lib.optional (lib.elem "zfs" nixosConfig.config.disko.extraRootModules) zfs);
   };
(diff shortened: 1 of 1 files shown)


lib/make-disk-image.nix
```diff
@@ -10,7 +10,7 @@ let
   vmTools = pkgs.vmTools.override {
     rootModules = [ "9p" "9pnet_virtio" "virtio_pci" "virtio_blk" ] ++ nixosConfig.config.disko.extraRootModules;
     kernel = pkgs.aggregateModules
-      (with nixosConfig.config.boot.kernelPackages; [ kernel ]
+      (with pkgs.linuxPackages; [ kernel ]
```
Mic92, line 13: Can we make this configurable instead but default to the configured kernel?
I see the usecase of people building smaller kernel, but this is the exception
rather than the norm and people doing that will know how to fix it.
There are advantages if filesystem are created with the same kernel that there are used with afterwards.

Mic92: I would prefer some kernel options as shown here: https://github.com/nix-community/disko/pull/698/files#diff-d08d64ddbf5f8a384c8213d871167c6f55002970d7312c302b4bd441fafb5e8dR32

Lassulus:
this should be now configurable through https://github.com/nix-community/disko/pull/698/files#diff-d08d64ddbf5f8a384c8213d871167c6f55002970d7312c302b4bd441fafb5e8dR32

PR #740, opened 2024-08-20
f2fs: init

MatthewCroughan:
Just allows f2fs to be used.

### DIFF
--- a/lib/types/filesystem.nix
+++ b/lib/types/filesystem.nix
@@ -90,6 +90,7 @@
           else if (config.format == "ext3") then [ pkgs.e2fsprogs ]
           else if (config.format == "ext4") then [ pkgs.e2fsprogs ]
           else if (config.format == "bcachefs") then [ pkgs.bcachefs-tools ]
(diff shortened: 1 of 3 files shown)


lib/types/filesystem.nix
```diff
@@ -90,6 +90,7 @@
           else if (config.format == "ext3") then [ pkgs.e2fsprogs ]
           else if (config.format == "ext4") then [ pkgs.e2fsprogs ]
           else if (config.format == "bcachefs") then [ pkgs.bcachefs-tools ]
+          else if (config.format == "f2fs-tools") then [ pkgs.f2fs-tools ]
```
Lassulus, line 93: is this correct? or should the first be f2fs?

MatthewCroughan: Should be `f2fs`, but I think due to https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/tasks/filesystems/f2fs.nix#L9-L20, it doesn't fail the VM Test, and maybe the others also wouldn't if typos were made there too.

PR #746, opened 2024-08-22
zfs_fs: Fix errors when not changing mountpoint

iFreilicht:
Running `zfs set mountpoint=/mnt/my-ds tank/my-ds`, ZFS may try to unmount the dataset even if the mountpoint didn't change. To avoid the confusing error message, this command is now only run when the mountpoint actually changes.

@Mic92 I'm tagging you as you're the author of #689


Mic92:
@mergify queue

PR #752, opened 2024-08-30
swap: mountOptions; btrfs priority, options

nvmd:
Add missing configuration options for regular (types/swap) and btrfs (types/btrfs) swap.

-  `types/swap`: `mountOptions`
   - will allow specifying such options as `nofail` which is currently not possible
   - named to be consistent with `mountOptions` for other fs types
   - relevant mount script changes

- `types/btrfs`:
  - `priority` and `options`
  - naming consistent with underlying nixos options


Mic92:
@mergify queue

PR #757, opened 2024-09-03
disk-deactivate: only run zfs command if they are present

Mic92:


disk-deactivate/disk-deactivate.jq
```diff
@@ -1,7 +1,7 @@
 # since lsblk lacks zfs support, we have to do it this way
 def remove:
   if .fstype == "zfs_member" then
-    "zpool destroy -f \(.label)"
+    "if type zpool >/dev/null; then zpool destroy -f \(.label); zpool labelclear -f \(.label); fi"
```
Mic92, line 4: So it looks like `zpool destroy` only works if the pool is mounted. `zpool labelclear` works in more cases.

PR #760, opened 2024-09-08
apply extraSystemConfig to diskoScript also

Qubasa:
If an option in the originalSystemConfig is null and is supposed to be set by the extraSystemConfig then eval fails at diskoScript because extraSystemConfig wasn't applied there. This PR fixes this issue.

### DIFF
--- a/install-cli.nix
+++ b/install-cli.nix
@@ -7,7 +7,17 @@
 ,
 }:
 let
-  originalSystem = (builtins.getFlake "${flake}").nixosConfigurations."${flakeAttr}";
+  extraConfig = { _file = "disko-install --system-config"; } // (builtins.fromJSON extraSystemConfig);
+
+  originalSystem = (builtins.getFlake "${flake}").nixosConfigurations."${flakeAttr}".extendModules {
+    modules = [
+      {
+        imports = [ extraConfig ];
+      }
+    ];
+  };
+
+
   lib = originalSystem.pkgs.lib;
 
   deviceName =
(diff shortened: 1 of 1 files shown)


Lassulus:
I don't understand, your original system should be a valid config for it to be a valid flake?

Qubasa:
@Lassulus No the possibility exists that your original system is not a valid config and becomes valid only after the extraSystemConfig. 
A more concrete example for clan:
You disable fact generation in clanCore because you want an immutable system config. Now disk_id / machine_id is null, and the systemConfig should override it to a random uuid generated from the caller.

Qubasa:
Even if you make the default some string, if you try to override the disk name inside extraSystemConfig it doesn't get applied to the diskoScript which does the disk formatting so overriding disko options over the extraSystemConfig is currently not possible

Lassulus:
but the flake won't eval then without the extraSystemConfig, I think the better fix would be to make the original system valid instead of relying on extraSystemConfig to make your system a valid one.

install-cli.nix
```diff
@@ -7,7 +7,17 @@
...
 }:
 let
-  originalSystem = (builtins.getFlake "${flake}").nixosConfigurations."${flakeAttr}";
+  extraConfig = { _file = "disko-install --system-config"; } // (builtins.fromJSON extraSystemConfig);
+
+  originalSystem = (builtins.getFlake "${flake}").nixosConfigurations."${flakeAttr}".extendModules {
+    modules = [
+      {
+        imports = [ extraConfig ];
+      }
```
Mic92, line 16:
```suggestion
      extraConfig
```

PR #762, opened 2024-09-09
zfs_fs: support updating mountpoint properly

Enzime:
Use `zfs set -u` to update the `mountpoint` flag without mounting or unmounting the filesystem. This flag was added in OpenZFS 2.2.0, which was released October 2023.

The previous logic would not update `mountpoint` if `config.options` was empty or only contained `mountpoint`.

### DIFF
--- a/lib/types/zfs_fs.nix
+++ b/lib/types/zfs_fs.nix
@@ -73,8 +73,7 @@
               "pbkdf2salt"
               "keyformat"
             ];
-            updateOptions = builtins.removeAttrs config.options (onetimeProperties ++ [ "mountpoint" ]);
-            mountpoint = config.options.mountpoint or config.mountpoint;
+            updateOptions = builtins.removeAttrs createOptions onetimeProperties;
           in
           ''
             if ! zfs get type ${config._name} >/dev/null 2>&1; then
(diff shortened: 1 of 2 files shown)


Enzime:
@iFreilicht could you test if this change works on your system?

@Mic92 @Lassulus now that `_create` is idempotent, is it meant to be run online (with your disks already mounted) or offline?

Mic92:
> @iFreilicht could you test if this change works on your system?
> 
> @Mic92 @Lassulus now that `_create` is idempotent, is it meant to be run online (with your disks already mounted) or offline?

It's meant to run from a installer or in-memory system that doesn't depend on the disks. However ideally it could also unmount disks as needed to perform the operation.

iFreilicht:
> @iFreilicht could you test if this change works on your system?

Don't have access right now, will do next week. Feel free to ping me again if I didn't reply by next Wednesday.

Enzime:
@iFreilicht out of curiosity, were you running the format script on an online machine or an offline one?

iFreilicht:
I ran on an online system if by online you mean "already installed and with drives mounted".

Enzime:
@Mic92 any idea why `make-disk-image` and `make-disk-image-impure` are failing in BuildBot? They succeed locally and there shouldn't be any changes in this PR compared to `master`

Mic92:
@Enzime they seem to succeed now, but `.#checks.x86_64-linux.zfs-with-vdevs`

Enzime:
@Mic92 yeah that check succeeds locally for me

Enzime:
@Mic92 is there a way to trigger it to run again? I'm wondering if it's a flaky test

Lassulus:
I can login into buildbot with github on the overview and then retrigger it, but maybe that's one of my undocumented superpowers I forgot about

Mic92:
We just need to invite you to the nix-community org and then you can trigger rebuilds

Mic92:
Invite sent.

Enzime:
@mergify queue

Issue #769, opened 2024-09-13
xcp exhausts memory, probably avoidably


colemickens:
1. I have to bump memsize to build a disk image with a moderate (but not huge) closure.
2. I am surprised I don't hit cmdline arg length limits.
3. I think that using something other than `xcp` with ALL store paths at once, would avoid needing a larger guest just to install the closure.

Mic92:
disko-images uses cp not xcp.

Mic92:
Did you maybe use an older version of disko? Because we are using xargs + cp: https://github.com/nix-community/disko/blob/624fd86460e482017ed9c3c3c55a3758c06a4e7f/lib/make-disk-image.nix#L91

colemickens:
I think I just crossed wires. I am definitely using recent disko. However, I'm still suspicious that we can somehow batch/chunk around `xargs cp`.

It's very apparent that this copy step uses a LOT of memory, it's where my image build fails every time with a large enough closure and not enough VM ram.

Obviously it's easy to just bump the builder ram, but again, I'm guessing that this can be massaged to not use so much RAM for the copy.

iFreilicht:
Hmm, I'm not sure why this happens. I can definitely confirm that:

1. running `nom build .#checks.x86_64-linux.make-disk-image` causes the qemu process to use up to 2.1GB of RAM
2. before the line `nixos-disko-images> ++ xargs cp --recursive --target /mnt/nix/store`, the RAM usage was less than 1GB

However, I also observed:

3. After `xargs cp` is done, the RAM usage of qemu does not decrease

Additionally, I tried to reproduce this by running this copy operation on the exact same `closure-info` locally:

```
$ cat ./xargs-cp.sh
#!/usr/bin/env bash
xargs cp --recursive --target /mnt/scratch/test-store < "$1/store-paths"
$ nix run nixpkgs#time -- -v ./xargs-cp.sh /nix/store/a3s32wbdg5yain492c3gq8fbv9aak6vd-closure-info
        Command being timed: "./xargs-cp.sh /nix/store/a3s32wbdg5yain492c3gq8fbv9aak6vd-closure-info"
        User time (seconds): 0.25
        System time (seconds): 2.14
        Percent of CPU this job got: 56%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 0:04.25
        Average shared text size (kbytes): 0
        Average unshared data size (kbytes): 0
        Average stack size (kbytes): 0
        Average total size (kbytes): 0
[...]

PR #770, opened 2024-09-13
make-disk-image: honor disko.memSize as default

colemickens:
related to https://github.com/nix-community/disko/issues/769

I had to bump `memSize` to work around #769. I couldn't figure out why it needed _so much RAM_ until I realized that the module does not influence the impure script.

I think this is "ok", to plumb through the default.

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -200,7 +200,7 @@ in
     ''}
     export postVM=${diskoLib.writeCheckedBash { inherit pkgs checked; } "postVM.sh" postVM}
 
-    build_memory=''${build_memory:-1024}
+    build_memory=''${build_memory:-${nixosConfig.config.disko.memSize}}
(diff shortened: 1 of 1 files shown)


lib/make-disk-image.nix
```diff
@@ -200,7 +200,7 @@ in
     ''}
     export postVM=${diskoLib.writeCheckedBash { inherit pkgs checked; } "postVM.sh" postVM}
 
-    build_memory=''${build_memory:-1024}
+    build_memory=''${build_memory:-${nixosConfig.config.disko.memSize}}
```
Lassulus, line 203:
```suggestion
    build_memory=''${build_memory:-${toString nixosConfig.config.disko.memSize}}
```

Enzime (changes requested):
Can you update the `--help` output as well?

Enzime:
Superseded by #801

PR #771, opened 2024-09-14
make-disk-image: convert into NixOS module

Enzime:
As `makeDiskImages` always requires a NixOS configuration, we can simplify the code by converting it into a NixOS module. Then we can make it responsible for populating `system.build.diskoImages` and `system.build.diskoImagesScript`.

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -1,72 +1,108 @@
-{ config, lib, pkgs, extendModules, ... }@args:
+{ config, lib, pkgs, extendModules, diskoLib, ... }:
 let
-  diskoLib = import ./lib {
-    inherit lib;
-    rootMountPoint = config.disko.rootMountPoint;
-    makeTest = import (pkgs.path + "/nixos/tests/make-test-python.nix");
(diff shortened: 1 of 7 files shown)


Enzime:
cc @Lassulus @Mic92

Lassulus (approved):
lgtm, not sure if it breaks some obscure use cases, but from what I can imagine it should be fine. Otherwise I hope people can complain we can revert it or fix :)

module.nix
```diff
@@ -1,96 +1,128 @@
...
+      };
     };
+
     memSize = lib.mkOption {
       type = lib.types.int;
       description = ''
         size of the memory passed to runInLinuxVM, in megabytes
       '';
       default = 1024;
     };
```
colemickens, line 112: should memSize have been moved, too?

Enzime: I didn't move it because it's being shared by interactive VM: https://github.com/nix-community/disko/blob/51e3a7e51279fedfb6669a00d21dc5936c78a6ce/lib/interactive-vm.nix#L59

Maybe the comment needs updating

PR #776, opened 2024-09-17
 module: always populate system.build attributes

Enzime:
cc @Mic92 @Lassulus @MatthewCroughan


Mic92:
It's a bit confusing that we don't expose diskoScript by default, because it doesn't print a proper error message if a user forgot to add a Disko configuration. Can we throw an error instead of hiding the attribute?

Mic92 (approved):
Thanks!

PR #778, opened 2024-09-18
make-disk-image: allow pre/post format files in vm

phaer:
...via  `disko.imageBuilder.preFormatFiles` and `disko.imageBuilder.postFormatFiles`.

This enables users to add secrets to their diskoVMs.

`diskoImageScripts` does include `--pre-format-files`, but `diskoImages` didn't support
the same feature so far - making it harder to pre-generate encrypted disk images or to test such images in a VM.


Enzime (commented):
Could you add some tests for this?

module.nix
```diff
@@ -97,6 +97,47 @@ in
         default = { };
       };
 
+      preFormatFiles = lib.mkOption {
+        type = lib.types.attrsOf (lib.types.either lib.types.str lib.types.path);
+        description = ''
+          Files to copy into the image builder VM before disko is run.
+          This is useful to provide secrets like LUKS keys, or other files you need for formatting.
```
Mic92, line 104: Should we not have a disclaimer, that this leaks secrets into the nix store?
For image builder that runs outside of the nix store, we have a parameter that can be used instead.

phaer: A warning that it could leak is probably a good idea. 

But the second variant of passing a shell snippet should be usable without leaking secrets into the store? i.e.

`preFormatFiles."/tmp/secret" = "cat /my-secret-out-side-nix store > $out"` shouldn't leak. $out is just a tempfile here, see prepareFiles above.

module.nix
```diff
@@ -97,6 +97,47 @@ in
...
+
+          Names are interpreted as destination paths. If the value is a store path,
+          that path will be copied as-is. If it's not a store path, the value will be interpreted
+          as shell code is expected to write files into $out.
+        '';
+        default = {};
+        example = lib.literalExpression ''
+          {
+            "/tmp/pre/file" = pkgs.writeText "foo" "bar";
+            "/tmp/pre/script" = "mkdir -p $out/foo; echo bar > $out/foo";
```
phaer, line 114:
```suggestion
            "/tmp/pre/script" = "mkdir -p $out; echo bar > $out/foo";
```

module.nix
zackw: "`cat /my-secret-outside-nix > $out`" doesn't work:

```
error: builder for '/nix/store/ragb6hgfl9wvdyb9g6721mfyymn3w0mg-myvm-disko-images.drv' failed with exit code 1;
       last 2 log lines:
       > Formatting 'main.qcow2', fmt=qcow2 cluster_size=65536 extended_l2=off compression_type=zlib size=1073741824 lazy_refcounts=off refcount_bits=16
       > cat: /path/to/secret.txt: No such file or directory
```

I think this is because the `cat` command is being executed inside the builder sandbox, so it cannot access files outside the store, or something like that.

musjj: Is it not possible to add the files outside of nix's build process (e.g. via a shell script run externally by `disko`)? I think leaking secrets like this is a huge footgun.

PR #779, opened 2024-09-18
Add simpler alternative to running disko

adam248:
When this command is run it is usually inside of the nixos installer live env so you have to type it all out. using nix-shell is easier to type.

### DIFF
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -128,8 +128,10 @@ The following step will partition and format your disk, and mount it to `/mnt`.
 
 **Please note: This will erase any existing data on your disk.**
 
-```console
+```sh
 sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko -- --mode disko /tmp/disk-config.nix
+# Or use nix-shell. Be aware that this might run an older version of disko!
+nix-shell -p disko --run "sudo disko --mode disko /tmp/disk-config.nix"
 ```
 
 After the command has run, your file system should have been formatted and


docs/quickstart.md
```diff
@@ -130,6 +130,8 @@ The following step will partition and format your disk, and mount it to `/mnt`.
 
 ```console
 sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko -- --mode disko /tmp/disk-config.nix
+# or alternatively you can use nix-shell
```
Mic92, line 133: The difference here is that this comes from nixpkgs, which might be ok for the tutorial actually.

adam248:
thanks for the review @Mic92. It was just a small addition to the docs that I noticed from my VM testing that it was a bit of a pain point for me.
I will mark the comment as resolved.

iFreilicht:
I added a disclaimer that the disko from nixpkgs might be outdated and fixed the syntax highlighting of the comment. I think this is a good addition and I would merged it if @Mic92 is fine with it now.

I also opened #819 so people running outdated versions of disko should be easily detectable for us in bug reports.

docs/quickstart.md
```diff
@@ -128,8 +128,10 @@ The following step will partition and format your disk, and mount it to `/mnt`.
 
 **Please note: This will erase any existing data on your disk.**
 
-```console
+```sh
 sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko -- --mode disko /tmp/disk-config.nix
+# Or use nix-shell. Be aware that this might run an older version of disko!
```
Mic92, line 133:
```suggestion
# Or use the potentially older version from nixpkgs. When reporting github issues, make sure though that you also test with the latest version from this github repository!
```

Enzime: I think if we're just going to tell the users to run the latest version of disko from this repository if issues come up, then I don't think we should add the `nix-shell -p disko` command to the docs

Enzime (changes requested):

Mic92:
I also think that https://github.com/nix-community/disko/pull/826 now is short enough while using the same disko version for both nixos modules and the cli, which is the preferred way of using disko.

PR #788, opened 2024-09-20
test build time with xcp

Mic92:
Depends on https://github.com/NixOS/nixpkgs/pull/362081

### DIFF
--- a/disko-install
+++ b/disko-install
@@ -256,7 +256,7 @@ main() {
   if [[ ! -f "${mountPoint}/nix/var/nix/db/db.sqlite" ]]; then
     echo "Copying store paths" >&2
     mkdir -p "${mountPoint}/nix/store"
-    xargs cp --recursive --target "${mountPoint}/nix/store"  < "${closure_info}/store-paths"
+    xargs xcp --recursive --target-directory "${mountPoint}/nix/store"  < "${closure_info}/store-paths"
     echo "Loading nix database" >&2
     NIX_STATE_DIR=${mountPoint}/nix/var/nix nix-store --load-db < "${closure_info}/registration"
   fi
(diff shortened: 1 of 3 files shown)


Mic92:
Small improvement: 3:18 -> 2:48
Maybe more for bigger trees.

Mic92:
The difference is not big enough, to justify merging this now.
I will wait until the patch is merged & released.

Mic92:
The flag was added. Waiting for an xcp release.

Mic92:
@mergify rebase

iFreilicht:
Notes:
- current latest version of xcp (without the flag) is v0.21.3
- check at https://github.com/tarka/xcp/releases if a new version is available
- submit upgrade PR to nixpkgs if necessary

Once upgrade has landed in nixpkgs, this PR is unblocked.

iFreilicht:
xcp 0.22.0 was released, nixpkgs PR opened: https://github.com/NixOS/nixpkgs/pull/355776

tests/disko-install/default.nix
```diff
@@ -21,6 +21,7 @@ pkgs.nixosTest {
   nodes.machine = {
     virtualisation.emptyDiskImages = [ 4096 ];
     virtualisation.memorySize = 3000;
+    virtualisation.cores = 8;
```
iFreilicht, line 24: Why? I feel like this might not be a good idea, 8 threads are not as common as 3GB of RAM.

iFreilicht (changes requested):

tests/disko-install/default.nix
Mic92: That was me testing how qemu + virtio + xcp scales with threads. Not so much so far.

PR #790, opened 2024-09-21
make-disk-image: add option to disable KVM requirement.

0x450x6c:


module.nix
```diff
@@ -25,6 +25,15 @@ in
         example = lib.literalExpression "\${pkgs.qemu_kvm}/bin/qemu-system-aarch64";
       };
 
+      requireKVM = lib.mkOption {
+        type = lib.types.bool;
+        description = ''
+          By default vmTools.runInLinuxVM requires KVM, see https://github.com/NixOS/nixpkgs/issues/67079
+          This option allows to drop this requirement.
+        '';
+        default = pkgs.stdenv.isx86_64;
```
Mic92, line 34: Why only x86_64? KVM also works on other architectures fine.

Mic92: Also could this not be fixed in nixpkgs instead? https://github.com/NixOS/nixpkgs/blob/e44529f70c6b36e18232909f4f15ee7aae50c109/pkgs/build-support/vm/default.nix#L337

Enzime: If this is something specific that you need that's not upstreamable you can do this in your own configuration:

```nix
(lib.overrideDerivation config.system.build.diskoImages (_: { requiredSystemFeatures = [ ]; }))
```

0x450x6c: I've encountered this issue on building `vmWithDisko`:
```
error: a 'aarch64-linux' with features {kvm} is required to build '/nix/store/gadh261bs1lrhbhpa925ig38yz3kl4md-nixos-disko-images.drv', but I am a 'aarch64-darwin' with features {apple-virt, benchmark, big-parallel, nixos-test}
```

I'm using aarch64-linux builder launched via `nix run nixpkgs#darwin.linux-builder` on aarch64-darwin system.

> Why only x86_64? KVM also works on other architectures fine.

Assumed that on x86_64 the probability of having builder with KVM is higher, it could be wrong assumption though.

Can you suggest better default value?

> Also could this not be fixed in nixpkgs instead?

It can be added as option, however not sure how to change that option when building `vmWithDisko`.

I've tried to change `runInLinuxVM` via overlay and via `disko.imageBuilder.pkgs`, but it didn't work.

---
@Enzime I'm not using `config.system.build.diskoImages` directly, but using `vmWithDisko`.

Tried `system.build.vmWithDiskoNoKVM = lib.overrideDerivation config.system.build.vmWithDisko (_: { requiredSystemFeatures = [ ]; });`, but didn't work.

iFreilicht: > Can you suggest better default value?

KVM is linux-specific, so `isLinux`?

iFreilicht: > Can you suggest better default value?

From the qemu docs:

image

If the system is Linux and one of these architectures, it should be true, otherwise false.

0x450x6c: Discovered better solution, I'll close this PR, please check: https://github.com/nix-community/disko/pull/791

0x450x6c: Using  `diskoImagesScript` instead of `diskoImages` in #791 didn't help either, because some packages used in script `diskoImagesScript` failing to build on aarch64-darwin.

0x450x6c: Changed default value to `isLinux && (isAarch64 || isMips || isx86_64 || isx86_32)`, couldn't find `PPC`, `RISC-V` and `s390x` in stdenv.

0x450x6c: Reason why substituting `runInLinuxVM` via overlay didn't work:

I used `vmTools = vmTools // { runInLinuxVM = ...; }`, however [here](https://github.com/nix-community/disko/blob/master/lib/make-disk-image.nix#L14) we can see override of `vmTools`.

It works fine with `lib.makeOverridable`:
```
        disko.imageBuilder.pkgs =
          pkgs
          // {
            vmTools =
              lib.makeOverridable
              (opts: let
                vmTools = pkgs.vmTools.override opts;
              in
                vmTools
                // {
                  runInLinuxVM = drv:
                    lib.overrideDerivation
                    (vmTools.runInLinuxVM drv)
                    (_: {requiredSystemFeatures = [];});
                })
              {};
          };
```

0x450x6c:
Closing, as there is workaround:
```
        disko.imageBuilder.pkgs =
          pkgs
          // {
            vmTools =
              lib.makeOverridable
              (opts: let
                vmTools = pkgs.vmTools.override opts;
              in
                vmTools
                // {
                  runInLinuxVM = drv:
                    lib.overrideDerivation
                    (vmTools.runInLinuxVM drv)
                    (_: {requiredSystemFeatures = [];});
                })
              {};
          };
```

PR #795, opened 2024-09-23
Add binlore for disko and disko-install

Jayman2000:
See the commit message for details. Also, you might want to view this PR’s commit with `git show --word-diff`. GitHub’s diff for this commit is kind of hard to read.

### DIFF
--- a/package.nix
+++ b/package.nix
@@ -1,29 +1,36 @@
-{ stdenvNoCC, makeWrapper, lib, path, nix, coreutils, nixos-install-tools }:
+{ stdenvNoCC, makeWrapper, lib, path, nix, coreutils, nixos-install-tools, binlore }:
 
-stdenvNoCC.mkDerivation (finalAttrs: {
-  name = "disko";
-  src = ./.;
-  nativeBuildInputs = [
-    makeWrapper
-  ];
-  installPhase = ''
-    mkdir -p $out/bin $out/share/disko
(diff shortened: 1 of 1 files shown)


package.nix
```diff
@@ -1,29 +1,35 @@
...
-})
+      for i in disko disko-install; do
+        sed -e "s|libexec_dir=\".*\"|libexec_dir=\"$out/share/disko\"|" "$i" > "$out/bin/$i"
+        chmod 755 "$out/bin/$i"
+        wrapProgram "$out/bin/$i" \
+          --prefix PATH : ${lib.makeBinPath [ nix coreutils nixos-install-tools ]} \
+          --prefix NIX_PATH : "nixpkgs=${path}"
+      done
+    '';
+    passthru.binlore.out = binlore.synthesize self ''
```
Mic92, line 22:
```suggestion
    # Otherwise resholve thinks that disko and disko-install might be able to execute their arguments
    passthru.binlore.out = binlore.synthesize self ''
```

Jayman2000: Done.

iFreilicht:
@Jayman2000 I'm currently updating the package in nixpkgs. Is it necessary to add `passthru.binlore` there as well?

Jayman2000:
It’s not necessary, but it would definitely be appreciated.

PR #796, opened 2024-09-24
lvm: add missing dm-snapshot

Mic92:
Usually this is added by nixos-generate-config.
However now that we run nixos-generate-config before disko or we even run nixos-facter, we likely not see the LVM block devices anymore.


Mic92:
cc @Lassulus

Mic92:
@mergify queue

PR #801, opened 2024-09-26
make-disk-image: use memSize as default build-memory in diskoImagesScript

danjujan:
Quality of life improvement. Forgetting to specify `--build-memory` otherwise leads to long waiting times until the vm runs out of memory and the script fails.


iFreilicht:
Yeah that seems reasonable, thank you!

PR #802, opened 2024-09-26
Example // Btrfs raid with LUKS encryption. 

Sk7Str1pe:


iFreilicht:
Thank you! Do you need help with adding the test as well?

Mic92:
Unless this handles an edge case we haven't covered in disko yet, this can also go into https://github.com/nix-community/disko-templates/

PR #803, opened 2024-09-28
interactive-vm: override `forceImportRoot`

Enzime:
Some users will have `boot.zfs.forceImportRoot = false;` in their configurations which conflicts with `boot.zfs.forceImportAll = true;`, so we set it to `true` to match.


Mic92:
@mergify queue

Issue #651, opened 2024-05-28
resumeDevice for swap on btrfs


MithicSpirit:
How does one enable `resumeDevice` for a swapfile that is in a btrfs subvolume? Looking through https://github.com/nix-community/disko/blob/master/lib/types/btrfs.nix, there seem to be no options for this. I think it would also be nice to support other options from https://github.com/nix-community/disko/blob/master/lib/types/swap.nix if possible, such as `priority`, `randomEncryption`, and `discardPolicy` (although I am not as interested in these).

Anomalocaridid:
As far as I am aware, there is currently no way to enable this with disko. Instead, I use LVM and a dedicated swap volume instead of a btrfs subvolume, if that helps at all.

nvmd:
@MithicSpirit some of the options you mention are a part of PR #752

fhilgers:
Hey, just wanted to let you know that with `boot.initrd.systemd.enable` swap just works.
Systemd automatically uses the swap device as resume target and also caclulates the resume offset when hibernating. On an efi device this information will be stored in the efivars and systemd reads it again on resume.
Read more here: https://wiki.archlinux.org/title/Power_management/Suspend_and_hibernate#Hibernation

KingKrouch:
> Hey, just wanted to let you know that with `boot.initrd.systemd.enable` swap just works. Systemd automatically uses the swap device as resume target and also caclulates the resume offset when hibernating. On an efi device this information will be stored in the efivars and systemd reads it again on resume. So you do not have to specify any kernel parameters or resumeDevice manually. Read more here: https://wiki.archlinux.org/title/Power_management/Suspend_and_hibernate#Hibernation

Got a good guide on how I'd enable swap hibernate on my Nix config then? I'm using a swapfile on a LUKS encrypted BTRFS subvolume. I already have systemd enabled in the initrd.

fhilgers:
I just created my swapfile using disko like in this example: https://github.com/nix-community/disko/blob/master/example/luks-btrfs-subvolumes.nix

Then when i do systemctl hibernate everything just works. On boot I select the latest generation in systemd boot (the autoselected one) and enter the luks passphrase.
Maybe it has something to do with the systemd version. Storing resume_offset in efivars is a rather recent feature.

KingKrouch:
> I just created my swapfile using disko like in this example: https://github.com/nix-community/disko/blob/master/example/luks-btrfs-subvolumes.nix
> 
[...]

PR #805, opened 2024-10-01
Fix flaky zfs test

Mic92:


lib/tests.nix
```diff
@@ -101,7 +101,8 @@ let
           # config for tests to make them run faster or work at all
           documentation.enable = false;
           hardware.enableAllFirmware = lib.mkForce false;
-          boot.initrd.preDeviceCommands = ''
+          # FIXME: we don't have an systemd in stage-1 equialvent for this
+          boot.initrd.preDeviceCommands = lib.mkIf (!config.boot.initrd.systemd.enable) ''
```
iFreilicht, line 105: This is an unrelated drive-by, right? It doesn't seem to be related to the rest of the PR, and the test still built fine multiple times without it. What's it for? Something to do with the discussion in #651?

iFreilicht (commented):
I ran it with the following script:

```
count=0; while nom build --rebuild -f tests/zfs-with-vdevs.nix; do count=$((count+1)); echo "---SUCCESS COUNT: $count ---"; done
```

After about 20 iterations, each taking 1:00-1:20 minutes, I think it's safe to say the test is no longer flaky :+1:

Enzime:
@iFreilicht does that command fail before this PR?

iFreilicht:
@Enzime Not fail, but the test never finishes. I've left it running for 7 minutes, and it still hangs at the reboot. After the PR, it finishes successfully consistently.

iFreilicht (approved):

PR #808, opened 2024-10-01
docs: Fix /boot security hole warning in examples

### LINKED ISSUE #527: security hole /boot and /boot/loader/random-seed
I'm using disko to setup the partitions of my system via flake. 

started to get these warnings recently:

image

the proposed solution without disko is
```
  fileSystems."/boot" = {
    options = [ "umask=0077" ];
  };
```

though not sure how to set it with disko

iFreilicht:
The alternative would be to do this automatically if format=="vfat" and mountpoint=="/boot", but it's better to be upfront about this.

Fixes #527

### DIFF
--- a/docs/HowTo.md
+++ b/docs/HowTo.md
@@ -150,6 +150,7 @@ or with pinning:
                 type = "filesystem";
                 format = "vfat";
                 mountpoint = "/boot";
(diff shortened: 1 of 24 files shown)


Mic92:
@mergify queue

PR #810, opened 2024-10-03
enable LUKS initrdUnlock only if Systemd's initrd available by default

missrce:
I could be wrong but I am pretty sure without Systemd's initrd available this will not function so I think it should be dependent on the state of boot.initrd.systemd.enable not always enabled.

### DIFF
--- a/lib/types/luks.nix
+++ b/lib/types/luks.nix
@@ -81,7 +81,7 @@ in
     };
     initrdUnlock = lib.mkOption {
       type = lib.types.bool;
-      default = true;
+      default = config.boot.initrd.systemd.enable;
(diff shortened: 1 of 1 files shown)


missrce:
I did not realise this would potentially require messing with the whole type system for the lib, this is not worth the effort when NixOS assertions will notify the user instead I am pretty sure.

Enzime (commented):
`boot.initrd.luks.devices` is used with both systemd and non-systemd stage1 boots

This option existed before systemd stage1 boots was merged:

https://github.com/NixOS/nixpkgs/blob/fd78240ac82ada2b598d4491dbf6ff8622bd3dff/nixos/modules/system/boot/luksroot.nix#L537-L545

PR #811, opened 2024-10-04
interactive-vm: allow to override virtualisation.memorySize

Mic92:
Users might set this value for building images but than need different values when they run interactive virtual machines.

### DIFF
--- a/docs/interactive-vm.md
+++ b/docs/interactive-vm.md
@@ -5,13 +5,7 @@ config.system.build.vm). Simply import the disko module and build the vm runner
 with:
 
 ```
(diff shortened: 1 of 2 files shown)


Enzime (approved):
LGTM

Enzime:
I wonder if we should just remove that line altogether, and tell users to set `virtualisation.vmVariantWithDisko.virtualisation.memorySize`

iFreilicht:
I like that you don't have to do it by default to get a VM you can test with, so I would vote against that.

iFreilicht:
@mergify queue

Issue #789, opened 2024-09-20
Rewrite it in python, rust, or go


iFreilicht:
## Problem

Disko is currently somewhat tricky to maintain. Part of this comes from the current architecture that puts nix itself at the center of the entire operation, and uses it as a crude templating engine for bash scripts.

This makes implementation of some features difficult. A few examples (I'll add more when I come across them):

- #728
- #772 

## Proposed solution

Switch to a different architecture where the nix files only define configuration values, and a bigger program written in a proper language (not bash or perl) runs `nix eval --json` and works with the JSON output to plan and execute all the required actions.

## Considerations

### Compatibility

The interface for users should not change at all. `disko.devices` should have the exact same type and result in the same configuration values as before.

Some programs (most notable nixos-anywhere, AFAIK) use disko's `diskoScript` output and call this script directly. Because this is a derivation, it should be possible to just swap them out no matter the language, as long as the resulting path has a binary with the same name. However, a simple shim with a single-line script that just calls the CLI with the required flags might be sufficient for this.

### Language to use

The language should make it easy for others to contribute, have decent error-handling capabilities and be well-suited to write a CLI in.

I am therefore immediately discarding bash (or any other shell language), perl, JS, TS and Haskell.

I see three main contenders: python, rust, and go. They are all well-known and have good general support. Rust is additionally loved in the nix community for its ethos of correctness, python is super easy to write, and go is kind-of a middle-ground between the two.

One thing to consider with this choice might be closure size. As disko is often run in installers, we can expect most dependencies to not be present yet (this is the singular advantage of bash; it is guaranteed to already be present). The closure sizes of the main executable vary quite a bit:

```
$ nix path-info --closure-size -h nixpkgs#cargo^out
/nix/store/j6vpbh78ngj2x7jrg27pqjn2cd5mzywl-cargo-1.79.0          1.2G
$ nix path-info --closure-size -h nixpkgs#go^out                                                                                                                                                                                                                                                                                          
/nix/store/kqf4gwhq1yrchvmvqjh4adc4y9hxp9xk-go-1.22.5    248.0M
$ nix path-info --closure-size -h nixpkgs#python3^out                                                                                                                                                                                                                                                                                    
/nix/store/l014xp1qxdl6gim3zc0jv3mpxhbp346s-python3-3.12.4       166.1M
```

It seems somewhat unreasonable to me to require the user to download 1.2 gigabytes just to be able to run disko. A binary cache could alleviate this, though this has additional usability implications.

[...]

PR #818, opened 2024-10-09
Allow config to be dumped to json file

Yethal:
I'll accept any tips&tricks on this one, I'm still learning Nix
Necessary for https://github.com/nix-community/disko/issues/789

Tested via following script:
```
#!/usr/bin/env nu
ls example
| where type == file
| update name {path expand}
| par-each {|example|
    let dir_name = ('configs'|path join ($example.name|path parse|get stem))
    mkdir $dir_name
    let store_path = (nix-build ./cli.nix --argstr mode config --no-out-link --arg diskoFile $example.name)
        cp $store_path $dir_name
}
```

### DIFF
--- a/cli.nix
+++ b/cli.nix
@@ -24,6 +24,7 @@ let
         # legacy aliases
         create = "createScriptNoDeps";
         zap_create_mount = "diskoScriptNoDeps";
+        config = "configFile";
       }.${mode}
     else
       {
@@ -34,6 +35,7 @@ let
         # legacy aliases
         create = "createScript";
         zap_create_mount = "diskoScript";
+        config = "configFile";
       }.${mode};
 
   hasDiskoConfigFlake =
--- a/default.nix
+++ b/default.nix
@@ -40,5 +40,7 @@ in
   diskoNoDeps = cfg: pkgs: builtins.trace "the diskoNoDeps output is deprecated, please use disko instead" ((eval cfg).config.disko.devices._scripts { inherit pkgs checked; }).diskoScriptNoDeps;
 
   config = cfg: (eval cfg).config.disko.devices._config;
+  configFile = cfg: pkgs: pkgs.writers.writeJSON "config.json" cfg.disko.devices;
+
   packages = cfg: (eval cfg).config.disko.devices._packages;
 }


Yethal:
@phaer

iFreilicht:
You can already do this with any disko configuration with the following command:

```
nix eval --json -f example/boot-raid1.nix
```

Yethal:
Will it end up in the store where the new(?) improved(?) disko script will be able to read it? The design discussed under https://github.com/nix-community/disko/issues/789 assumes a script that accepts the entire disko config as a json file

iFreilicht:
> Will it end up in the store where the new(?) improved(?) disko script will be able to read it? The design discussed under https://github.com/nix-community/disko/issues/789 assumes a script that accepts the entire disko config as a json file

No, but it doesn't need to as far as I see it. There are basically two cases for where the  configuration can come from: a bare nix file which can be evaluated on its own, or a flake.
For the deployment scenario, we have local execution or a remote deployment with nixos-anywhere.

When you deploy with nixos-anywhere, you always send over the full configuration (including the disko config) to the target's store and build there. The same happens when you use a flake, even on a local deployment. Only when running disko locally with a bare `.nix` file will the config not end up in the store, but it doesn't need to, because you can evaluate it directly.

In all these cases, the program (I don't want to call it a script) will have to do the following things:

1. `nix eval --json` (probably with an intermediate nix file that injects function arguments if required, like `cli.nix` does now) to ingest the configuration
2. Transform it into a representation of the **desired** state the drives should have
3. Inform the user about any logical errors in that configuration (e.g. raid1 with only one drive)
3. Probe the OS and create the same representation of the **current** state the drives have
4. Generate a diff of sorts that shows what steps have to be taken to turn the **current** state into the **desired** state
5. Tell the user what these steps are and ask for confirmation*
6. Execute as many of these steps as possible
7. Comprehensively explain what went wrong and why

*the UX for this has not been decided. For sure there will have to be a flag to allow for an unattended installation, and it would be nice if the user could interactively decide which steps to take, but that's a stretch goal.

Maybe this explanation should go into the main thread.

One thing that is very important here is that we do NOT want to use nix as a templating engine to generate this program. The program's **behavior** should be a function of the configuration, not the **program itself**.

> Also, `nix-eval --json` fails if a given disko config is a function not an attrset (i.e example/luks-on-mdadm.nix)

True, for these configs you'll need to do something more sophisticated, like[ `cli.nix` already does](https://github.com/nix-community/disko/blob/d39ee334984fcdae6244f5a8e6ab857479cbaefe/cli.nix#L53C1-L56C19).

Yethal:
So nix would perform steps 1-3 and if no errors are present it would generate a plan (in the form of json) that would then be fed to the program. Steps 4-8 would be executed by the program with no nix involvement.
If that's how the concerns should be separated then how should this PR be adjusted? Or is it even needed at all?

I can do the program part of this but need help/guidance in regards to Nix side of the process.

btw, steps 4-8 look a lot like how Terraform works but with local resources instead of remote.

iFreilicht:
> So nix would perform steps 1-3 and if no errors are present it would generate a plan (in the form of json) that would then be fed to the program. Steps 4-8 would be executed by the program with no nix involvement.

No, not even that. Nix will only do step 1. Everything else is handled by the program, even step 3 (which currently is done by nix). We have quite a few issues with validating the config that are very difficult if not impossible to solve in nix.

Well, the only exception is things that influence the NixOS configuration, but that is implemented already, so we don't need to worry about it for this.

> I can do the program part of this but need help/guidance in regards to Nix side of the process.

My suggestion would be to not worry about disko configs that are functions for now, only those that can be evaluated easily by a call to `nix eval`. I also started working on this (though I don't have much time right now, so please don't let that stop you), and the best approach for prototyping this is to write a disko config for your current system eval it, and then try to implement step 3 and 4 until both reach the same result.

If you need additional help, feel free to add me on matrix (@ifreilicht:matrix.org) or DM me on the NixOS discourse :)

> btw, steps 4-8 look a lot like how Terraform works but with local resources instead of remote.

Jup, that's where I got the idea from. I think the biggest difference will be that we don't store the current state but regenerate it, and how step 6 will be presented, but the general concept is indeed the same.

> If that's how the concerns should be separated then how should this PR be adjusted? Or is it even needed at all?

I don't think it's needed. I'll close it :)

PR #819, opened 2024-10-09
Add version command

### LINKED ISSUE #745: Add a `--version` command
Right now, it's impossible to see which version of disko I'm running. Nixpkgs shows it:

Screenshot 2024-08-22 at 17 33 05

but I can't verify whether I'm running the latest release:

```
$ disko --version
aborted: disko config must be an existing file or flake must be set
$ which disko
/nix/store/ply01w0y86dipp2n2ysfi7cs2xgj4pvw-disko/bin/disko
```

The only way to know is to check out the commit of nixpkgs I have installed and look at the code of `pkgs/by-name/di/disko/package.nix`.

It would be nicer if there was a `--version` command. Not sure how this could best be implemented, though.

iFreilicht:
Fixes #745

This adds a script to create a release. Running it will create two
commits that would appear like this in `git log --oneline`:

    67b5fff (master) release: reset released flag
    0b808f3 (tag: v1.8.1) release: v1.8.1
    100d2f3 docs: last change before release

It also re-creates the `version.nix` file, which is used by the flake
to determine the final version the disko CLI will print.

If `disko --version` is run from exactly the commit tagged `v1.8.1`, it
will print `1.8.1`. If it is run from any commit past that (like
master), it will print `1.8.1-67b5fff`, and if it is run from a local
folder with uncommitted changes, it will print `1.8.1-67b5fff-dirty`.

I decided that the version should always be printed as well. This should make it easy to determine whether users were running an outdated version when reporting an issue.

Additionally, I added the version tag to all flake references of the docs. This ensures that people will always use the current release, not just whatever the current master is.
The release script will update this as well. I understand this might be a controversial step, and I'm happy to hear arguments against it.

When this PR is merged, I would like to make a v1.8.1 release and update the disko package in nixpkgs as well so starting with NixOS 24.11, disko can report its own version properly.

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ a disk named /dev/sda, you would run the following command to partition, format
 and mount the disk.
 
 ```console
-sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko -- --mode disko /tmp/disk-config.nix
+sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko/latest -- --mode disko /tmp/disk-config.nix
 ```
 
 ## Related Tools
--- a/flake.nix
+++ b/flake.nix
@@ -8,13 +8,17 @@
 
   outputs = { self, nixpkgs, ... }:
     let
+      lib = nixpkgs.lib;
       supportedSystems = [
         "x86_64-linux"
         "i686-linux"
         "aarch64-linux"
         "riscv64-linux"
       ];
-      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+      forAllSystems = lib.genAttrs supportedSystems;
+
+      versionInfo = import ./version.nix;
+      version = versionInfo.version + (lib.optionalString (!versionInfo.released) "-dirty");
     in
     {
       nixosModules.default = self.nixosModules.disko; # convention
@@ -27,12 +31,14 @@
           pkgs = nixpkgs.legacyPackages.${system};
         in
         {
-          disko = pkgs.callPackage ./package.nix { };
+          disko = pkgs.callPackage ./package.nix { diskoVersion = version; };
           # alias to make `nix run` more convenient
           disko-install = self.packages.${system}.disko.overrideAttrs (_old: {
             name = "disko-install";
           });
           default = self.packages.${system}.disko;
+
+          create-release = pkgs.callPackage ./scripts/create-release.nix { };
         } // pkgs.lib.optionalAttrs (!pkgs.buildPlatform.isRiscV64) {
           disko-doc = pkgs.callPackage ./doc.nix { };
         });
@@ -42,7 +48,7 @@
         let
           pkgs = nixpkgs.legacyPackages.${system};
           # FIXME: aarch64-linux seems to hang on boot
-          nixosTests = nixpkgs.lib.optionalAttrs pkgs.hostPlatform.isx86_64 (import ./tests {
+          nixosTests = lib.optionalAttrs pkgs.hostPlatform.isx86_64 (import ./tests {
             inherit pkgs;
             makeTest = import (pkgs.path + "/nixos/tests/make-test-python.nix");
             eval-config = import (pkgs.path + "/nixos/lib/eval-config.nix");
@@ -50,6 +56,7 @@
 
           disko-install = pkgs.callPackage ./tests/disko-install {
             inherit self;
+            diskoVersion = version;
           };
 
           shellcheck = pkgs.runCommand "shellcheck" { nativeBuildInputs = [ pkgs.shellcheck ]; } ''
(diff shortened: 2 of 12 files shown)


README.md
```diff
@@ -99,7 +99,7 @@ a disk named /dev/sda, you would run the following command to partition, format
 and mount the disk.
 
 ```console
-sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko -- --mode disko /tmp/disk-config.nix
+sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko/v1.8.0 -- --mode disko /tmp/disk-config.nix
```
Enzime, line 102:
```suggestion
sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko/latest -- --mode disko /tmp/disk-config.nix
```

It would be nice if we had a branch or tag that always pointed to the latest release, you could also use it in your `flake.nix`:

```nix
{
  inputs.disko.url = "github:nix-community/disko/latest";
}

File: scripts/create-release.sh [hunk 2]
```diff
@@ -0,0 +1,56 @@
+#!/usr/bin/env bash
+
+set -eu -o pipefail
```
Enzime, line 3:
```suggestion
```

It would be good if this was a package/output in the flake, that way you could use `pkgs.writeShellApplication` to ensure it gets run through `shellcheck` and all the dependencies of the script are provided

scripts/create-release.sh
```diff
@@ -0,0 +1,56 @@
...
+# Commit and tag the release
+git commit -am "release: v$version"
+git tag -a "v$version" -m "release: v$version"
+
+# Reset released flag so disko only displays the version without
+# a revsion suffix when run from the tagged release commit
+echo "{ version = \"$version\"; released = false; }" > version.nix
+git commit -am "release: reset released flag"
+
+echo "now run 'git push --tags origin master'"
```
Enzime, line 56: This is missing a trailing newline

flake.nix
```diff
@@ -8,13 +8,19 @@
...
         "aarch64-linux"
         "riscv64-linux"
       ];
-      forAllSystems = nixpkgs.lib.genAttrs supportedSystems;
+      forAllSystems = lib.genAttrs supportedSystems;
+
+      versionInfo = import ./version.nix;
+      version = versionInfo.version +
+        (lib.optionalString (!versionInfo.released)
+          "-${self.shortRev or self.dirtyShortRev or "dirty"}");
```
Mic92, line 23: Can we skip the revision for now? This will trigger a rebuild on very commit no matter if anything changed.

Mic92: If we encourage users to use a release version, than error reporting should be no longer a big issue.

scripts/create-release.sh
```diff
@@ -0,0 +1,56 @@
...
+previous_version=$(nix --extra-experimental-features nix-command eval --raw -f version.nix version)
+
+# Update the version file
+echo "{ version = \"$version\"; released = true; }" > version.nix
+
+# Update all references to the previous version in the docs
+sed -i "s|v${previous_version}|v${version}|g" **/*.md
+
+# Commit and tag the release
+git commit -am "release: v$version"
```
Mic92, line 48: Ideally also run `nix-fast-build` here to make sure that all tests passes.
It delays the release process a bit, but makes sure we don't deliver broken releases.

scripts/create-release.sh
iFreilicht: True! Fixed.

scripts/create-release.sh
iFreilicht: Ahh yes, this also applies the `-e` and `pipefail` options automatically.

It's now available at `nix run .#create-release`

flake.nix
iFreilicht: Ah yes, I didn't consider that. I'd still like to have the `-dirty` suffix for all non-released versions, I think that's a reasonable compromise.

scripts/create-release.sh
iFreilicht: Hmm yes, that makes sense, but I think I need a bit of guidance here. What exactly should `nix-fast-build` build, and with what parameters? I tried to run without `-j` and `--eval-workers`, but that clogged my system so hard I had to reset it. I'm on an i7-4790K, not the newest CPU, but it has 4 cores, 8 threads.

Only `-j 3` and `--eval-workers 3` allowed me to still use the system while the build was happening, but hardcoding these values seems unwise. I'm also unsure if I should just build all `.#checks` or anything else as well.

Mic92: Just go with --eval-workers 3 for now. We only need to run it on a limited set of maintainer machines.

Mic92: Even 1 would be acceptable, because it than just prevents unbound memory usage.

iFreilicht: Alright, done. I tried it out a bit more, and it seems because the worker RAM limit is at 4GB, only two workers is the safest while still just taking 25 minutes for a full build. It seems like maybe it's possible to use more builder threads and less eval workers, eval seems very memory intensive (and potentially leaky), but it's good enough for now.

iFreilicht:
So I incorporated all the suggestions and tried it out on my remote, see [commits](https://github.com/iFreilicht/disko/commits/master/) and [tags](https://github.com/iFreilicht/disko/tags).

However, I'm noticing that the tags are usually signed, and the release script currently isn't doing that. I'll have to set this up locally before trying it out but don't have any more time today.

Enzime:
Do you plan to create GitHub releases as well? If not, it might be worth just adding a reminder message

Mic92 (approved):
@mergify queue

Mic92:
@mergify queue

Issue #823, opened 2024-10-11
`latest` tag might not work as intended


iFreilicht:
Despite having force-pushed the `latest` tag, I still get the previous version:

```console
# nix flake metadata github:nix-community/disko/v1.8.2
Resolved URL:  github:nix-community/disko/v1.8.2
Locked URL:    github:nix-community/disko/ff0a471763faaaca1859fd6de80f44fa0fce91a6
Description:   Disko - declarative disk partitioning
Path:          /nix/store/4k3ka3z47d5aysf8wijvfn9hnv7nvsal-source
Revision:      ff0a471763faaaca1859fd6de80f44fa0fce91a6
Last modified: 2024-10-11 21:02:24
Inputs:
└───nixpkgs: github:NixOS/nixpkgs/b833ff01a0d694b910daca6e2ff4a3f26dee478c
    Last modified: 2024-09-01 14:44:31
# nix flake metadata github:nix-community/disko/latest
Resolved URL:  github:nix-community/disko/latest
Locked URL:    github:nix-community/disko/624fd86460e482017ed9c3c3c55a3758c06a4e7f
Description:   Disko - declarative disk partitioning
Path:          /nix/store/mk1zl7iip7nlf0fggg7dmgnrgjfzfahc-source
Revision:      624fd86460e482017ed9c3c3c55a3758c06a4e7f
Last modified: 2024-09-19 21:58:46
Inputs:
└───nixpkgs: github:NixOS/nixpkgs/b833ff01a0d694b910daca6e2ff4a3f26dee478c
    Last modified: 2024-09-01 14:44:31
```

After running `nix-collect-garbage`, it gets the correct version again:

```console
# nix flake metadata github:nix-community/disko/latest
Resolved URL:  github:nix-community/disko/latest
Locked URL:    github:nix-community/disko/ff0a471763faaaca1859fd6de80f44fa0fce91a6
Description:   Disko - declarative disk partitioning
Path:          /nix/store/4k3ka3z47d5aysf8wijvfn9hnv7nvsal-source
Revision:      ff0a471763faaaca1859fd6de80f44fa0fce91a6
Last modified: 2024-10-11 21:02:24
Inputs:
└───nixpkgs: github:NixOS/nixpkgs/b833ff01a0d694b910daca6e2ff4a3f26dee478c
    Last modified: 2024-09-01 14:44:31
```

So, yeah. Something to figure out before the next release.

Maybe revert to using the auto-updating docs I originally had in #819

[...]

PR #826, opened 2024-10-12
add `--refresh` to all nix run examples

Mic92:
fixes https://github.com/nix-community/disko/issues/823

### DIFF
--- a/README.md
+++ b/README.md
@@ -99,7 +99,7 @@ a disk named /dev/sda, you would run the following command to partition, format
 and mount the disk.
 
 ```console
-sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko/latest -- --mode disko /tmp/disk-config.nix
+sudo nix run --refresh disko/latest -- --mode disko /tmp/disk-config.nix
 ```
 
 ## Related Tools
--- a/docs/disko-install.md
+++ b/docs/disko-install.md
@@ -22,7 +22,7 @@ For a fresh installation, where **disko-install** will handle partitioning and
 setting up the disk, use the following syntax:
 
 ```console
-sudo nix run 'github:nix-community/disko#disko-install/latest' -- --flake <flake-url>#<flake-attr> --disk <disk-name> <disk-device>
+sudo nix run --refresh 'disko/latest#disko-install' -- --flake <flake-url>#<flake-attr> --disk <disk-name> <disk-device>
 ```
 
 Example:
@@ -107,7 +107,7 @@ nvme0n1     259:0    0  1.8T  0 disk
 In our example, we want to install to a USB-stick (/dev/sda):
 
 ```console
-$ sudo nix run 'github:nix-community/disko/latest#disko-install' -- --flake '/tmp/config/etc/nixos#mymachine' --disk main /dev/sda
+$ sudo nix run --refresh 'disko/latest#disko-install' -- --flake '/tmp/config/etc/nixos#mymachine' --disk main /dev/sda
 ```
 
 Afterwards you can test your USB-stick by either selecting during the boot or
@@ -126,7 +126,7 @@ new hardware or to prioritize it in your current machine's boot order, use the
 --write-efi-boot-entries option:
 
 ```console
-$ sudo nix run 'github:nix-community/disko/latest#disko-install' -- --write-efi-boot-entries --flake '/tmp/config/etc/nixos#mymachine' --disk main /dev/sda
+$ sudo nix run --refresh 'disko/latest#disko-install' -- --write-efi-boot-entries --flake '/tmp/config/etc/nixos#mymachine' --disk main /dev/sda
 ```
 
 This command installs NixOS with **disko-install** and sets the newly installed
--- a/docs/quickstart.md
+++ b/docs/quickstart.md
@@ -129,7 +129,7 @@ The following step will partition and format your disk, and mount it to `/mnt`.
 **Please note: This will erase any existing data on your disk.**
 
 ```console
-sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko/latest -- --mode disko /tmp/disk-config.nix
+sudo nix run --refresh disko/latest -- --mode disko /tmp/disk-config.nix
 ```
 
 After the command has run, your file system should have been formatted and


Enzime (changes requested):
I don't think this is necessary because I believe the fetcher cache should get invalidated after a couple hours

README.md
```diff
@@ -99,7 +99,7 @@ a disk named /dev/sda, you would run the following command to partition, format
 and mount the disk.
 
 ```console
-sudo nix --experimental-features "nix-command flakes" run github:nix-community/disko/latest -- --mode disko /tmp/disk-config.nix
```
Mic92, line 102: I figured most new people where quickly have flakes enabled anyway and the other ones know there way around.

README.md
```diff
@@ -99,7 +99,7 @@ a disk named /dev/sda, you would run the following command to partition, format
...
+sudo nix run --refresh 'github:nix-community/disko/latest#disko-install' -- --mode disko /tmp/disk-config.nix
```
Enzime, line 102:
```suggestion
sudo nix run disko/latest#disko-install -- --mode disko /tmp/disk-config.nix
```

Enzime (changes requested):

Enzime:
According to https://github.com/NixOS/nix/issues/6830#issuecomment-1209169589, the timeout is controlled by the Nix option [tarball-ttl](https://nix.dev/manual/nix/2.18/command-ref/conf-file.html#conf-tarball-ttl) which defaults to 1 hour

Mic92:
> According to [NixOS/nix#6830 (comment)](https://github.com/NixOS/nix/issues/6830#issuecomment-1209169589), the cache expiry is controlled by the Nix option [tarball-ttl](https://nix.dev/manual/nix/2.18/command-ref/conf-file.html#conf-tarball-ttl) which defaults to 1 hour

Yes and this is very confusing if you made a fix it's not immediately available. I had this issue so many times. I don't want this anymore.

README.md
Mic92: I wouldn't rely on the registry anymore. Many people including me are overriding that with local versions now.

Mic92:
People can copy and paste commands. So I don't think adding a few more characters for --refresh and be sure it's the same thing you see on the website can save you so much time.

README.md
Enzime: I believe when you specify a ref, then it won't use the pinned registry entry locally. For context, I'm using `flake-utils-plus` to pin my registry:

```
$ nix registry list
system flake:disko path:/nix/store/4p84shmrgsdbq2j0dfa2cbqfbxxm05h7-source?lastModified=1728334376&narHash=sha256-CTKEKPzD/j8FK6H4DO3EjyixZd3HHvgAgfnCwpGFP5c%3D&rev=d39ee334984fcdae6244f5a8e6ab857479cbaefe
global flake:disko github:nix-community/disko
```

```
$ nix flake metadata disko
Resolved URL:  path:/nix/store/4p84shmrgsdbq2j0dfa2cbqfbxxm05h7-source?lastModified=1728334376&narHash=sha256-CTKEKPzD/j8FK6H4DO3EjyixZd3HHvgAgfnCwpGFP5c%3D&rev=d39ee334984fcdae6244f5a8e6ab857479cbaefe
Locked URL:    path:/nix/store/4p84shmrgsdbq2j0dfa2cbqfbxxm05h7-source?lastModified=1728334376&narHash=sha256-CTKEKPzD/j8FK6H4DO3EjyixZd3HHvgAgfnCwpGFP5c%3D&rev=d39ee334984fcdae6244f5a8e6ab857479cbaefe
Description:   Disko - declarative disk partitioning
Path:          /nix/store/4p84shmrgsdbq2j0dfa2cbqfbxxm05h7-source
Revision:      d39ee334984fcdae6244f5a8e6ab857479cbaefe
Last modified: 2024-10-08 07:52:56
Fingerprint:   67a8c3587b0de92d19455ea675f0b7e2573bef80b8f9d9b5612df0edd072164b
Inputs:
└───nixpkgs: github:NixOS/nixpkgs/b833ff01a0d694b910daca6e2ff4a3f26dee478c?narHash=sha256-tLGCFEFTB5TaOKkpfw3iYT9dnk4awTP/q4w%2BROpMfuw%3D (2024-09-01 12:44:31)
```

```
$ nix flake metadata disko/latest
Resolved URL:  github:nix-community/disko/latest
Locked URL:    github:nix-community/disko/ff0a471763faaaca1859fd6de80f44fa0fce91a6?narHash=sha256-O0QVhsj9I/hmcIqJ4qCqFyzvjYL%2BdtzJP0C5MFd8O/Y%3D
Description:   Disko - declarative disk partitioning
Path:          /nix/store/4k3ka3z47d5aysf8wijvfn9hnv7nvsal-source
Revision:      ff0a471763faaaca1859fd6de80f44fa0fce91a6
Last modified: 2024-10-12 06:02:24
Fingerprint:   b11e370bd711abe0f9ceb7c64b0d6738b4327e7ee6384d0f6d1d17250d33aa83
Inputs:
└───nixpkgs: github:NixOS/nixpkgs/b833ff01a0d694b910daca6e2ff4a3f26dee478c?narHash=sha256-tLGCFEFTB5TaOKkpfw3iYT9dnk4awTP/q4w%2BROpMfuw%3D (2024-09-01 12:44:31)
```

Enzime:
> > According to [NixOS/nix#6830 (comment)](https://github.com/NixOS/nix/issues/6830#issuecomment-1209169589), the cache expiry is controlled by the Nix option [tarball-ttl](https://nix.dev/manual/nix/2.18/command-ref/conf-file.html#conf-tarball-ttl) which defaults to 1 hour
> 
> Yes and this is very confusing if you made a fix and it's not immediately available. I had this issue so many times. I don't want this anymore.

I agree it's annoying when you run Nix commands and they're silently cached. In this case though, all these commands refer to `disko/latest` which specifically points to the latest release which I doubt would get updated often enough that users would commonly run into this issue as the cache time is 1 hour and they'd have to have already fetched disko in the last hour.

iFreilicht:
Ahh the ttl bit me. So it doesn't matter then if `latest` is even used or not.

> People can copy and paste commands.

Not always, some use disko on bare metal and have to type everything. But for that case we have the nix-shell invocation now.

> Yes and this is very confusing if you made a fix it's not immediately available. I had this issue so many times. I don't want this anymore.

Yeah I agree, this is annoying. The command looks like it got kinda out-of hand, though.

Mic92:
Ok. Now it should be short enough to type out.

README.md
iFreilicht: Ah that's an interesting point. I don't really like relying on the registry because it has been called for to be removed. But I guess if that happens we can still change it. And using it makes the command less intimidating, which is always nice for first-time users.

Mic92:
> Yeah I agree, this is annoying. The command looks like it got kinda out-of hand, though.

The issue is that if you tell someone you just fixed X and they should test it and than it's still their because they use the old version...

README.md
iFreilicht: I don't know that that's true. Maybe a short note that you might have to add the extra experimental commands would be good for first-time users. This is the sort of stuff that makes the nix ecosystem super frustrating to get into: see command in docs, copypaste, doesn't work.

Enzime:
@iFreilicht I'm fairly sure the TTL applies specifically to each flakeref so `disko` and `disko/latest` will be cached independently

@Mic92 I agree that it's annoying if they're not running the latest code, it just seems like an unlikely situation because we're specifically specifying the latest version which I don't think will change super often. If we were using `master` instead then I think it would be a lot more common and make more sense to add `--refresh`

Mic92:
> @Mic92 I agree that it's annoying if they're not running the latest code, it just seems like an unlikely situation because we're specifically specifying the latest version which I don't think will change super often. If we were using `master` instead then I think it would be a lot more common and make more sense to add `--refresh`

If you just closed someone issue with a remark that it has been fixed in the latest version, it will still happen. Especially if you did a faux pas in your latest release and just fixed it with a new version. I would rather prefer version tags over a latest tag, if adding a refresh is so problematic.

Mic92:
And this only will become more common as disko and NixOS get more popular.

Enzime:
@Mic92 in that situation, I think it would make sense to respond to the issue with the new version (so they can also check themselves what version they're running) and mention to them to use the `--refresh` flag

Even though I think that we shouldn't add the `--refresh` flag unconditionally to the docs (how about instead adding a small explanation about running the command with `--refresh` if their `disko` version is out of date), I'm not going to block this PR, feel free to merge it if you want

README.md
Mic92: You get this error message:

bash-5.2# nix run github:nix-community/disko/latest -- --mode disko /tmp/disk-config.nix
error: experimental Nix feature 'nix-command' is disabled; add '--extra-experimental-features nix-command' to enable it

Mic92:
I leave this fix for someone else to pick up.

PR #832, opened 2024-10-14
options: make extraPostVM mergable.

qbisi:
commit1 change extraPostVM type to types.lines.

commit2 add imageName option to disk type.


iFreilicht:
Code in general looks good, the change to extraPostVM is obviously correct 👍 For the `imageName` could you please add an example of how to use it to [the docs here](https://github.com/nix-community/disko/blob/3b2e19fe7c667a4fb2da3d6bb5ae6ee3e3552f74/docs/disko-images.md?plain=1#L91C1-L98C1)?

iFreilicht (approved):
Awesome, thank you!

iFreilicht:
@mergify queue

PR #834, opened 2024-10-14
swap: fix partition type

### LINKED ISSUE #391: Disko sets the wrong partition type for the swap partition
This my disko setting: https://fars.ee/dKd8/nix
After running `nix --extra-experimental-features 'flakes nix-command' run github:nix-community/disko -- --mode disko ./disk/disko-config.nix`, I'll get the partition situation like this: https://fars.ee/4gPJ.png
As you can see, the Swap partition type has been set to `Linux Filesystem' instead of `Linux Swap'.

iFreilicht:
Fixes #391

### DIFF
--- a/lib/types/gpt.nix
+++ b/lib/types/gpt.nix
@@ -26,7 +26,8 @@ in
               lib.types.either
                 (lib.types.strMatching (hexPattern 4))
                 (lib.types.strMatching (lib.concatMapStringsSep "-" hexPattern [ 8 4 4 4 12 ]));
-            default = "8300";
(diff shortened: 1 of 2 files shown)


iFreilicht:
Alright, this was a bit more complicated than I though, but it's done. The test currently requires the GUID of the dm-crypt partition type because lsblk doesn't print it, see https://github.com/util-linux/util-linux/issues/3238

I don't really like that I have to change something in `gpt.nix` that only affects the swap type, but I don't see another way to do it. Maybe it's possible to use `parent` somehow?

Mic92:
@mergify queue

PR #835, opened 2024-10-15
Fix disko using system zfs

### LINKED ISSUE #815: Script not working under hetzner rescue system
I found a "funny" issue.

When running `disko-install`, there are some calls to the `zfs` binary.

It turns out that, on a [Hetzner rescue system](https://docs.hetzner.com/robot/dedicated-server/troubleshooting/hetzner-rescue-system/) (the one used in Hetzner Cloud/Robot for installing other systems on the machine), `zfs` is not available per se, but there exists a different `zfs` binary that prints a warning to you and asks you to press <kbd>y</kbd> to install zfs itself.

See this shell session:

```shell
root@rescue / # zdb --help
The Hetzner Rescue System does not come with preinstalled ZFS support,
however, we will attempt to compile and install the latest release for you.
Please read the information below thoroughly before entering any response.

ATTENTION

This script will attempt to install the current OpenZFS release
which is available in the OpenZFS git repository to the Rescue
System. If this script fails, do not contact Hetzner Support, as
it is provided AS-IS and Hetzner will not support the installation
or usage of OpenZFS due to License incompatiblity (see below).
Due to github.com limitations, this script only works via IPv4.

Licenses of OpenZFS and Linux are incompatible

OpenZFS is licensed under the Common Development and Distribution
License (CDDL), and the Linux kernel is licensed under the GNU General
Public License Version 2 (GPL-2). While both are free open source
licenses they are restrictive licenses. The combination of them causes
problems because it prevents using pieces of code exclusively available
under one license with pieces of code exclusively available under the
other in the same binary.

Please be aware that distributing of the binaries may lead to infringing.

Press y to accept this.
```

Installation logs when I press <kbd>y</kbd>: [zfs-press-y-install-on-hetzner-rescue.log](https://github.com/user-attachments/files/17292356/zfs-press-y-install-on-hetzner-rescue.log)

After installation, `zfs` seems to finally work:

```shell
root@rescue /nix/store/3ys93zsghig63qpcs67vc37x0rwn1zhl-disk-deactivate # zfs --help
usage: zfs command args ...
[...]
```

So, when running `disko-install --flake ...`, I got errors like this:

```
+ lsblk --output-all --json
+ bash -x
++ dirname /nix/store/3ys93zsghig63qpcs67vc37x0rwn1zhl-disk-deactivate/disk-deactivate
+ jq -r --arg disk_to_clear /dev/sda -f /nix/store/3ys93zsghig63qpcs67vc37x0rwn1zhl-disk-deactivate/disk-deactivate.jq
+ set -fu
++ type zdb
++ zdb -l /dev/sda1
++ sed -nr 's/ +name: '\\''(.*)'\\''/\\1/p'
+ zpool=
bash: line 3: syntax error near unexpected token `then'
bash: line 3: `f [[ -n \"${zpool}\" ]]; then zpool destroy -f \"$zpool\"; zpool labelclear -f \"$zpool\"; fi'
```

Which seem to related to this: https://github.com/nix-community/disko/blob/d39ee334984fcdae6244f5a8e6ab857479cbaefe/disk-deactivate/disk-deactivate.jq#L17-L18

But in reality they are related to the fact that `zfs` is not the `zfs` that Disko is expecting.

I can automate this with a simple `yes | zfs` command before running `disko-install`, but I wonder why is Disko using the preinstalled `zfs` binary. Shouldn't it use the one brought by nixpkgs? This would make the script more predictable.

@moduon MT-7504

iFreilicht:
Fixes #815

Also a drive-by improvement: The `with lib;` was completely unused, so I removed it.

@yajo FYI. Maybe you could test this again on the hetzner rescue system? I'm fairly confident it will work, but getting confirmation would be great.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -393,21 +393,22 @@ let
             default = { pkgs, checked ? false }:
               let
                 throwIfNoDisksDetected = _: v: if devices.disk == { } then throw "No disks defined, did you forget to import your disko config?" else v;
+                destroyDependencies = with pkgs; [
+                  util-linux
+                  e2fsprogs
+                  mdadm
+                  zfs
+                  lvm2
+                  bash
+                  jq
+                  gnused
+                  gawk
+                  coreutils-full
+                ];
               in
               mapAttrs throwIfNoDisksDetected {
                 destroyScript = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "disko-destroy" ''
-                  export PATH=${lib.makeBinPath (with pkgs; [
-                    util-linux
-                    e2fsprogs
-                    mdadm
-                    zfs
-                    lvm2
-                    bash
-                    jq
-                    gnused
-                    gawk
-                    coreutils-full
-                  ])}:$PATH
+                  export PATH=${lib.makeBinPath destroyDependencies}:$PATH
                   ${cfg.config._destroy}
                 '';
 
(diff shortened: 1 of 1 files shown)


iFreilicht:
Huh, indeed. I guess this is a bug in nixd then.

iFreilicht:
Alright, I removed that commit again, should be good now.

Lassulus:
but removing `with lib;` would be nice as a separate task :)

PR #836, opened 2024-10-16
lib: Remove global `with lib;`

iFreilicht:
This is generally regarded as a code-smell. I did utilize `with lib;` in some places where I felt it aided readability, but it's very clearly scoped in those situations.

See https://github.com/nix-community/disko/pull/835#issuecomment-2416126497


Mic92:
@mergify queue

PR #841, opened 2024-10-18
Fix whitespace handling

### LINKED ISSUE #130: Spaces in GPT partition names not properly escaped and general question on escaping for generated scripts

iFreilicht:
Fixes #130
Closes #330 

This should fix pretty much all cases where spaces or other special
characters would break disko due to improper quoting. I searched for all
instances of '.label', '.device' and '.name', so I believe I caught
whatever I could.

In some cases I changed single quotes to double quotes for consistency.

I know we don't usually fix bugs in the legacy table type, but it was so
easy I couldn't resist.

### DIFF
--- a/lib/interactive-vm.nix
+++ b/lib/interactive-vm.nix
@@ -20,7 +20,7 @@ let
   disks = lib.attrValues cfg_.disko.devices.disk;
   rootDisk = {
     name = "root";
-    file = ''"$tmp"/${(builtins.head disks).name}.qcow2'';
+    file = ''"$tmp/${(builtins.head disks).name}.qcow2"'';
     driveExtraOpts.cache = "writeback";
     driveExtraOpts.werror = "report";
     deviceExtraOpts.bootindex = "1";
@@ -29,7 +29,7 @@ let
   otherDisks = map
     (disk: {
       name = disk.name;
-      file = ''"$tmp"/${disk.name}.qcow2'';
+      file = ''"$tmp/${disk.name}.qcow2"'';
       driveExtraOpts.werror = "report";
     })
     (builtins.tail disks);
@@ -72,8 +72,8 @@ in
     trap 'rm -rf "$tmp"' EXIT
     ${lib.concatMapStringsSep "\n" (disk: ''
       ${hostPkgs.qemu}/bin/qemu-img create -f qcow2 \
-      -b ${config.system.build.diskoImages}/${disk.name}.qcow2 \
-      -F qcow2 "$tmp"/${disk.name}.qcow2
+      -b "${config.system.build.diskoImages}/${disk.name}.qcow2" \
+      -F qcow2 "$tmp/${disk.name}.qcow2"
     '') disks}
     set +f
     ${config.system.build.vm}/bin/run-*-vm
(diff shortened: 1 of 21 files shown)


lib/interactive-vm.nix
```diff
@@ -20,7 +20,7 @@ let
   disks = lib.attrValues cfg_.disko.devices.disk;
   rootDisk = {
     name = "root";
-    file = ''"$tmp"/${(builtins.head disks).name}.qcow2'';
+    file = ''"$tmp/${(builtins.head disks).name}.qcow2"'';
```
Enzime, line 23: Maybe `lib.escapeShellArg` could work better here?

Mic92: I also agree. Also we should keep shell escaping very close to the command so we don't accidentally double quote.

iFreilicht: Just for my understanding, do you mean in general or only in these concatenated instances?

It seems like it would just be the better choice in general, though it's a bit verbose unfortunately.

Enzime:
```suggestion
    file = ''"$tmp"/${lib.escapeShellArg ((builtins.head disks).name)}.qcow2'';
```

This is how I would do it, maybe move the `(builtins.head disk).name` into the `let` above as well if it seems a bit dense

iFreilicht: Okay then I think I understood correctly. I updated the changes, do they look better now?

Mic92 (approved):

Mic92:
@mergify queue

Issue #725, opened 2024-08-03
feat: prompt before formatting


crabdancing:
`disko` is dangerous. In particular, it will happily destroy any disk if you accidentally set the mode `disko` instead of `mount`. I've ended up making this mistake twice, and having to copy files again to correct it. For someone without a backup, the effects could be even worse.

Making `disko` ask for confirmation by default, with a `-y` flag to force it to continue without prompting, would be ideal for avoiding most mistakes.

geekodour:
What is the behavior if I add a new disk, eg. new EBS volume. We need to run with `mode=disko` on the new EBS volume but want to keep the older disk to be intact and untouched. Is there a way to make that happen? or you would just comment out the previous disk block before running with mode=disko. 

I've to test this out. I was just about to do this in a production system but think i'll re-think.

Mic92:
Disko can run incrementally, I used this to create new zfs datasets. However I would test if it works for your use case on a system where you don't care about the data, I found a bug in the process where it deleted lost a zpool. Meaning there might be many edge cases we have not discovered yet.

sedlund:
It would be interesting to contemplate implementing the disko primitives in a terraform provider. Then you gain the tf planning mechanism, state, and other niceties.

iFreilicht:
> Disko can run incrementally

Yes, with `--mode format`. But that mode isn't very dangerous anyway, it will never delete anything. The issue mentions `--mode disko`, which wipes everything without asking by default.

> Making disko ask for confirmation by default, with a -y flag to force it to continue without prompting, would be ideal for avoiding most mistakes.

I am very much in favor of this.

One note for implementation: [nixos-anywhere runs `diskoScript` directly](https://github.com/nix-community/nixos-anywhere/blob/d5ced5dc30c7a14587f002a1ab67f5201911f03d/src/nixos-anywhere.sh#L466). As nixos-anywhere does **not** lock the version of disko it uses (which comes from the nixosConfiguration), we need to be careful about potential compatibility issues. A simple solution would be to put this check into the CLI, not into the `diskoScript`.

adam248:
quick overview of #772 (closed)

4 implementation ideas:

add one or more of the following to the CLI.

1. add a `--confirm` flag to the CLI requring a `yes | no | abort` for each disk. *This is backwards compatible.*
2. add a `--yes` flag to the CLI to force auto-confirm. This would mean changing the default to prompt per disk. ***This is a breaking change.***
[...]

Issue #454, opened 2023-11-24
`disko-create` and `disko-mount` scripts not created


zmitchell:
I have a flake that looks like this:
```
{
  description = "A very basic flake";
  inputs.nixpkgs.url = "github:NixOS/nixpkgs/release-23.05";
  inputs.flake-utils.url = "github:numtide/flake-utils";
  inputs.nixos-generators.url = "github:nix-community/nixos-generators";
  inputs.nixos-generators.inputs.nixpkgs.follows = "nixpkgs";
  inputs.nixos-hardware.url = "github:NixOS/nixos-hardware";
  inputs.disko.url = "github:nix-community/disko";
  inputs.disko.inputs.nixpkgs.follows = "nixpkgs";

  outputs = inputs @ { self, nixpkgs, flake-utils, nixos-generators, disko, ... }: 
  let 
    common = {
      modules = [
        ./stacks/users.nix
        ./stacks/system.nix
      ];
      specialArgs = { inherit inputs; };
    };
    chonkerVmConfig = {
      system = "aarch64-linux";
      specialArgs = { pkgs = nixpkgs.legacyPackages.aarch64-linux; };
      modules = [
        disko.nixosModules.disko
        nixpkgs.legacyPackages.aarch64-linux.lib.callPackage ./stacks/zfs_single_drive.nix {
          device = "/dev/vda";
          user = "zmitchell";
        }
      ];
    } // common;
    behemothConfig = {
      system = "x86_64-linux";
      modules = [
        ./stacks/behemoth.nix
        nixpkgs.legacyPackages.x86_64-linux.lib.callPackage ./stacks/zfs_single_drive.nix {
          device = "/dev/sda";
          user = "zmitchell";
        }
      ];
      specialArgs = { pkgs = nixpkgs.legacyPackages.x86_64-linux; };
    } // common;
  in
  {
    nixosModules = {
      chonker-vm = chonkerVmConfig;
      behemoth = behemothConfig;
    };
    nixosConfigurations = {
      chonker-vm = nixpkgs.lib.nixosSystem self.nixosModules.chonker-vm // { disko.enableConfig = false; };
      behemoth = nixpkgs.lib.nixosSystem self.nixosModules.behemoth;
    };
  }
  //
  flake-utils.lib.eachDefaultSystem (system:
    {
      packages.chonkerVmInstaller = nixos-generators.nixosGenerate ({
        format = "iso";
      } // self.nixosModules.chonker-vm);
[...]

PR #842, opened 2024-10-18
outputs: make compatible with nix run and package lists

iFreilicht:
This adds new outpus like `diskoFormat` and `diskoFormatNoDeps` which are compatible with `nix run` so you can do something like

    nix run .#nixosConfigurations.myhostname.config.system.build.diskoFormatNoDeps

as originally intended in #78, or add disko to your configuration like

    environment.systemPackages = [
        config.system.build.diskoFormat
        config.system.build.diskoMount
        config.system.build.diskoDisko
    ];

as mentioned in #454.

The old outputs are marked as deprecated and a warning like this is shown to the user:

```console
# nix run .#nixosConfigurations.testmachine.config.system.build.mountScript
trace: evaluation warning: 
The output .mountScript is deprecated in favor of .diskoMount and will be removed in
version 2.0.0! To fix this issue, you need to do the following:
  - If you're currently calling "$‍{system.build.mountScript}" as a script,
    use "$‍{system.build.diskoMount}/bin/disko-mount" instead!
  - If you added "system.build.mountScript" to your systemPackages, users.<name>.packages
    or home.packages, or used it with buildEnv or mkShell, this has never worked
    and using "system.build.diskoMount" will fix it.

error: unable to execute '/nix/store/c48m5ffrfzhgsri19b4mlzlbv31ba8yg-disko-mount/bin/disko-mount': Not a directory
```

Fixes part of #454 (which then only requires proper documentation)
Supersedes #78

@Mic92 I know you suggested in https://github.com/nix-community/disko/pull/78#discussion_r1769601172 to use `diskoRecreate`, but I think as long as we still have `--mode disko`, we should keep it consistent, so I went for `diskoDisko` instead. This also helps differentiate this output from the regular `disko`, which is the CLI.

### DIFF
--- a/disko
+++ b/disko
@@ -25,10 +25,12 @@ or else from the disko module of a NixOS configuration of that name under .nixos
 Options:
 
 * -m, --mode mode
-  set the mode, either format, mount or disko
-    format: create partition tables, zpools, lvms, raids and filesystems
-    mount: mount the partition at the specified root-mountpoint
-    disko: first unmount and destroy all filesystems on the disks we want to format, then run the create and mount mode
+  set the mode, either distroy, format, mount, format,mount or destroy,format,mount
+    destroy: unmount filesystems and destroy partition tables of the selected disks
+    format: create partition tables, zpools, lvms, raids and filesystems if they don't exist yet
+    mount: mount the partitions at the specified root-mountpoint
+    format,mount: run format and mount in sequence
+    destroy,format,mount: run all three modes in sequence. Previously known as --mode disko
 * -f, --flake uri
   fetch the disko config relative to this flake's root
 * --arg name value
@@ -126,8 +128,16 @@ nixBuild() {
   fi
 }
 
-if ! { [[ $mode = "format" ]] || [[ $mode = "mount" ]] || [[ $mode = "disko" ]] || [[ $mode = "create" ]] || [[ $mode = "zap_create_mount" ]] ; }; then
-  abort "mode must be either format, mount or disko"
+if ! {
+  # Base modes
+  [[ $mode = "destroy" ]] || [[ $mode = "format" ]] || [[ $mode = "mount" ]] ||
+  # Combined modes
+  [[ $mode = "format,mount" ]] ||
+  [[ $mode = "destroy,format,mount" ]] || # Replaces --mode disko
+  # Legacy modes, will be removed in next major version
+  [[ $mode = "disko" ]] || [[ $mode = "create" ]] || [[ $mode = "zap_create_mount" ]] ;
+}; then
+  abort 'mode must be one of "destroy", "format", "mount", "destroy,format,mount" or "format,mount"'
 fi
 
 if [[ -n "${flake+x}" ]]; then
--- a/disko
+++ b/disko
@@ -9,6 +9,7 @@ declare disko_config
 # mount was chosen as the default mode because it's less destructive
 mode=mount
 nix_args=()
+skip_destroy_safety_check=false
 
 # DISKO_VERSION is set by the wrapper in package.nix
 DISKO_VERSION="${DISKO_VERSION:="unknown! This is a bug, please report it!"}"
@@ -44,6 +45,8 @@ Options:
 * --no-deps
   avoid adding another dependency closure to an in-memory installer
     requires all necessary dependencies to be available in the environment
+* --yes-wipe-all-disks
+  skip the safety check for destroying partitions, useful for automation
 * --debug
   run with set -x
 * --help
@@ -94,6 +97,9 @@ while [[ $# -gt 0 ]]; do
       ;;
     --no-deps)
       nix_args+=(--arg noDeps true)
+      ;;
+    --yes-wipe-all-disks)
+      skip_destroy_safety_check=true
       ;;
     --show-trace)
       nix_args+=("$1")
@@ -169,8 +175,14 @@ script=$(nixBuild "${libexec_dir}"/cli.nix \
   --argstr mode "$mode" \
   "${nix_args[@]}"
 )
+
+command=("$(echo "$script"/bin/*)")
+if [[ $mode = "destroy,format,mount" && $skip_destroy_safety_check = true ]]; then
+  command+=("--yes-wipe-all-disks")
+fi
+
 if [[ -n "${dry_run+x}" ]]; then
-  echo "$script"
+  echo "${command[@]}"
 else
-  exec "$script"
+  exec "${command[@]}"
 fi
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -462,7 +462,7 @@ let
                 '';
                 destroyFormatMount = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-destroy-format-mount" ''
                   export PATH=${lib.makeBinPath ((cfg.config._packages pkgs) ++ [ pkgs.bash ] ++ destroyDependencies)}:$PATH
-                  ${cfg.config._disko}
+                  ${cfg.config._destroyFormatMount}
                 '';
 
                 # These are useful to skip copying executables uploading a script to an in-memory installer
@@ -479,7 +479,7 @@ let
                   ${cfg.config._formatMount}
                 '';
                 destroyFormatMountNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "/bin/disko-destroy-format-mount" ''
-                  ${cfg.config._disko}
+                  ${cfg.config._destroyFormatMount}
                 '';
 
 
@@ -490,7 +490,7 @@ let
                 # `user.users.<name>.packages` and `home.packages`, see https://github.com/nix-community/disko/issues/454
                 destroyScript = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "disko-destroy" ''
                   export PATH=${lib.makeBinPath destroyDependencies}:$PATH
-                  ${cfg.config._destroy}
+                  ${cfg.config._legacyDestroy}
                 '';
 
                 formatScript = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "disko-format" ''
@@ -510,7 +510,7 @@ let
 
                 # These are useful to skip copying executables uploading a script to an in-memory installer
                 destroyScriptNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "disko-destroy" ''
-                  ${cfg.config._destroy}
+                  ${cfg.config._legacyDestroy}
                 '';
 
                 formatScriptNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "disko-format" ''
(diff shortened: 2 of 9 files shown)


lib/tests.nix
```diff
@@ -260,24 +260,24 @@ let
...
+            machine.succeed("${tsp-format}/bin/disko-format") # verify that format is idempotent
             machine.succeed("test -e /mnt/home/testfile")
           ''}
           ${lib.optionalString (testMode == "module") ''
             #  running module mode
-            machine.succeed("${nodes.machine.system.build.formatScript}")
-            machine.succeed("${nodes.machine.system.build.mountScript}")
-            machine.succeed("${nodes.machine.system.build.mountScript}") # verify that mount is idempotent
-            machine.succeed("${nodes.machine.system.build.diskoScript}") # verify that we can destroy and recreate again
+            machine.succeed("${nodes.machine.system.build.diskoFormat}/bin/disko-format")
```
Enzime, line 274:
```suggestion
            machine.succeed("${lib.getExe nodes.machine.system.build.diskoFormat}")
```

Mic92:
Well we can also call it recreate in the cli. `diskoDisko` looks weird and the mode disko was never really descriptive to begin with.

Mic92:
I also would prefer to not have the warning in the beginning until we have also migrated the nixos-anywhere executable to it. We can add the warning later when also nixos-stable no longer has the old nixos-anywhere version to avoid showing users warnings they cannot silence.

Enzime:
Would love to see a new mode or maybe a flag for the disko/recreate script that doesn’t wipe the partition table as create is kinda idempotent now, I think this mode would be quite useful with `nixos-anywhere` as a create/update + mount

Mic92:
@Enzime that mode already exists afaik and it's called `format`.

Enzime:
I thought `format` doesn't mount the filesystems afterwards

iFreilicht:
> I also would prefer to not have the warning in the beginning until we have also migrated the nixos-anywhere executable to it. We can add the warning later when also nixos-stable no longer has the old nixos-anywhere version to avoid showing users warnings they cannot silence.

That's a good point, I'll split those into a separate PR.

> Well we can also call it recreate in the cli, while still allowing user the old name. diskoDisko looks weird and the mode disko was never really descriptive to begin with.

That's a good point, and I agree, we should do that. I think I would like to take this opportunity to integrate a solution for #725, which is to simply display a warning when running the "recreate" mode that all disks will be wiped, and asking for confirmation. This dialogue can be skipped with the new flag `--force-wipe-disks`, and will only appear for `--mode recreate`, not `--mode disko`.

Would you agree with that change?

> Would love to see a new mode or maybe a flag for the disko/recreate script that doesn’t wipe the partition table as create is kinda idempotent now, I think this mode would be quite useful with nixos-anywhere as a create/update + mount

Hmm, this makes me feel like we might want to make the names of our modes more direct. Instead of inventing new names for joint modes, why don't we just leave the base modes named `destroy`, `format` and `mount`, and allow passing the following values to `--mode`:

- `destroy,format,mount` - what `disko` currently does
- `format,mount` - the new mode @Enzime suggested
- `format` - unchanged
- `mount` - unchanged

The outputs would then be:

- `diskoDestroyFormatMountScript`
- `diskoFormatMountScript`
- `diskoFormatScript`
- `diskoMountScript`
- `diskoDestroyFormatMountScriptNoDeps`
- `diskoFormatMountScriptNoDeps`
- `diskoFormatScriptNoDeps`
- `diskoMountScriptNoDeps`

This isn't pretty, but it sure as hell is clear.

lib/default.nix
```diff
@@ -403,42 +403,101 @@ let
...
+                      and using "system.build.${newAttr}" will fix it.
+                '';
               in
               lib.mapAttrs throwIfNoDisksDetected {
-                destroyScript = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "disko-destroy" ''
+                diskoDestroy = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-destroy" ''
                   export PATH=${lib.makeBinPath destroyDependencies}:$PATH
                   ${cfg.config._destroy}
                 '';
+                diskoFormat = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-format" ''
```
Mic92, line 425:
```suggestion
                format = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-format" ''
```

lib/default.nix
```diff
@@ -403,42 +403,101 @@ let
...
+                  export PATH=${lib.makeBinPath (cfg.config._packages pkgs)}:$PATH
+                  ${cfg.config._create}
+                '';
+                diskoMount = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-mount" ''
```
Mic92, line 429:
```suggestion
                mount = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-mount" ''
```

lib/default.nix
```diff
@@ -403,42 +403,101 @@ let
...
+                  export PATH=${lib.makeBinPath (cfg.config._packages pkgs)}:$PATH
+                  ${cfg.config._mount}
+                '';
+                diskoDisko = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-disko" ''
```
Mic92, line 433:
```suggestion
                disko = (diskoLib.writeCheckedBash { inherit pkgs checked; }) "/bin/disko-disko" ''
```

lib/default.nix
Mic92: This is also less to type considering this is used in `nix run`.

Mic92: Also this separates the implementation from disko making it more of an interface with  potential other implementations.

lib/default.nix
iFreilicht: This just side-steps the issue that `disko` is a bad name for this mode.

I agree that we need a better name and that `diskoDisko` is awkward (though also funny), but just going back to `disko` is not helpful. The only thing that should be called `disko` without a suffix is the CLI itself.

lib/default.nix
iFreilicht: Okay, that seems reasonable to me 👍

lib/default.nix
iFreilicht: I don't think that's a good idea. We can call the output `destroyFormatMount`, but just calling it `disko` makes it too confusing what it's supposed to be, and goes against your other idea of making the outputs a generic interface that has nothing to do with disko directly.

The only output I would expect to be called `disko` is the CLI.

lib/default.nix
iFreilicht: Okay so I updated the PR now to use the shorter names, and solved the problem in `default.nix` by using a `_cli` prefix for the new outputs that `cli.nix` uses.

I also added deprecation warnings to all of these outputs. I believe they are not used by most users, see my commit message.

lib/default.nix
```diff
@@ -403,42 +403,101 @@ let
...
+                diskoDestroyNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "/bin/disko-destroy" ''
+                  ${cfg.config._destroy}
+                '';
+                diskoFormatNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "/bin/disko-format" ''
+                  ${cfg.config._create}
+                '';
+                diskoMountNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "/bin/disko-mount" ''
+                  ${cfg.config._mount}
+                '';
+                diskoDiskoNoDeps = (diskoLib.writeCheckedBash { inherit pkgs checked; noDeps = true; }) "/bin/disko-disko" ''
```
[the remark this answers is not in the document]
iFreilicht: Done!

lib/default.nix
iFreilicht: Done.

iFreilicht:
As I hinted at yesterday, I now used the opportunity to integrate a confirmation dialogue into the modes `destroy` and `destroy,format,mount`, which fixes #725.

The existing `disko` mode is not affected.

I can also split this into a separate PR, but I wanted to avoid having a state on master that is forwards-incompatible with this.

iFreilicht:
I'll fix the build errors soon, gotta go to bed now, but would appreciate feedback on the changes.

Mic92:
Sorry, currently busy with NixCon preparation. I would like to get back to this later.

Lassulus:
changes look good to me, I haven't tested everything though, but it seems like there are no breaking changes and CI is happy.
What would be interesting (in the future in a new PR) would be a real unmount mode, where we don't destroy disks but just umount -Rl and deactivate the LVM/mdadm stuff for affected devices, something similiar to disk-deactivate but less destructive :)

disko
```diff
@@ -159,8 +175,15 @@ script=$(nixBuild "${libexec_dir}"/cli.nix \
   --argstr mode "$mode" \
   "${nix_args[@]}"
 )
+
+command="$(echo "$script"/bin/*)"
```
Mic92, line 179: Does this break using a new version of the cli with an old version of disko?
Ideally we would fallback or warn the user in this case.
Maybe this could become an array?

```suggestion
command=("$(echo "$script"/bin/*")
```

disko
```diff
@@ -159,8 +175,15 @@ script=$(nixBuild "${libexec_dir}"/cli.nix \
...
+if [[ $mode = "destroy,format,mount" && $skip_destroy_safety_check = true ]]; then
+  command+=' --yes-wipe-all-disks'
```
Mic92, line 181:
```suggestion
  command+=("--yes-wipe-all-disks")
```

disko
```diff
@@ -159,8 +175,15 @@ script=$(nixBuild "${libexec_dir}"/cli.nix \
...
+fi
+
 if [[ -n "${dry_run+x}" ]]; then
-  echo "$script"
+  echo "$command"
```
Mic92, line 185:
```suggestion
  echo "${command[@]}"
```

disko
```diff
@@ -159,8 +175,15 @@ script=$(nixBuild "${libexec_dir}"/cli.nix \
...
 else
-  exec "$script"
+  # Don't use quotes to allow passing arguments to the script. We know the path does not contain spaces.
+  exec $command
```
Mic92, line 188:
```suggestion
  exec "${command[@]}"
```

disko
Mic92: Ah wait. The CLI will always bring its own version of the disko modules anway. Than we don't need to care about compat.

iFreilicht: Ah yes, that is much better, thank you!

disko
iFreilicht: Done.

disko
iFreilicht: Done.

disko
iFreilicht: Done.

iFreilicht:
Alright I addressed the last few comments. Once this is merged, I'll open a new issue regarding how we handle the deprecation etc. to make sure an updated version of nixos-anywhere is released before then.

lib/tests.nix
iFreilicht: Done!

Mic92 (approved):

PR #847, opened 2024-10-25
UX: be specific if disko-install failed or succeeded

DavHau:
I was running into a problem where disko install failed but the logs didn't indicate that.

### DIFF
--- a/disko-install
+++ b/disko-install
@@ -244,4 +244,10 @@ main() {
   nixos-install --no-channel-copy --no-root-password --system "$nixos_system" --root "$mountPoint"
 }
 
-main "$@"
+if main "$@"; then
+  echo "disko-install succeeded"
+  exit 0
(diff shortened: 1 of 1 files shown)


disko-install
```diff
@@ -244,4 +244,10 @@ main() {
   nixos-install --no-channel-copy --no-root-password --system "$nixos_system" --root "$mountPoint"
 }
 
-main "$@"
+if main "$@"; then
+  echo "disko-install succeeded" >&2
```
iFreilicht, line 248: Shouldn't this go to stdout?

Mic92:
```suggestion
  echo "disko-install succeeded"
```

iFreilicht:
@mergify squash

iFreilicht:
@mergify queue

PR #848, opened 2024-10-25
lvm_vg: fix size=100% leading to crash

DavHau:
lvcreate -l does not accept a '100%' parameter which currently leads to a crash. THis change automatically changes `100%` to `100%FREE` leading to the intended behavior.

### DIFF
--- a/example/luks-lvm.nix
+++ b/example/luks-lvm.nix
@@ -45,7 +45,7 @@
         type = "lvm_vg";
         lvs = {
           root = {
-            size = "100M";
+            size = "100%";
             content = {
               type = "filesystem";
               format = "ext4";
--- a/lib/types/lvm_vg.nix
+++ b/lib/types/lvm_vg.nix
@@ -95,7 +95,7 @@ in
                 --yes \
                 ${if (lv.lvm_type == "thinlv") then "-V"
                   else if lib.hasInfix "%" lv.size then "-l" else "-L"} \
-                  ${lv.size} \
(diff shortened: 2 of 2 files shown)


phaer (dismissed):
This feels a bit too magic from my perspective; I'd prefer failing loudly on non-supported values instead of preprocessing them like this.

While `%FREE` is probably the most commonly used suffix, it's not the same as "%" in other backends semantically. There's also %VG, %PVS and %ORIGIN.

Maybe it's good enough to just update the example to use `100%FREE`?

iFreilicht:
@phaer actually, sgdisk doesn't support 100% either. It uses `-0`, and [we actively convert from `100%` as a special value](https://github.com/nix-community/disko/blob/09a776702b004fdf9c41a024e1299d575ee18a7d/lib/types/gpt.nix#L105).

I do somewhat agree with your sentiment, the correct solution here is [to fix this todo](https://github.com/nix-community/disko/blob/09a776702b004fdf9c41a024e1299d575ee18a7d/lib/types/lvm_vg.nix#L42), but I think in the context of disko, being able to just use `100%` for `size` everywhere is a nice, user-friendly abstraction for the most common usecase, so I am in support of this change.

I would like at least one other maintainer to agree on this, I don't wanna just merge something like this without some sort of consensus.

phaer:
Thanks for noticing that @iFreilicht. Still not the biggest fan in terms of UX here, but I guess it's consistent then at least. So Lgtm!

iFreilicht (approved):
Alright, great!

iFreilicht:
@mergify queue

PR #851, opened 2024-10-27
disk-image: Fix "unexpected argument 'customQemu'" eval error

### LINKED ISSUE #850: Building interactive VM fails with called with unexpected argument 'customQemu'
When I try to build the interactive VM with `nix run -L '.#nixosConfigurations.test-raspi.config.system.build.vmWithDisko'`

It fails with the following error. It looks like the `customQemu` argument isn't present in Nixpkgs stable 24.05.
This issue here might me related as well: https://github.com/nix-community/disko/issues/758

```
       error: function 'anonymous lambda' called with unexpected argument 'customQemu'

       at /nix/store/lbqj1cndic4121whnx8xm0jgb1c8x4xx-source/pkgs/build-support/vm/default.nix:1:1:

```

iFreilicht:
Previously, an error like

    error: function 'anonymous lambda' called with unexpected argument 'customQemu'

    at /nix/store/lbqj1cndic4121whnx8xm0jgb1c8x4xx-source/pkgs/build-support/vm/default.nix:1:1:

was printed when trying to evaluate `config.system.build.vmWithDisko` or `config.system.build.diskoImagesScript` with nixpkgs 24.05 (and below).

Fixes #850

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -105,12 +105,13 @@ let
       hexEscapeUdevSymlink "all0these@char#acters+_are-allow.ed"
       => "all0these@char#acters+_are-allow.ed"
     */
-    hexEscapeUdevSymlink = let
-      allowedChars = "[0-9A-Za-z#+-.:=@_/]";
-      charToHex = c: lib.toHexString (lib.strings.charToInt c);
-    in
-    lib.stringAsChars
-      (c: if lib.match allowedChars c != null || c == "" then c else "\\x" + charToHex c);
+    hexEscapeUdevSymlink =
+      let
+        allowedChars = "[0-9A-Za-z#+-.:=@_/]";
+        charToHex = c: lib.toHexString (lib.strings.charToInt c);
+      in
+      lib.stringAsChars
+        (c: if lib.match allowedChars c != null || c == "" then c else "\\x" + charToHex c);
 
     /* get the index an item in a list
 
@@ -333,6 +334,14 @@ let
        packages :: lib.types.devices -> pkgs -> [ derivation ]
     */
     packages = toplevel: toplevel._packages;
+
+    /* Checks whether nixpkgs is recent enough for vmTools to support the customQemu argument.
+
+      Returns false, which is technically incorrect, for a few commits on 2024-07-08, but we can't be more accurate.
+
+      vmToolsSupportsCustomQemu :: pkgs -> bool
+    */
+    vmToolsSupportsCustomQemu = pkgs: lib.versionAtLeast pkgs.lib.version "24.11.20240709";
 
     optionTypes = rec {
       filename = lib.mkOptionType {
--- a/module.nix
+++ b/module.nix
@@ -194,6 +194,17 @@ in
   };
 
   config = {
+    assertions = [
+      {
+        assertion = config.disko.imageBuilder.qemu != null -> diskoLib.vmToolsSupportsCustomQemu pkgs;
+        message = ''
+          You have set config.disko.imageBuild.qemu, but vmTools in your nixpkgs version "${lib.version}"
+          does not support overriding the qemu package with the customQemu option yet.
+          Please upgrade nixpkgs so that `lib.version` is at least "24.11.20240709".
+        '';
+      }
+    ];
+
     _module.args.diskoLib = import ./lib {
       inherit lib;
       rootMountPoint = config.disko.rootMountPoint;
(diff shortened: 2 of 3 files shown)


lib/default.nix
```diff
@@ -334,6 +334,20 @@ let
...
+      technically the minimum version is 24.11.20240708.65c851c. However, versionAtLeast only compares versions
+      alphabetically, so the comparison's result will be incorrect for other commits made on the same day.
+
+      Instead, we compare against 24.11.20240709, which is the next day. This means this function returns false for
+      some commits that DO support the customQemu argument, but if it returns true, we can be 100% certain that this
+      is correct, and we can pass the customQemu argument to vmTools without an evaluation error.
+
+      vmToolsSupportsCustomQemu :: pkgs -> bool
+    */
+    vmToolsSupportsCustomQemu = pkgs: lib.versionAtLeast pkgs.lib.version "24.11.20240709";
```
phaer, line 349: I think it's good enough to assume people are either on a release channel or on the *latest* unstable one, so no need to check for the date IMO.

And while I really appreciate the detailed description, I do think it's a bit verbose for an inline comment and the commit message might be a better place to explain how it came to be?

module.nix
```diff
@@ -194,6 +194,18 @@ in
...
   config = {
+    assertions = [
+      {
+        assertion = config.disko.imageBuilder.qemu != null -> diskoLib.vmToolsSupportsCustomQemu pkgs;
+        message = ''
+          You have set config.disko.imageBuild.qemu, but vmTools in your nixpkgs version "${lib.version}"
+          does not support overriding the qemu package with the customQemu option yet, which was added in
+          https://github.com/NixOS/nixpkgs/commit/65c851cd7523c669b8fb25236b1c48283a2f43ec
+          Please upgrade nixpkgs so that `lib.version` is at least "24.11.20240709".
+        '';
```
phaer, line 205: Same here, I think if people need to find the exact commit it was introduced in, the files history makes it easy to do so. For an error message it's good enough to know that the current nixpkgs is too old

lib/default.nix
iFreilicht: > I think it's good enough to assume people are either on a release channel or on the latest unstable one, so no need to check for the date IMO.

I disagree. The entire nixpkgs revision history lasts forever. People might use outdated versions for example to bisect compatibility issues upstream, doing research or to get a specific version of a package. The `qemu` option is simply incompatible with a certain range of nixpkgs. I found what that range was, and this function encapsulates that knowledge as good as is feasible. I don't think I should actively write a more incorrect implementation of this check. What would be the benefit of that?

> And while I really appreciate the detailed description 

Thank you :)

>  I do think it's a bit verbose for an inline comment and the commit message might be a better place to explain how it came to be?

Why? The contribution experience for developers depends on good documentation. We're not generating formatted docs from these comments yet, but they are de-facto documentation and IMO should be of the quality that we would expect from documentation for developers.

Again, I don't understand what would be the benefit of shortening this comment? If you can think of a more terse formulation, I'm all for it, but the fact that this function even exists bears an explanation, and if I saw something like this, I would appreciate it being documented.

module.nix
iFreilicht: Okay I agree the exact commit is not really needed for the user-facing error. I removed that part now :)

lib/default.nix
phaer: > I disagree. The entire nixpkgs revision history lasts forever. People might use outdated versions for example to bisect compatibility issues upstream, doing research or to get a specific version of a package. The qemu option is simply incompatible with a certain range of nixpkgs.

Sure, but the current state of things is that diskos master is broken for the latest official release (without this PR) and we didn't catch that in CI or otherwise until @Nebucatnetzer ran into it at NixCon. So I don't think supporting each and every nixpkgs commit is a realistic goal and therefore would rather focus on actually supporting both -unstable and the latest stable release.
No big deal to keep the date in, I don't care too much about it as long as it's just a single check and not an if-else-tree of date-specific clauses ;)

> Again, I don't understand what would be the benefit of shortening this comment? 

I am not arguing against documentation :D the first and last line of that comment are totally fine. The 2 paragraphs in between would IMO be better placed in the commit message, because the do provide *historical* context. Comments should tell me what that function does and possible pitfalls. I think its good to keep documentation somewhat consistent and don't think we document the genesis of other values anywhere? (And neither do I think we should).

But again, no strong feelings. I'd rather have VMs on diskos master branch work again with nixos stable sooner than later and don't want to block anything here.

iFreilicht: > until @Nebucatnetzer ran into it at NixCon

oh no, was it during a live demo? 😬

Nebucatnetzer:
I used the flake.On Oct 28, 2024, at 17:00, Felix Uhl ***@***.***> wrote:﻿
@iFreilicht commented on this pull request.



In lib/default.nix:
> @@ -334,6 +334,20 @@ let
     */
     packages = toplevel: toplevel._packages;
 
+    /* Checks whether nixpkgs is recent enough for vmTools to support the customQemu argument.
+
+      This feature was added in https://github.com/NixOS/nixpkgs/commit/65c851cd7523c669b8fb25236b1c48283a2f43ec, so
+      technically the minimum version is 24.11.20240708.65c851c. However, versionAtLeast only compares versions
+      alphabetically, so the comparison's result will be incorrect for other commits made on the same day.
+
+      Instead, we compare against 24.11.20240709, which is the next day. This means this function returns false for
+      some commits that DO support the customQemu argument, but if it returns true, we can be 100% certain that this
+      is correct, and we can pass the customQemu argument to vmTools without an evaluation error.
+
+      vmToolsSupportsCustomQemu :: pkgs -> bool
+    */
+    vmToolsSupportsCustomQemu = pkgs: lib.versionAtLeast pkgs.lib.version "24.11.20240709";

@Nebucatnetzer just to be sure, did you use the latest disko from this flake, or whatever was packaged in nixpkgs 24.05? I'm preparing a release right now to make the fix public and will create PRs for nixpkgs as well, and was wondering whether this should be backported to 24.05.

PR #852, opened 2024-10-27
treewide: Fix using pkgs.{build,host}Platform alias

amarshall:
This was recently changed in nixpkgs to be an alias; if nixpkgs is configured to disable aliases, this usage will fail.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -39,7 +39,7 @@
           default = self.packages.${system}.disko;
 
           create-release = pkgs.callPackage ./scripts/create-release.nix { };
-        } // pkgs.lib.optionalAttrs (!pkgs.buildPlatform.isRiscV64) {
+        } // pkgs.lib.optionalAttrs (!pkgs.stdenv.buildPlatform.isRiscV64) {
           disko-doc = pkgs.callPackage ./doc.nix { };
         });
       # TODO: disable bios-related tests on aarch64...
@@ -48,7 +48,7 @@
         let
           pkgs = nixpkgs.legacyPackages.${system};
           # FIXME: aarch64-linux seems to hang on boot
-          nixosTests = lib.optionalAttrs pkgs.hostPlatform.isx86_64 (import ./tests {
+          nixosTests = lib.optionalAttrs pkgs.stdenv.hostPlatform.isx86_64 (import ./tests {
             inherit pkgs;
             makeTest = import (pkgs.path + "/nixos/tests/make-test-python.nix");
             eval-config = import (pkgs.path + "/nixos/lib/eval-config.nix");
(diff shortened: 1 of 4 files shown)


iFreilicht:
I need a bit more context for this change. [The docs for `allowAliases`](https://nixos.org/manual/nixpkgs/unstable/#opt-allowAliases) say:

> Projects that aren’t Nixpkgs should be cautious of instantly removing all usages of aliases, as migrating too soon can break compatibility with the stable Nixpkgs releases.

As I understand, `pkgs.stdenv.hostPlatform` and `pkgs.stdenv.buildPlatform` already existed for a long time, and so this change will not break compatibility with old releases, right?

> This was recently changed in nixpkgs to be an alias

Could you maybe link me to the PR or discussion where this change was made? I tried to find it in the git log or just full-text search, but I've now given up after 15 minutes of searching.

iFreilicht:
Ah, also, regarding this:

> if nixpkgs is configured to disable aliases, this usage will fail.

Could you show me how I can reproduce this issue? My attempt was the following:

```
nix-repl> pkgs = (import (builtins.getFlake "github:NixOS/nixpkgs/nixpkgs-unstable")  { config.allowAliases = false; })

nix-repl> disko = (import ./. { lib = pkgs.lib; })

nix-repl> disko.format (import example/simple-efi.nix)                                                           
"set -efux\n\ndisko_devices_dir=$(mktemp -d)\ntrap 'rm -rf \"$disko_devices_dir\"' EXIT\nmkdir -p \"$disko_devices_dir\"\n\n( ... etc you get the idea
```

To me it seems like this should trigger the behavior, but maybe I'm missing something?

iFreilicht:
> Upstream commit is https://github.com/NixOS/nixpkgs/commit/e0464e47880a69896f0fb1810f00e0de469f770a.

Is it? I found that commit as well, but it replaces `stdenv.is*` with `stdenv.hostPackages.is*`. I searched through the entire diff, and there's no mention of `pkgs.hostPlatform` or `pkgs.buildPlatform`.

> disko’s flake doesn’t take a lib param, so that is probably a no-op.

`import ./.` doesn't do anything with the flake, it imports [`default.nix`](https://github.com/nix-community/disko/blob/89e458a3bb3693e769bfb2b2447c3fe72092d498/default.nix).

> It also needs to be the per-system pkgs probably, not one just used for lib.

From my understanding of our codebase, overriding the `lib` argument of `default.nix` should ensure that at least for the `create`, `format`, `mount` and `disko` outputs, there is no dependency on per-system pkgs, but maybe I'm missing something.

> I can maybe get back to tomorrow with a repro.

That would be great!

iFreilicht (approved):
Ohh, so the original change was already made in Jan, 2023, and the most recent commit just fixed that the old definitions were still available within `callPackage`! After seeing the code changes in nixpkgs, I don't need a repro, this change is clearly correct. Thank you!

iFreilicht:
@mergify queue

iFreilicht:
NOTE: Github shows a failed check on this PR. This is just because changes to the base branch were not allowed, so mergify queue failed. The relevant checks all passed, so this was fine to merge.

PR #863, opened 2024-11-08
docs: fix inconsistent disko-install cmd

raexera:
fix: 

```sh
error: flake 'github:nix-community/disko' does not provide attribute `apps.x56_64-linux.disko-install/latest`, `packages.x86_64-linux.disko-install/latest', 'legacyPackages.x86_64-linux.disko-install/latest' or 'disko-install/latest'
```

caused by the previous command


Enzime:
@mergify queue

PR #865, opened 2024-11-09
luks.nix | TPM implementation

Sk7Str1pe:
@iFreilicht first attempt of contribution XD. my very first work so i may made some(many) mistakes, waiting for feedback


Sk7Str1pe:
@Mic92 daammmit why build would fail?

PR #877, opened 2024-11-13
cli: Fix "attribute missing" error when using flake

### LINKED ISSUE #873: confusing error while building system configuration using disko as module
I've been working through enabling disko as a proper flake input which is then consumed as a module on all of my systems.  As I'm discovering, that process is not entirely as straightforward as I was expecting.

One example of that is where I'm trying to migrate an existing VirtualBox based VM over to using the previously created disko configuration I'd used originally to install it, but as a standalone file at the time of installation.  Now I'm trying to roll the disko configuration into the system configuration directly as a module, and I'm encountering a somewhat bizarre error:

```
       … while evaluating the attribute 'value'
         at /nix/store/0sxsfzswhjck7f4sbsvznz285s4i7y64-source/lib/modules.nix:614:53:
          613|                 (n: value:
          614|                   [{ inherit (module) file; inherit value; }]
             |                                                     ^
          615|                 )

       … while calling the 'throw' builtin
         at /nix/store/0sxsfzswhjck7f4sbsvznz285s4i7y64-source/nixos/modules/system/activation/top-level.nix:73:10:
           72|   baseSystemAssertWarn = if failedAssertions != []
           73|     then throw "\nFailed assertions:\n${concatStringsSep "\n" (map (x: "- ${x}") failedAssertions)}"
             |          ^
           74|     else showWarnings config.warnings baseSystem;

       error:
       Failed assertions:
       - You cannot have duplicated devices in mirroredBoots
```

The full error log can be found at https://gist.github.com/nipsy/61652eaf9705093536169f4765e8950c.

What's curious to me about this is I have two almost identical systems.  This one if still using legacy boot with the VM configured to support that (BIOS being the default Virtualbox VM configuration) and a separate one with virtually an identical configuration, that is using EFI mode for the VM in Virtualbox, and therefore is also configured to use EFI on the Linux side in NixOS of course.

The current system configuration flake is at https://arrakis.bitgnome.net/nipsy/git/nix/snapshot/nix-master.tar#nixosConfigurations.richese.  What's weird to me is that the referenced supposedly duplicated attribute appears to be the same regardless of whether I'm using the configuration with disko enabled as a module:

```
% nix repl https://arrakis.bitgnome.net/nipsy/git/nix/snapshot/nix-master.tar#nixosConfigurations.richese
Nix 2.24.10
Type :? for help.
Loading installable 'https://arrakis.bitgnome.net/nipsy/git/nix/snapshot/nix-master.tar#nixosConfigurations.richese'...
Added 10 variables.
nix-repl> config.boot.loader.grub.mirroredBoots
[
  { ... }
]
```

or using the configuration where I've removed disko as a module and gone back to a statically defined set of fileSystems:

```
% nix repl https://arrakis.bitgnome.net/nipsy/git/nix/snapshot/nix-8d4fc67e4cdf158b29b19d03781aff420b313157.tar#nixosConfigurations.richese
Nix 2.24.10
Type :? for help.
Loading installable 'https://arrakis.bitgnome.net/nipsy/git/nix/snapshot/nix-8d4fc67e4cdf158b29b19d03781aff420b313157.tar#nixosConfigurations.richese'...
Added 10 variables.
nix-repl> config.boot.loader.grub.mirroredBoots
[
  { ... }
]
```

If I roll back the handful of changes to unmodularize disko for this particular host, then everything works fine again.

The other almost identical system is kaitain.  But since it's using EFI, there doesn't appear to be any problem referring to any attribute named mirroredBoots since I'm using systemd-boot and so none of that grub configuration gets pulled in there.

I'm also not using that mirroredBoots attribute anywhere at all in my entire repository, so it can pretty much only be happening as a result of disko, but I'm really unclear as to why.

I also thought initially that it had something to do with the fact I had named my BIOS partition in my disko configuration as boot and my /boot partition as ESP.  I tried changing that to simply bios and then later MBR to no avail, so that doesn't seem to be the source of the problem.

Obviously reverting back to the non-disko enabled version of the configuration for this particular host works fine for now.  But it would be nice to know if this is simply a bug somewhere or if something else is happening here that I'm not accounting for properly elsewhere in this system's configuration somehow.

iFreilicht:
Fixes #873

Previously, if a user used v2.10.0 (currently unreleased master), an
error like this would be displayed:

    error: attribute '_cliDestroyFormatMount' missing
       at /nix/store/fbqgaij9d3b373rq1iav5glzr4dy77qf-disko/share/disko/cli.nix:73:7:
           72|     else if (lib.traceValSeq hasDiskoModuleFlake) then
           73|       (builtins.getFlake flake).nixosConfigurations.${flakeAttr}.config.system.build.${diskoAttr}
             |       ^
           74|     else

This situation now produces a more helpful error:

    Error: Attribute `nixosConfigurations.testmachine.config.system.build._cliDestroyFormatMount`
           not found in flake `/home/felix/repos-new/disko`!
           This is probably caused by the locked version of disko in the flake
           being different from the version of disko you executed.
    EITHER set the `disko` input of your flake to `github:nix-community/disko/latest`,
           run `nix flake update disko` in the flake directory and then try again,
    OR run `nix run github:nix-community/disko/v1.9.0 -- --help` and use one of its modes.
    
Also, the final config's attributes have different names than those defined in ./default.nix, so even when updating the input on master did not fix this issue previously.


Mic92:
@mergify queue

PR #882, opened 2024-11-16
module: expose tests.bootCommands and tests.enableOCR

amarshall:
This makes it possible to e.g. run `installTest` for configs that need interactive input. See individual commits for details.


iFreilicht (approved):
Ah yes, that was an oversight from when those options were added. Thank you!

iFreilicht:
@mergify queue

PR #883, opened 2024-11-18
ZFS: Add log, spare & dedup vdev types, support multiple special vdevs, fix bug with mixed vdev types

MaienM:
This adds support for the missing categories of devices: `log`, `spare`, and `dedup`.

`log` & `dedup` can each have multiple vdevs assigned (the workload will be load-balanced over these), so these accept a list of vdev configurations. This is actually also true for `special`, but the existing implementation only supported a single vdev. This has been updated to accept either a list of vdevs (like the other two), or a single vdev (for backwards compatibility).

`spare` (like `cache`) just takes a list of device names, neither type supports more complicated vdevs like mirrors.

---

While testing this I ran into a bug that affected all types which support having multiple vdevs (which includes the regular data vdevs). The following topology config:

```nix
{
  vdev = [
    {
      mode = "mirror";
      members = [ "data1" "data2" ];
    }
    {
      members = [ "data3" ];
    }
  ];
}
```
 
would result in the following command:

```shell
zpool create -f <name> mirror /dev/data1 /dev/data2 /dev/data3
```

which would result in a single vdev with a 3-way mirror, rather than a vdev with a 2-way mirror and a second vdev with a single disk. As far as I could find there is no keyword to indicate "mirror config ended, next is a vdev made up of a single disk" (which kinda makes sense given that ZFS would reject such a setup without the `-f` flag), so to fix this I added a sort so that the vdevs with an empty mode are moved to the start. This means the above config will now result in the following command:

```shell
zpool create -f <name> /dev/data3 mirror /dev/data1 /dev/data2
```

(An alternative would have been to instead generate a `zpool add` command for each vdev, but that would be a larger change, and we'd need to start handling partial successes so I opted not to at this point. Doing that would also be a step towards handing making changes to an existing pool such as adding a vdev, so it might be worth considering at some point.)

I extended the `zfs-with-vdevs` example to have this 3-device setup for each of the types that support it so that this is tested, which results in the following pool (which is also validated in the tests):

```
  pool: zroot
 state: ONLINE
config:

        NAME                   STATE     READ WRITE CKSUM
        zroot                  ONLINE       0     0     0
          disk-data3-zfs       ONLINE       0     0     0
          mirror-1             ONLINE       0     0     0
            disk-data1-zfs     ONLINE       0     0     0
            disk-data2-zfs     ONLINE       0     0     0
        dedup   
          disk-dedup3-zfs      ONLINE       0     0     0
          mirror-5             ONLINE       0     0     0
            disk-dedup1-zfs    ONLINE       0     0     0
            disk-dedup2-zfs    ONLINE       0     0     0
        special 
          disk-special3-zfs    ONLINE       0     0     0
          mirror-7             ONLINE       0     0     0
            disk-special1-zfs  ONLINE       0     0     0
            disk-special2-zfs  ONLINE       0     0     0
        logs    
          disk-log3-zfs        ONLINE       0     0     0
          mirror-3             ONLINE       0     0     0
            disk-log1-zfs      ONLINE       0     0     0
            disk-log2-zfs      ONLINE       0     0     0
        cache
          disk-cache-zfs       ONLINE       0     0     0
        spares
          disk-spare-zfs       AVAIL   
```

### DIFF
--- a/lib/types/zpool.nix
+++ b/lib/types/zpool.nix
@@ -68,11 +68,38 @@ in
                         members = [ "x" "y" "/dev/sda1" ];
                       }];
                     };
-                    special = lib.mkOption {
-                      type = lib.types.nullOr vdev;
-                      default = null;
+                    spare = lib.mkOption {
+                      type = lib.types.listOf lib.types.str;
+                      default = [ ];
                       description = ''
-                        A vdev definition for a special device. See
+                        A list of devices to use as hot spares. See
+                        https://openzfs.github.io/openzfs-docs/man/master/7/zpoolconcepts.7.html#Hot_Spares
+                        for details.
+                      '';
+                    };
+                    log = lib.mkOption {
+                      type = lib.types.listOf vdev;
+                      default = [ ];
+                      description = ''
+                        A list of vdevs used for the zfs intent log (ZIL). See
+                        https://openzfs.github.io/openzfs-docs/man/master/7/zpoolconcepts.7.html#Intent_Log
+                        for details.
+                      '';
+                    };
+                    dedup = lib.mkOption {
+                      type = lib.types.listOf vdev;
+                      default = [ ];
+                      description = ''
+                        A list of vdevs used for the deduplication table. See
+                        https://openzfs.github.io/openzfs-docs/man/master/7/zpoolconcepts.7.html#dedup
+                        for details.
+                      '';
+                    };
+                    special = lib.mkOption {
+                      type = lib.types.either (lib.types.listOf vdev) (lib.types.nullOr vdev);
+                      default = [ ];
+                      description = ''
+                        A list of vdevs used as special devices. See
                         https://openzfs.github.io/openzfs-docs/man/master/7/zpoolconcepts.7.html#special
                         for details.
                       '';
@@ -86,7 +113,6 @@ in
                         for details.
                       '';
                     };
-                    # TODO: Consider supporting log, spare, and dedup options.
                   };
                 });
           };
@@ -182,8 +208,17 @@ in
                 entries=()
                 ${lib.optionalString (hasTopology && topology.vdev != null)
                     (lib.concatMapStrings formatVdev topology.vdev)}
-                ${lib.optionalString (hasTopology && topology.special != null)
-                    (formatOutput "special ${topology.special.mode}" topology.special.members)}
+                ${lib.optionalString (hasTopology && topology.spare != [])
+                    (formatOutput "spare" topology.spare)}
+                ${lib.optionalString (hasTopology && topology.log != [])
+                    (lib.concatMapStrings (vdev: formatOutput "log ${vdev.mode}" vdev.members) topology.log)}
+                ${lib.optionalString (hasTopology && topology.dedup != [])
+                    (lib.concatMapStrings (vdev: formatOutput "dedup ${vdev.mode}" vdev.members) topology.dedup)}
+                ${lib.optionalString (hasTopology && topology.special != null && topology.special != [])
+                    (lib.concatMapStrings
+                      (vdev: formatOutput "special ${vdev.mode}" vdev.members)
+                      (lib.lists.toList topology.special) 
+                    )}
                 ${lib.optionalString (hasTopology && topology.cache != [])
                     (formatOutput "cache" topology.cache)}
                 all_devices=()
--- a/lib/types/zpool.nix
+++ b/lib/types/zpool.nix
@@ -1,6 +1,6 @@
 { config, options, lib, diskoLib, rootMountPoint, ... }:
 let
-  # TODO: Consider expanding to handle `disk` `file` and `draid` mode options.
+  # TODO: Consider expanding to handle `file` and `draid` mode options.
   modeOptions = [
     ""
     "mirror"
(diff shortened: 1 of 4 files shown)


lib/types/zpool.nix
```diff
@@ -181,18 +210,31 @@ in
...
+                  # lineformat is type mode=device1 device2 device3
+                  mode="''${line%%=*}"
+                  type="''${mode%% *}"
+                  mode="''${mode#"$type "}"
+                  devs="''${line#*=}"
                   IFS=' ' read -r -a devices <<< "$devs"
                   all_devices+=("''${devices[@]}")
+                  if ! [ "$type" = "$last_type" ]; then
+                    topology+=" $type"
+                    last_type="$type"
```
iFreilicht, line 236: What is last_type used for?

lib/types/zpool.nix
```diff
@@ -68,11 +68,38 @@ in
...
+                      '';
+                    };
+                    dedup = lib.mkOption {
+                      type = lib.types.listOf vdev;
+                      default = [ ];
+                      description = ''
+                        A list of vdevs used for the deduplication table. See
+                        https://openzfs.github.io/openzfs-docs/man/master/7/zpoolconcepts.7.html#dedup
+                        for details.
+                      '';
```
iFreilicht, line 96: It would be nice to add different examples to each of these.

lib/types/zpool.nix
```diff
@@ -134,13 +160,16 @@ in
...
+            entries+=("${type} ${mode}=${
               lib.concatMapStringsSep " "
               (d: if lib.strings.hasPrefix "/" d then d else "/dev/disk/by-partlabel/disk-${d}-zfs") members
             }")
           '';
-          formatVdev = vdev: formatOutput vdev.mode vdev.members;
+          formatVdev = type: vdev: formatOutput type vdev.mode vdev.members;
+          formatVdevList = type: vdevs: lib.concatMapStrings
+            (formatVdev type)
+            (builtins.sort (a: b: a.mode < b.mode) vdevs);
```
iFreilicht, line 172: Is it normal for programming languages to sort the empty string first? Otherwise, a comment might be useful to communicate how much lifting the `<` does.

iFreilicht (commented):
Looks really good! I have a few small notes, but in general, this looks very mergeable 👍

lib/types/zpool.nix
```diff
@@ -1,6 +1,6 @@
 { config, options, lib, diskoLib, rootMountPoint, ... }:
 let
-  # TODO: Consider expanding to handle `disk` `file` and `draid` mode options.
+  # TODO: Consider expanding to handle `file` and `draid` mode options.
```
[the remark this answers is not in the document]
MaienM: It doesn't, or at least not as a keyword like `mirror`/`raidz`/etc. The `disk` section of the man page simply refers to passing in the path of a disk (e.g. `/dev/sda`), which is already supported.

This also applies to `file`, but I left that one since I'm not sure passing in a file is possible with the current setup.

lib/types/zpool.nix
MaienM: It's there to ensure the type's keyword isn't repeated. If you want for example two log mirror vdevs you need to do 

```shell
zpool create -f <name> /dev/sda log mirror /dev/sdb /dev/sdc mirror /dev/sdd /dev/sde
```

and **not**

```shell
zpool create -f <name> /dev/sda log mirror /dev/sdb /dev/sdc log mirror /dev/sdd /dev/sde
```

lib/types/zpool.nix
MaienM: I believe so, yes. In general, a word ending would be sorted before all words that start with the same letters and then continue on (e.g. `a` is before `apple`). This is also how Wikipedia defines an alphabetical sort.

That being said, I think I'll adjust this so that the empty string is first and everything else is left in the order defined by the user. This makes it more explicit, and makes the resulting order more closely match what the user defined.

lib/types/zpool.nix
```diff
@@ -63,19 +63,79 @@ in
                         https://openzfs.github.io/openzfs-docs/man/master/7/zpoolconcepts.7.html#Virtual_Devices_(vdevs)
                         for details.
                       '';
-                      example = [{
-                        mode = "mirror";
-                        members = [ "x" "y" "/dev/sda1" ];
```
MaienM, line 68: I also adjusted this example to

1. be in line with the others
2. actually be valid (passing in a path like `/dev/sda` results in the script failing because `all_devices_list` and `zfs_devices_list` won't match).

lib/types/zpool.nix
MaienM: It is actually, all the special types that support actual vdevs (`log`, `dedup`, and `special`) have two in `zfs-with-vdevs`, and those would all run into the issue that this code avoids. In that example the vdevs are it's a single disk + a mirror, but the principle still applies.

```nix
{
  log = [
    {
      mode = "mirror";
      members = [ "log1" "log2" ];
    }
    {
      members = [ "log3" ];
    }
  ];
}
```

would result in

```shell
zpool create ... log /dev/log3 log mirror /dev/log1 /dev/log2 ...
```

without this logic (which results in the error `invalid vdev specification: 'log' can be specified only once`), and results in

```shell
zpool create ... log /dev/log3 mirror /dev/log1 /dev/log2 ...
```

with it.

iFreilicht: Ahh you're absolutely right!

iFreilicht (approved):

iFreilicht:
@mergify queue

PR #886, opened 2024-11-19
Fix examples/zfs-over-legacy.nix

carschandler:
Shouldn't the legacy example have `options.mountpoint = "legacy"`?


example/zfs-over-legacy.nix
```diff
@@ -48,6 +48,7 @@
           "root/zfs_fs" = {
             type = "zfs_fs";
             mountpoint = "/zfs_fs";
+            options.mountpoint = "legacy";
```
Mic92, line 51: No the idea is that ext4 is mounted with systemd and zfs will mount its own filesystem on top of that.

PR #887, opened 2024-11-20
rewrite disko in C++ (WIP)

Sk7Str1p3:
@iFreilicht as i promised, my current code. for now, only basic functions implemented. it feels bit shitty for me, waiting for response and advice , thx.

### DIFF
--- a/src/disko/cli.cpp
+++ b/src/disko/cli.cpp
@@ -1,9 +1,8 @@
 #include <any>
-#include <cstdlib>
 #include <cstring>
+#include <filesystem>
 #include <getopt.h>
 #include <iostream>
-#include <ostream>
 #include <string>
 
 /*help message*/
@@ -52,7 +51,7 @@ bool verbose = false;
 
 std::any parseArgs(int argc, char **argv) {
 
-  const char *const short_opts = "f:m:dhv";
+  const char *const short_opts = "f:m:dhvF";
   const option long_opts[] = {
       {"flake", required_argument, nullptr, 'f'},
       {"mount", required_argument, nullptr, 'm'},
@@ -60,9 +59,10 @@ std::any parseArgs(int argc, char **argv) {
       {"yes-wipe-everything", no_argument, nullptr, 'y'},
       {"help", no_argument, nullptr, 'h'},
       {"verbose", no_argument, nullptr, 'v'},
-      {"arg", required_argument, nullptr, 'a'}};
+      {"arg", required_argument, nullptr, 'a'},
+      {"file", required_argument, nullptr, 'F'}};
 
-  // break program and show help message if no arguments passed:
+  // break program if no args set:
   if (argc == 1) {
     std::cout << "No mode set!\n" << std::endl;
     showUsage();
@@ -85,8 +85,6 @@ std::any parseArgs(int argc, char **argv) {
       showUsage();
       exit(1);
     }
-
-    // find configuration file:
   }
   // parse additional flags
   while (true) {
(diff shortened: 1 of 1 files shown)


src/disko/cli.cpp
```diff
@@ -0,0 +1,152 @@
...
+      exit(1);
+    }
+  }
+  return disko_Mode;
+  return flake_Path;
+  return mount_Path;
+  return nix_Args;
+  return dry_Run;
+  return omit_Check;
+  return verbose;
```
iFreilicht, line 134: Are multi-returns like this in C++ even an actual feature? I would  expect the first return to immediately end the function.

iFreilicht:
I think you might have misunderstood what "rewrite" means in the context of #789. A rewrite of just the CLI entrypoint in another language will solve nothing. We need to re-implement all the `_create` options in all the `lib/types/*.nix` files. These contain the bash code that's so hard to maintain. In addition, I want to use this opportunity to change to a different execution strategy that allows a user to verify which steps are executed and why before running them.

If you want to make a case for C++ being a good language for this rewrite, you need to show a PoC of that. This is a high ask, I know, but I've been working through multiple iterations of the Python rewrite and haven't even reached that point yet; I myself have not yet shown that a rewrite in Python is even feasible.

src/disko/cli.cpp
Sk7Str1p3: fixed, now function adds variables into array and returns it.

Sk7Str1p3:
> I think you might have misunderstood what "rewrite" means in the context of #789. A rewrite of just the CLI entrypoint in another language will solve nothing. We need to re-implement all the `_create` options in all the `lib/types/*.nix` files. 

No, i understood it correctly. CLI entrypoint is just the beginning of this, I'm going to reimplement those functions and make other minor changes.

iFreilicht:
Okay, good. But in that case I'll unsubscribe from this PR and would ask you to mention me again if anything interesting comes up. I don't want to get pinged every time you push a commit.

Mic92:
If I am honest, I don't feel comfortable maintaining disko if it was written in C++. I maintain a few C++ project and I run into way too many footguns and we would have the issue that we have to compile inside an installer when used in nixos-anywhere beyond just the entrypoint CLI.
If you want to go this route, I would suggest forking the project. We can link to it from here.

Sk7Str1p3:
Wouldn't we have to compile inside an installer in any way? I don't think this is C++ specific problem. Also (at least for now) maintaining C++ code feels not so bad and hard. 

However, if you insist, I can rewrite this in other language. But which one?

Enzime:
My preference would be Rust over C++, however the lower bar to entry in Python is also nice

If we just passed a JSON blob to a compiled program then you wouldn’t need to build it locally if it was cached by Hydra/packaged in Nixpkgs

Mic92:
The rust toolchain is also too big to be downloaded in an in-memory installer. Further details in https://github.com/nix-community/disko/issues/789

Mic92:
I think, I prefer the python approach that @iFreilicht is following. pythonMinimal is still way smaller to upload than a whole gcc toolchain.
If you want to create and maintain your own an alternative disko implementation, we can link it from our README.

Sk7Str1p3:
@Mic92 python **requires** toolchain to be downloaded while gcc toolchain is **included** in NixOS live iso by default.
 
Also, as I said previously, we can use binary cache with C++, but we **can't** go that way with python. Using C++ seems more logical for this.

Sk7Str1p3:
Then why you were offering rewrite Disko in Rust? It has literally same syntax with C++

PR #891, opened 2024-11-21
Add declarative unmount

Mic92:


lib/types/mdadm.nix
```diff
@@ -58,6 +58,17 @@
         lib.optionalAttrs (config.content != null) config.content._mount;
       # TODO we probably need to assemble the mdadm somehow
     };
+    _unmount = diskoLib.mkUnmountOption {
+      inherit config options;
+      default = lib.optionalAttrs (config.content != null) config.content._unmount;
+      # Why does this not work?
```
Mic92, line 64: @Lassulus I don't quite understand why I cannot use `{ fs = ...; dev = ...; }` here.

pogobanane (approved):
from looking at it, looks good

Lassulus (approved):
changes lgtm, since this just adds new functionality and doesn't touch old one. I would be fine with testing this one "in production" :)

Lassulus:
@mergify queue

Mic92:
I actually want to make it a dependency in nixos-anywhere and disko-install

PR #892, opened 2024-11-22
fix: restore nix 2.3 compatibility

zimbatm:
Apparently nix 2.3 didn't support path interpolations.


Mic92:
All because of Tvix, I suppose...

PR #897, opened 2024-11-25
Fix evaluation when a GPT partition is left unformatted

nbdd0121:
Since dcabccaad6, if `content` is not specified, it will generate a confusing error "value is null while a set was expected". This changes disko back to its old behaviour in this specific case (i.e. defaulting to 8300 for an unformatted partition).

### DIFF
--- a/example/gpt-unformatted.nix
+++ b/example/gpt-unformatted.nix
@@ -1,0 +1,36 @@
+# Example to create a GPT partition but doesn't format it
+{
+  disko.devices = {
+    disk = {
+      main = {
+        device = "/dev/vdb";
+        type = "disk";
+        content = {
+          type = "gpt";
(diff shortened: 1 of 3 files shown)


Enzime:
Out of curiosity, what’s the purpose in creating unformatted partitions?

iFreilicht:
Looks great, thank you! 

@Enzime this is useful for creating partitions that you want to write data to directly or if you're working with a filesystem that isn't supported by disko.

iFreilicht:
@mergify queue

nbdd0121:
> Out of curiosity, what’s the purpose in creating unformatted partitions?

In my case I'm just using RAID but not mirroring my ESP partition. So my two disks are layout-wise identical except one holds ESP and another empty space.

PR #898, opened 2024-11-26
module: add default for tests.bootCommand

Lassulus:


iFreilicht (approved):
Can't see anything wrong with this 👍

PR #902, opened 2024-11-27
Rewrite disko in Python

iFreilicht:
See #789 for rationale.

This is very barebones, but I'm at a point where I'm confident enough in the basic structure to get feedback from the community.

## How to test

You can use `nix develop` to get the same shell I've been developing in.
After that, you can just run `./disko2` to try out the CLI, or `pytest` to run the test suite.

Right now, only the `disk`, `gpt` and `filesystem` types are implemented.

The most interesting commands to try are `./disko2 generate`, which generates a `disko-config.nix` at the project root, and `./disko2 format,mount --dry-run disko-config.nix`, which will show you what commands disko would run to align your current system with the target configuration.

Right now, even without `--dry-run`, no commands are executed.

## Thoughts on the rewrite

In general, it was quite a bit of effort to get a disko config into python and use it in a type-safe manner. Now that that's possible, working with the object feels quite nice.

Error handling and communicating issue to the user properly is significantly improved, which is a huge bonus, and being able to generate configs from real world systems and test the plan generation against that should help us test disko a lot more thoroughly.

You will notice that there is a lot more code, though. While you can template complete scripts in python as well, I wanted to actively avoid that to obviate the need for interpreting commands with a shell in the first place. The idea is that all commands will be executed with [`subprocess.run`](https://docs.python.org/3/library/subprocess.html#subprocess.run), which means the binaries are called directly. This will solve escaping, but might make some things (like conditional execution) more complicated to implement.

## To Do

- [ ] Display plan in a human-readable way
- [ ] Add ability to actually run the plan
- [ ] Add VM tests (will require JSON input)
- [ ] Find a good design for conditional execution of some steps
- [ ] Implement more of the types
- [ ] Resolve merge conflicts

After that, I also have to think about how many of the currently nix-only functionality can be implemented. How would something like `config.system.build.format` be generated, for example?


scripts/generate_python_types.py
```diff
@@ -0,0 +1,303 @@
...
+        case "anything":
+            return "Any"
+        # Set up discriminated unions to reduce error messages when validation fails
+        case "deviceType":
+            return '"deviceType" = Field(..., discriminator="type")'
+        case "partitionType":
+            return '"partitionType" = Field(..., discriminator="type")'
+        case _:
+            # Probably a type alias, needs to be quoted in case the type is defined later
+            return f'"{type_str}"'
```
bittner, line 122: With traditional Python you would write it like that:

```suggestion
    types_map = {
        "str": "str",
        "absolute-pathname": "str",
        "bool": "bool",
        "int": "int",
        "anything": "Any",
        # Set up discriminated unions to reduce error messages when validation fails
        "deviceType": '"deviceType" = Field(..., discriminator="type")',
        "partitionType": '"partitionType" = Field(..., discriminator="type")',
    }
    try:
        return types_map[type_str]
    except KeyError:
        # Probably a type alias, needs to be quoted in case the type is defined later
        return f'"{type_str}"'
```

This is (still) the pythonic way. I understand that `match/switch/case` is known in other languages.

scripts/generate_python_types.py
```diff
@@ -0,0 +1,303 @@
...
+def generate_python_code(schema: JsonDict) -> io.StringIO:
+    assert isinstance(schema, dict)
+
+    # Convert disallowed characters in Python identifiers
+    schema = transform_dict_keys(schema, lambda k: k.replace("-", "_"))
+
+    buffer = io.StringIO()
+
+    buffer.write(
+        """# File generated by scripts/generate_python_types.py
```
bittner, line 222: You could write it like that, which helps to keep the comment lines visually aligned.

```suggestion
        """\
# File generated by scripts/generate_python_types.py
```

src/disko/mode_generate.py
```diff
@@ -0,0 +1,75 @@
...
+from disko_lib.eval_config import eval_config_dict_as_json
+from disko_lib.logging import info
+from disko_lib.result import DiskoPartialSuccess, DiskoResult, DiskoError
+from disko_lib.generate_config import generate_config
+from disko_lib.run_cmd import run
+from disko_lib.json_types import JsonDict
+
+DEFAULT_CONFIG_FILE = "disko-config.nix"
+
+HEADER_COMMENT = """
```
bittner, line 12: Add a backslash if you don't want an empty first line.

```suggestion
HEADER_COMMENT = """\
```

src/disko/mode_generate.py
```diff
@@ -0,0 +1,75 @@
...
+# This file was generated by disko generate
+# Some disk and partition names were auto-generated from device attributes
+# to be unique, but might not be very descriptive. Feel free to change them
+# to something more meaningful.
+
+"""
+
+PARTIAL_FAILURE_COMMENT = """
```
bittner, line 20: Add a backslash if you don't want an empty first line.

```suggestion
PARTIAL_FAILURE_COMMENT = """\
```

src/disko/mode_generate.py
```diff
@@ -0,0 +1,75 @@
...
+        ]
+    )
+    if isinstance(config_as_nix, DiskoError):
+        return config_as_nix
+
+    # Contract the main attribute path to a single line to match all the examples
+    nix_code = re.sub(r"^\{ disko = \{ devices", "{ disko.devices", config_as_nix.value)
+    nix_code = re.sub(r"\}; \}$", "}", nix_code)
+
+    with open(DEFAULT_CONFIG_FILE, "w") as f:
```
bittner, line 66: You might want to consider using `pathlib.Path` instead, e.g.

```suggestion
    with Path(DEFAULT_CONFIG_FILE).open("w") as f:
```

bittner (commented):
Just a few comments from a pythonista. :snake: 

Re development tooling, I'd suggest to embrace `uv`, which is about to become the de-facto installer and packaging tool, and `ruff` as a universal linter and code formatter. See [painless-software/cicd/app/cli](https://gitlab.com/painless-software/cicd/app/cli/-/tree/main/frameworks/argparse) (GitLab) for an example CLI setup with Tox that wraps Ruff for formatting and linting.

Some of your code mentions that a Nix test should be run in CI. I use the `docker.io/nixos/nix` container images to run `nix flake check`. See [painless-software/nixos-config](https://gitlab.com/painless-software/nixos-config) (GitLab). If you want this in GHA in this repo I could contribute that.

src/disko/mode_generate.py
iFreilicht: I know pathlib is useful for path operations and type safety, but does it have any advantage here?

src/disko/mode_generate.py
iFreilicht: Ah right, totally forgot about this.

scripts/generate_python_types.py
iFreilicht: True, it's more terse anyway. I used match here initially because I called a function in one of the branches, but when all the results are static, that's not needed anymore.

iFreilicht:
@bittner Thank you for your comments. `ruff` is already used, you can see it in the dependencies for devShell in `flake.nix`. It's not mentioned in the pyproject.toml because we use Nix as the package manager.
I do still have to configure it properly, though.

Regarding uv, I'm not sure I understand how it would fit into the picture here. It seems to be an alternative to pip or poetry, but as we use (and must use) Nix, I'm unsure what it would even do in this project.

We already run nix tests in CI using [buildbot](https://github.com/buildbot/buildbot). You can have a look at other PR's CI Actions if you're curious. Our tests are plentiful and take a long time to evaluate, multiple hours in fact if you were to run them in sequence. Thus, we [nix-fast-build](https://github.com/Mic92/nix-fast-build) to speed up the build time to 45 Minutes locally and ~10 Minutes in CI. Most of our tests also run VMs to check whether the formatting process works correctly, which I think would be tricky to do inside containers, right? If your setup does have unique advantages, let me know.

bittner:
> Regarding uv, I'm not sure I understand how it would fit into the picture here. It seems to be an alternative to pip or poetry, but as we use (and must use) Nix, I'm unsure what it would even do in this project.

It was meant for the dev-setup document. Ack, I'm a beginner in NixOS. What is [uv2nix](https://discourse.nixos.org/t/uv2nix-build-develop-python-projects-using-uv-with-nix/58563), btw, does that fit for using `uv` to develop with Python on NixOS?

> We already run nix tests in CI using [buildbot](https://github.com/buildbot/buildbot).

I wasn't aware of that, sorry. Is the source code of this setup included in this repo?

> If your setup does have unique advantages, let me know.

No, probably not. I had played around with Nix in container images for running checks over my nixos-config and thought I share my findings. I didn't see that anywhere else. Sounds like your setup is much more than just some light linting and `nix flake check`. Sounds good!

src/disko/mode_generate.py
bittner: It's more about consistency in modern Python.

KiaraGrouwstra:
I see a few thumbs-down reactions on this PR - would any of y'all mind elaborating on those concerns to make them more actionable?
/cc @dzmitry-lahoda, @Creative-Difficulty

Creative-Difficulty:
@KiaraGrouwstra 

The reason why I gave a thumbs-down to this PR is that, while I've only written about 5 or so projects in Python, I find its error handling, type system, performance, and general language semantics unsuitable for a lot of serious production software (disko being included). In my opinion, the tool which is responsible for managing the disk/s of computer, a vital part of the NixOS setup for many enthusiasts, shouldn't be written in a garbage-collected high-level language, which severely lacks in performance and is often only used for quick-and-dirty prototyping.

As Rust improves on many of these aspects, and since much of the nix ecosystem is written in it anyways, I'd recommend it as a better fit for this purpose.

Yethal:
There is the disko-rs utility from AerynOS project that could be utilized by disko instead of writing everything from scratch

Creative-Difficulty:
@Yethal Note, however that the crate is called `disks-rs` and per the README provides 

```
partitioning - A partitioning API for manipulating partition tables on block devices. This will be built atop disks and superblock to provide a high level API for partitioning. Currently focused on gpt.
```

Yethal:
There's also a cli installer called lichen built on top of disks-rs that could also be used

dzmitry-lahoda:
As per rational for rewrite I see: 

> Disko is currently somewhat tricky to maintain. Part of this comes from the current architecture that puts nix itself at the center of the entire operation, and uses it as a crude templating engine for bash scripts.

I got myself in my garbage code to same solution and it was hard to  maintain. It is fine for throw away code for my needs.

So seems for disko it should be done better as important tool.
Python feels fine, but sure - Rust(or zig) is better.

midischwarz12:
> So seems for disko it should be done better as important tool. Python feels fine, but sure - Rust(or zig) is better.

I'm thinking about doing a rust implementation and seeing if it is any better. There's also the general critique that disko must be memory constrained for testing. And python really doesn't provide that either.

PR #910, opened 2024-12-04
make-disk-image: improve xargs invocation

danjujan:
When the disk image is too small, `cp` fails because it is out of space. However, `xargs` does not abort immediately but still invokes all remaining `cp` commands, which will of course also fail. That can lead to long waiting times until the make-disk-image script exits with a failure. This PR decreases the time to failure by letting `xargs` abort directly by causing `cp` to fail with exti code 255. 

https://www.man7.org/linux/man-pages/man1/xargs.1.html#DESCRIPTION
```
       If any invocation of the command exits with a status of 255,
       xargs will stop immediately without reading any further input.
       An error message is issued on stderr when this happens.
```

Additionally allow `xargs` to invoke `cp` `nproc` times in parallel with `-P 0` as argument. Since by default the vm uses only one core, this does not change anything. However, for custom `imageBuilder.qemu` options like `-smp 4` this can cause a major speedup. For example, I observed a time reduction from 22 to 12 minutes when building an image with aarch64 binfmt emulation.

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -101,7 +101,7 @@ let
 
     # We copy files with cp because `nix copy` seems to have a large memory leak
     mkdir -p ${systemToInstall.config.disko.rootMountPoint}/nix/store
-    xargs -P 0 ${cpEarlyExit} < ${closureInfo}/store-paths
+    xargs ${cfg.xargsFlags} ${cpEarlyExit} < ${closureInfo}/store-paths
 
     ${systemToInstall.config.system.build.nixos-install}/bin/nixos-install --root ${systemToInstall.config.disko.rootMountPoint} --system ${systemToInstall.config.system.build.toplevel} --keep-going --no-channel-copy -v --no-root-password --option binary-caches ""
     umount -Rv ${systemToInstall.config.disko.rootMountPoint}
--- a/module.nix
+++ b/module.nix
@@ -102,6 +102,13 @@ in
         description = "QEMU image format to use for the disk images";
         default = "raw";
       };
+
+      xargsFlags = lib.mkOption {
+        type = lib.types.str;
+        description = "flags for xargs invocation";
+        default = "";
+        example = "-L 100 -P 0";
+      };
     };
 
     memSize = lib.mkOption {


Enzime (approved):
LGTM

the flags `-P` and `-I` both exist on macOS as well 👍

Enzime:
@mergify queue

lib/make-disk-image.nix
```diff
@@ -94,7 +94,7 @@ let
 
     # We copy files with cp because `nix copy` seems to have a large memory leak
     mkdir -p ${systemToInstall.config.disko.rootMountPoint}/nix/store
-    xargs cp --recursive --target ${systemToInstall.config.disko.rootMountPoint}/nix/store < ${closureInfo}/store-paths
+    xargs -P 0 -I {} sh -c 'cp --recursive --target ${systemToInstall.config.disko.rootMountPoint}/nix/store "$1" || exit 255' sh {} < ${closureInfo}/store-paths
```
Mic92, line 97: Does this not run than one cp for every store path?

Mic92:
@mergify dequeue

Mic92:
Btw we also have parallel copy implemented here: https://github.com/nix-community/disko/pull/788

lib/make-disk-image.nix
danjujan: Currently yes. We could wrap the cp invocation in a short script with `$@` to avoid that.

danjujan:
> Btw we also have parallel copy implemented here: #788

I have seen that. However, this does not do anything about aborting early in out of space situations which was my main motivation for this PR.

Mic92:
I also tried `xargs -P 0` but so far I wasn't able to get any speed up for it. It only got slower. Do you have an example?

Mic92:
Also if the disk is full, how long does it actually take for cp to finish? Should this not rather quickly?

danjujan:
> I also tried `xargs -P 0` but so far I wasn't able to get any speed up for it. It only got slower. Do you have an example?

No public example code but I build an aarch64-linux image on my x86_64-linux host. Just tested it again. 
12 mins with `xargs -P 0` and 8 cores.
18 mins with just `xargs` and 8 cores.
22 mins with just `xargs` and 1 core.


> Also if the disk is full, how long does it actually take for cp to finish? Should this not rather quickly?

cp exits instantly. The problem is that xargs keeps invoking cp with the remaining arguments. I've observed 1-5 mins delay because of that depending at which point the disk gets full.

lib/make-disk-image.nix
danjujan: Addressed by https://github.com/nix-community/disko/pull/910/commits/e6c2ca8c24a1a30ef24e5856436360a1db84a197

danjujan:
Made the xargs flags configurable. Since I had to use `-L 100` after dropping `-I` to get to the 12 minutes. Just `-P 0` without a limit took 16 mins for me.

PR #911, opened 2024-12-05
make-disk-image: create directories images in out path

Mic92:
In case we are crossing some filesystem boundary (i.e. /tmp -> /nix/store), this is faster. Also tmpfs might be probelemantic regarding memory usage.

### DIFF
--- a/disko-install
+++ b/disko-install
@@ -253,7 +253,7 @@ main() {
   # nix copy uses up a lot of memory and we work around issues with incorrect checksums in our store
   # that can be caused by using closureInfo in combination with multiple builders and non-deterministic builds.
   # Therefore if we have a blank store, we copy the store paths and registration from the closureInfo.
-  if [[ ! -d "${mountPoint}/nix/store" ]]; then
+  if [[ ! -f "${mountPoint}/nix/var/nix/db/db.sqlite" ]]; then
     echo "Copying store paths" >&2
     mkdir -p "${mountPoint}/nix/store"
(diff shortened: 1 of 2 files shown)


disko-install
```diff
@@ -253,7 +253,8 @@ main() {
   # nix copy uses up a lot of memory and we work around issues with incorrect checksums in our store
   # that can be caused by using closureInfo in combination with multiple builders and non-deterministic builds.
   # Therefore if we have a blank store, we copy the store paths and registration from the closureInfo.
-  if [[ ! -d "${mountPoint}/nix/store" ]]; then
+  if [[ ! -f "${mountPoint}/nix/var/nix/db/db.sqlite" ]]; then
+    /nix/var/nix/db/db.sqlite
```
iFreilicht, line 257: Is this valid semantically? It looks like you're executing the db, but surely that is not possible, right?

iFreilicht (commented):
I don't quite understand _why_ this is faster, but I'll trust you tried it out. However, changing the output directory should be considered a breaking change, right?
How do we deal with this? I don't want to break people's automation scripts that rely on the disk images being created in the current directory instead of `out`.

disko-install
Mic92: Yeah. I don't know how this got there.

Mic92:
If you do a mv across different filesystems, it's actually a copy. It depends a bit on the build machine if this is the case.

Mic92:
> However, changing the output directory should be considered a breaking change, right? 

No this change is not visible to the end-user. `cfg.extraPostVM` was called after the mv happened.

PR #912, opened 2024-12-07
Sk7Str1p3's python branch

Sk7Str1p3:
Draft PR for python version of Disko by me


Sk7Str1p3:
@iFreilicht question. You made devShell for development, but I have little problem with pylint. Vscode still says something like "could not find module for import", and I can't get rid of feeling something is wrong in the code. Is there any fix for that?

PR #916, opened 2024-12-13
Add extraArgs to zfs_volume.

0x450x6c:
This can be useful, for example we can pass `-s` option to create sparse zvol, etc.

### DIFF
--- a/lib/types/zfs_volume.nix
+++ b/lib/types/zfs_volume.nix
@@ -17,6 +17,11 @@
       default = { };
       description = "Options to set for the dataset";
     };
+    extraArgs = lib.mkOption {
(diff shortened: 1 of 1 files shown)


lib/types/zfs_volume.nix
```diff
@@ -17,6 +17,11 @@
       default = { };
       description = "Options to set for the dataset";
     };
+    extraArgs = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = [ ];
+      description = "Extra arguments";
```
Mic92, line 23:
```suggestion
      description = "Extra arguments passed to zfs create";
```

Mic92:
@mergify queue

PR #921, opened 2024-12-18
zfs: run load-key on mount/unmount

Mic92:


Enzime:
```
[web01] + zpool list zroot
[web01] + zpool import -l -R /mnt zroot
[web01] Key load error: Failed to open key material file: No such file or directory
[web01] 0 / 1 keys successfully loaded
```

This PR didn't work for me, however I'm not sure if it's because I already have an installed system and I was just trying to run the `mount` script

PR #928, opened 2024-12-28
doc: simple xfs with options

aos:
This PR documents an XFS example that uses the quota feature (`pquota`), based on a real project need.


Enzime:
@mergify queue

PR #930, opened 2025-01-06
fix: Apply umask=0077 to /boot partition in all examples

provokateurin:
Most of the examples already have it, but not all of them and I already ran into this problem multiple times when copying example configs.


phaer:
Thank you!

PR #932, opened 2025-01-07
exfat support

Mic92:
Untested. Do we need any kernel modules otherwise?


Lassulus:
I don't think we need any, but I never tried booting from it

PR #934, opened 2025-01-09
export nixosModule as file

Lassulus:


Enzime (approved):
LGTM

flake.nix
```diff
@@ -26,7 +26,7 @@
     in
     {
       nixosModules.default = self.nixosModules.disko; # convention
-      nixosModules.disko.imports = [ ./module.nix ];
+      nixosModules.disko = ./module.nix;
```
Mic92, line 29: What is the rational behind this? If we import modules by file they will have no _file defined. This means if disko is imported twice, there is not deduplication. Also error messages are worse.

Lassulus: it seems this fixes multiple imports of the disko module, since it is a file it implicitly has _file defined

Enzime: The reason why the old code failed to import twice is because the file was imported into Nix manually making it an inline module as opposed to being imported as a file and having `_file` automatically defined

https://github.com/nix-community/disko/commit/0e507d73bfb00a5633a6ee997453607caefbf8ae

PR #944, opened 2025-01-17
disko-install: add --disko-system-config parameter

Lassulus:


disko-install
```diff
@@ -110,6 +110,15 @@ parseArgs() {
       extraSystemConfig="$2"
       shift
       ;;
+    --disko-system-config)
```
Mic92, line 113: Should be part of `usage` with some explanation, since it's not obvious what it's useful for.

PR #949, opened 2025-01-23
mdadm-raid0: remove duplicate boot partition

asymmetric:


example/mdadm-raid0.nix
```diff
@@ -27,10 +27,6 @@
         content = {
           type = "gpt";
           partitions = {
-            boot = {
-              size = "1M";
-              type = "EF02"; # for grub MBR
```
Mic92, line 32: I don't understand the rational here. This is supposed to boot the system in case the first harddrive dies.

phaer: Doesn't make too much difference either way, but this is a raid0 partition with two boot partitions. So without mirroredBoot only the one on the first disk is in use.

But even if they were redundant that wouldn't take us far: / is on raid0 :D

PR #951, opened 2025-01-24
Add configration option for partition UUIDs

### LINKED ISSUE #948: Add option to set UUID of partitions
At the moment, it seems that the UUID of a partition is randomly generated by `sgdisk` at the time it is created. This can potentially lead to reproduceability problems if things elsewhere on the system refer to that UUID. UUID is also intended to uniquely identify a partition, rather than the label/device name as disko currently does by default. It would be useful to be able to specify the partition UUID in the configuration, so that it can be reproduced, and referred to deterministically that way.

Rua:
Closes #948. This is my first ever Nix-related PR, so I'd appreciate feedback if you have any!

I wasn't sure if mdadm does expose `partuuid`, I assumed that it did but that may be wrong. I did run the tests, but I wasn't sure if it was wise to include UUIDs in example files. It's probably not a good idea for users to copy the UUID, they should generate their own.

### DIFF
--- a/lib/types/gpt.nix
+++ b/lib/types/gpt.nix
@@ -210,7 +210,7 @@ in
         if ! blkid "${config.device}" >&2; then
           sgdisk --clear "${config.device}"
         fi
-        ${lib.concatStrings (map (partition:
+        ${lib.concatMapStrings (partition:
           let
             args = ''
               --partition-guid="${toString partition._index}:${if partition.uuid == null then "R" else partition.uuid}" \
@@ -234,7 +234,7 @@ in
               udevadm trigger --subsystem-match=block
               udevadm settle
             ''
-        ) sortedPartitions)}
+        ) sortedPartitions}
 
         ${
           lib.optionalString (sortedHybridPartitions != [])
--- a/lib/types/gpt.nix
+++ b/lib/types/gpt.nix
@@ -38,7 +38,7 @@ in
             type = lib.types.str;
             default =
               if partition.config.uuid != null then
-                "/dev/disk/by-partuuid/${diskoLib.hexEscapeUdevSymlink partition.config.uuid}"
+                "/dev/disk/by-partuuid/${partition.config.uuid}"
               else if config._parent.type == "mdadm" then
                 # workaround because mdadm partlabel do not appear in /dev/disk/by-partlabel
                 "/dev/disk/by-id/md-name-any:${config._parent.name}-part${toString partition.config._index}"
@@ -46,7 +46,7 @@ in
                 "/dev/disk/by-partlabel/${diskoLib.hexEscapeUdevSymlink partition.config.label}";
             defaultText = ''
               if `uuid` is provided:
-                /dev/disk/by-partuuid/''${diskoLib.hexEscapeUdevSymlink partition.config.uuid}
+                /dev/disk/by-partuuid/''${partition.config.uuid}
 
               otherwise, if the parent is an mdadm device:
                 /dev/disk/by-id/md-name-any:''${config._parent.name}-part''${toString partition.config._index}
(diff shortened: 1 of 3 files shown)


lib/types/gpt.nix
```diff
@@ -74,6 +79,16 @@ in
             description = "Name of the partition";
             default = name;
           };
+          uuid = lib.mkOption {
```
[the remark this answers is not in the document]
Rua: I called it `uuid` because, within the Linux context, it's always referred as that. For example, the name of `/dev/disk/by-partuuid`, `blkid` calls it `PARTUUID` etc. I could have named it `partuuid`, but I figured that if there is ever a separate filesystem UUID option, that would be distinguished as `partition.content.uuid` vs `partition.uuid`.

lib/types/gpt.nix
```diff
@@ -74,6 +79,16 @@ in
...
+            type = lib.types.nullOr lib.types.str;
+            default = null;
+            example = "809b3a2b-828a-4730-95e1-75b6343e415a";
+            description = ''
+              The UUID (also known as GUID) of the partition. Note that this is distinct from the UUID of the filesystem.
```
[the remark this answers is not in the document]
Rua: Does disko support anything other than GPT? At least, I included it in the `gpt.nix` module assuming that it would only be available for GPT.

lib/types/gpt.nix
```diff
@@ -37,13 +37,18 @@ in
           device = lib.mkOption {
             type = lib.types.str;
             default =
-              if config._parent.type == "mdadm" then
-              # workaround because mdadm partlabel do not appear in /dev/disk/by-partlabel
+              if partition.config.uuid != null then
+                "/dev/disk/by-partuuid/${diskoLib.hexEscapeUdevSymlink partition.config.uuid}"
```
Lassulus, line 41: do we need to hexescape here? Maybe it is better to add some checks to the type?

lib/types/gpt.nix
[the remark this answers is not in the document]
Lassulus: currently we still support DOS with the legacy table, but that is deprecated and it's fine to do it here for gpt only

lib/types/gpt.nix
```diff
@@ -195,24 +210,31 @@ in
...
+            args = ''
+              --partition-guid="${toString partition._index}:${if partition.uuid == null then "R" else partition.uuid}" \
+              --change-name="${toString partition._index}:${partition.label}" \
+              --typecode=${toString partition._index}:${partition.type} \
+              "${config.device}" \
+            '';
+            createArgs = ''
+              --align-end ${lib.optionalString (partition.alignment != 0) ''--set-alignment=${builtins.toString partition.alignment}''} \
+              --new=${toString partition._index}:${partition.start}:${partition.end} \
+            '';
```
Lassulus, line 224: not sure why this was split, but isn' the device more of a createArg?

Rua: Just to avoid repeating itself, nothing more. It seemed nicer this way. I can revert it to the original style if you want.

lib/types/gpt.nix
Rua: I wasn't sure if there is a special way to indicate in a type, that it should be a string with a particular format. I'd be happy to include that if there is one.

nbraud: You can do that with `lib.types.strMatching` and a regex.  I'm surprised there's no predefined one for UUIDs.

lib/types/gpt.nix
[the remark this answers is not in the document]
nbraud: Yeah, that works.

nbraud (approved):
Feedback addressed

Lassulus:
Changes look good for me. But using inside a test or adding a test would be good

Mic92:
> Changes look good for me. But using inside a test or adding a test would be good

You could extend an existing one.

PR #959, opened 2025-02-04
zfs: fix the zfs package

brianmcgillion:
https://github.com/NixOS/nixpkgs/pull/376115 broke all of downstream
that uses zfs. an explicit dependency on a specific version of zfs must
be defined for all uses.

Combined with the flake lock update to make bisect easier.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -2,11 +2,11 @@
   "nodes": {
     "nixpkgs": {
       "locked": {
-        "lastModified": 1737879851,
-        "narHash": "sha256-H+FXIKj//kmFHTTW4DFeOjR7F1z2/3eb2iwN6Me4YZk=",
+        "lastModified": 1738644632,
+        "narHash": "sha256-DyvJjOOGmTSkkEfHq0oWkwtZOgejYIB5S865wmf/qos=",
         "owner": "NixOS",
         "repo": "nixpkgs",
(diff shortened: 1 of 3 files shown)


phaer:
Would you mind adding a sentence or so to the description of this PR?

I don't think it's self-evident what pinning zfs to a version thats older than current nixos-unstable fixes, if so?

brianmcgillion:
> Would you mind adding a sentence or so to the description of this PR?
> 
> I don't think it's self-evident what pinning zfs to a version thats older than current nixos-unstable fixes, if so?

Absolutely. Just testing that it passes the ci and then I will cleanup the PR. sorry you were quick

phaer:
Thanks for the link and zfs 2.3 instead of 2.2!
...but I still don't get it tbh: The nixpkgs PR you've linked should only affect linuxPackages.zfs, while this PR seems to be about pkgs.zfs which seems to remain unchanged by the PR (and evals just fine in a nix repl on latest master as of posting).
Are you sure that this breaks down stream? Is there a public reproducer to check it's not something else?

brianmcgillion:
https://buildbot.thalheim.io/#/builders/89/builds/990

Mic92:
Can you check if https://github.com/nix-community/disko/pull/960 is enough?
Otherwise, please re-open.

PR #960, opened 2025-02-05
fix: use zfs kernel module which corresponds to user-space tools version in make-disk-image.nix

thenhnn:
nixpkgs removed kernelPackages.zfs with following error:
```
error: linuxPackages.zfs has been removed, use zfs_* instead, or linuxPackages.${pkgs.zfs.kernelModuleAttribute}
```

related to #959, but I don't understand the reason to hard-code the version.

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -27,7 +27,7 @@ let
         ++ cfg.extraRootModules;
       kernel = pkgs.aggregateModules
         (with cfg.kernelPackages; [ kernel ]
-          ++ lib.optional (lib.elem "zfs" cfg.extraRootModules || configSupportsZfs) zfs);
+          ++ lib.optional (lib.elem "zfs" cfg.extraRootModules || configSupportsZfs) cfg.kernelPackages.${config.boot.zfs.package.kernelModuleAttribute});
     }
   // lib.optionalAttrs (diskoLib.vmToolsSupportsCustomQemu lib)
     {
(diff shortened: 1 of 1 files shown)


phaer (approved):
Thanks, LGTM!

(nit: I think we don't really need the with here anymore as it's only used for cfg.kernelPackages.kernel afaict.)

thenhnn:
> Thanks, LGTM!
> 
> (nit: I think we don't really need the with here anymore as it's only used for cfg.kernelPackages.kernel afaict.)

I can do either `cfg.kernelPackages.kernel` or keep with as is and remove `cfg.` in `kernelPackages.${config.boot.zfs...}` invocation, what do you think?

thenhnn:
hm, is `config` always a NixOS system or it can be just attrset with disko?
i see the `config.boot.supportedFilesystems.zfs or false;` part, should I make zfs package default to pkgs.zfs?

phaer:
> I can do either cfg.kernelPackages.kernel or keep with as is and remove cfg. in kernelPackages.${config.boot.zfs...} invocation, what do you think?

I was thinking about the first one, but either way is fine as long as it works. Leaving it as-is is also okay :)

> hm, is config always a NixOS system or it can be just attrset with disko?

Should always be a nixos config, as make-disk-image.nix is a normal nixos module.

thenhnn:
> I was thinking about the first one, but either way is fine as long as it works. Leaving it as-is is also okay :)

I think it's better without 'with', thanks!

PR #961, opened 2025-02-08
Addition of granular bcachefs formatting and multi-disk support

ProjectInitiative:
These changes provide advanced usage of bcachefs formatting options. Because bcachefs provides
    a one-stop-shop for wiping, partitioning, formatting and mounting with unique features for
    each step, it has been broken out into it's own type if users wish to go beyond a single
    disk setup.

    examples/bcachefs.nix has been updated to showcase this advanced usage and some of the
    available options.

    Because all three processes take place via the type object, requiring "disks" attribute set
    is not longer a hard requirement if a bcachefs type is provided. A check has been added to
    account for that use case.

I will note that I have not gone through the official testing process called out in the contributing guidelines quite yet. I will looking into this soon, just wanted to get this PR out there for visibility :)

I have tested provisioning a VM with a 3 disk pool with this branch using nixos-anywhere, and also running the disko tool manually on another, already provisioned VM.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -46,6 +46,7 @@ let
       inherit (diskoLib.types)
         btrfs
         filesystem
+        bcachefs_member
         zfs
         mdraid
         luks
@@ -71,6 +72,7 @@ let
       inherit (diskoLib.types)
         table
         gpt
+        bcachefs_member
         btrfs
         filesystem
         zfs
@@ -92,6 +94,19 @@ let
         default = null;
         description = "The type of device";
       };
+
+    /**
+      Generate a random UUID using uuidgen
+  
+      mkUUID :: AttrSet -> str
+    **/
+    mkUUID =
+      { pkgs
+      , lib ? pkgs.lib
+      }: 
+      lib.removeSuffix "\n" (builtins.readFile (pkgs.runCommand "gen-uuid" {} ''
+        ${pkgs.util-linux}/bin/uuidgen -r > $out
+      ''));
 
     /**
         like lib.recursiveUpdate but supports merging of lists
@@ -612,6 +627,7 @@ let
         devices = {
           inherit (cfg.config)
             disk
+            bcachefs
             mdadm
             zpool
             lvm_vg
@@ -625,6 +641,11 @@ let
             type = lib.types.attrsOf diskoLib.types.disk;
             default = { };
             description = "Block device";
+          };
+          bcachefs = lib.mkOption {
+            type = lib.types.attrsOf diskoLib.types.bcachefs;
+            default = { };
+            description = "bcachefs device";
           };
           mdadm = lib.mkOption {
             type = lib.types.attrsOf diskoLib.types.mdadm;
--- a/lib/types/bcachefs.nix
+++ b/lib/types/bcachefs.nix
@@ -130,8 +130,7 @@
             mkdir -p "${config.mountpoint}"
     
             # Capture both the exit code and output of the mount command
-            output=$(bcachefs mount ${lib.optionalString (config.mountOptions != []) "-o ${lib.concatStringsSep "," config.mountOptions}"} UUID="${config.uuid}" "${config.mountpoint}" 2>&1)
-            exit_code=$?
+            output=$(bcachefs mount ${lib.optionalString (config.mountOptions != []) "-o ${lib.concatStringsSep "," config.mountOptions}"} UUID="${config.uuid}" "${config.mountpoint}" 2>&1 || true)
     
             # Check if the error contains "No such device"
             if echo "$output" | grep -iq "no such device"; then
@@ -141,7 +140,7 @@
             else
                 # Propagate the output and exit code if it's not the expected error
                 echo "$output"
-                exit $exit_code
+                exit 1
             fi
         fi
         '';
--- a/lib/types/bcachefs.nix
+++ b/lib/types/bcachefs.nix
@@ -134,7 +134,7 @@
             exit_code=$?
     
             # Check if the error contains "No such device"
-            if echo "$output" | grep -q "No such device"; then
+            if echo "$output" | grep -iq "no such device"; then
                 echo "Notice: bcachefs mount failed with 'No such device'. This is expected on kernels < 6.13."
                 echo "Current kernel version: $(uname -r)"
                 echo "The mount will succeed when you boot into your final system with a newer kernel."
--- a/lib/types/bcachefs.nix
+++ b/lib/types/bcachefs.nix
@@ -1,0 +1,224 @@
+# lib/types/bcachefs.nix
+{ config, options, lib, diskoLib, ... }:
+{
+  options = {
+    name = lib.mkOption {
+      type = lib.types.str;
+      default = config._module.args.name;
+      description = "Name of the bcachefs pool";
+    };
+
+    type = lib.mkOption {
+      type = lib.types.enum [ "bcachefs" ];
+      default = "bcachefs";
+      internal = true;
+      description = "Type";
+    };
+
+    formatOptions = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = [];
+      description = "Additional options for bcachefs format";
+    };
+
+    mountpoint = lib.mkOption {
+      type = lib.types.str;
+      description = "Mount point for the bcachefs pool";
+    };
+
+    mountOptions = lib.mkOption {
+      type = lib.types.listOf lib.types.str;
+      default = [ "defaults" ];
+      description = "Options to pass to mount";
+      apply = opts: lib.lists.unique (opts ++ [ "nofail" ]);
+    };
+
+    uuid = lib.mkOption {
+      type = lib.types.strMatching "[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{12}";
+      default = let 
+        # Generate a deterministic but random-looking UUID based on the pool name
+        # This avoids the need for impure access to nixpkgs at evaluation time
+        hash = builtins.hashString "sha256" "${config.name}";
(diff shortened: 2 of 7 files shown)


ProjectInitiative:
I am planning on re-working this to follow the mdadm and zfs arch with mult-drive support, it is proving to be a good challenge.

ProjectInitiative:
@Lassulus You mentioned wanting to take a look at the current state. This is very rough, but this is what I currently have.

I added in some markers in the individual `_create` scripts (BCACHEFS POSITION and BCACHEFS_MEMBER POSITION) within [bcachefs.nix](https://github.com/nix-community/disko/pull/961/files#diff-4be10413c7c371db617e028a1d48b0d64721a7710be56e0301a572c61aecc0cdR49) and [bcachefs_member.nix](https://github.com/nix-community/disko/pull/961/files#diff-73fbd1bc731ec534a7524115a810411fe194e9e527df94da1ef134f042d6209eR73) to track the order in which they get collated. When I load it into repl and look at the top level  `_create` function, I see them out of order (member should come first). I am running into an order of ops issue (I think) where I am not properly tracking that disks tagged with `bcachefs_member` need to be processed first.

```nix
 nixosConfigurations.testmachine.config.disko.devices._create
"set -efux\n\ndisko_devices_dir=$(mktemp -d)\ntrap 'rm -rf \"$disko_devices_dir\"' EXIT\nmkdir -p \"$disko_devices_dir\"\n\n( # bcachefs pool1   /mnt/pool #\n  declare -a formatOptions=('--compression=zstd')\n  declare -a mountOptions=(verbose degraded)\n  mountpoint=/mnt/pool\n  name=pool1\n  type=bcachefs\n  echo \"Creating bcachefs device: $device\" >&2\n  \n  echo BCACHEFS POSITION\n
...
...
...
```

Curious your thoughts. Thanks

lib/types/bcachefs.nix
```diff
@@ -0,0 +1,223 @@
...
+            if [ "$format_success" = "false" ]; then
+              echo "Failed to format bcachefs filesystem after $max_attempts attempts"
+              exit 1
+            fi
+
+            udevadm trigger --subsystem-match=block
+            udevadm settle
+          fi
+
+          ${lib.optionalString (config.content != null) config.content._create}
```
Lassulus, line 120: not sure if this makes sense? I don't think we can have other filesystems on top of bcachefs?

ProjectInitiative: I think `${lib.optionalString (config.content != null) config.content._create}` is a leftover artifact from the example I modeled this from, but you are correct, we cannot have any other filesystems on top of bcachefs. I'll remove this if this ever becomes viable.

lib/types/bcachefs.nix
```diff
@@ -0,0 +1,223 @@
...
+    
+            # Capture both the exit code and output of the mount command
+            output=$(bcachefs mount ${lib.optionalString (config.mountOptions != []) "-o ${lib.concatStringsSep "," config.mountOptions}"} UUID="${config.uuid}" "${config.mountpoint}" 2>&1 || true)
+    
+            # Check if the error contains "No such device"
+            if echo "$output" | grep -iq "no such device"; then
+                echo "Notice: bcachefs mount failed with 'No such device'. This is expected on kernels < 6.13."
+                echo "Current kernel version: $(uname -r)"
+                echo "The mount will succeed when you boot into your final system with a newer kernel."
+            else
```
Lassulus, line 140: wouldn't it be better to fail here instead of silently skipping? since we will end up with a wrong mounted filesystems afterwards

lib/default.nix
```diff
@@ -93,6 +95,19 @@ let
...
+      mkUUID :: AttrSet -> str
+    **/
+    mkUUID =
+      { pkgs
+      , lib ? pkgs.lib
+      }: 
+      lib.removeSuffix "\n" (builtins.readFile (pkgs.runCommand "gen-uuid" {} ''
+        ${pkgs.util-linux}/bin/uuidgen -r > $out
+      ''));
+
```
Lassulus, line 110: I guess this code is not needed?

lib/types/bcachefs.nix
ProjectInitiative: Yes, that is the downfall of the current implementation. I wrote a little more in the comments of this PR. It is essentially dead, because of some weird upstream bugs in bcachefs and systemd that prevent multi-disk UUIDs from mounting on boot, so does and will fail in almost every test I have completed using the disko testing framework. I worked around this by ignoring this error during provisioning and adding a custom systemd mount on my servers. This issue makes this PR DOA for the time being :(

lib/types/gpt.nix
```diff
@@ -103,17 +103,29 @@ in
...
-                type = lib.types.nullOr (
-                  lib.types.strMatching "[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{12}"
-                );
-                default = null;
-                defaultText = "`null` - generate a UUID when creating the partition";
-                example = "809b3a2b-828a-4730-95e1-75b6343e415a";
-                description = ''
-                  The UUID (also known as GUID) of the partition. Note that this is distinct from the UUID of the filesystem.
+                type = lib.types.nullOr (lib.types.strMatching "[[:xdigit:]]{8}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{4}-[[:xdigit:]]{12}");
+                default = let
```
Lassulus, line 107: why do we need to change the default?

lib/default.nix
ProjectInitiative: I guess not if I would be changing the default behavior to just hash the name. Good catch

lib/types/gpt.nix
ProjectInitiative: Consistency reasons. Bcachefs does not support mount by label, and managing a multi-disk string such as `/dev/sda1:/dev/sdX...` between steps seemed unmaintainable. Hashing the old determining factor such as the partition names into a deterministic, pre-defined UUID allows bcachefs to set the UUID before formatting. Otherwise if a user left it blank, there would be no easy way to determine which devices are maintained across reboots. Unlike zpool or mdadm which create `/dev/devices` predictably, bcachefs does provide that functionality

lib/types/bcachefs.nix
```diff
@@ -0,0 +1,223 @@
...
+          systemd.services."mount-${lib.escapeSystemdPath config.mountpoint}".serviceConfig.Environment = "BCACHEFS_BLOCK_SCAN=1";
+          systemd.services."unlock-bcachefs-${lib.escapeSystemdPath config.mountpoint}".serviceConfig.Environment = "BCACHEFS_BLOCK_SCAN=1";
+    
+##############################################################################
+# WORKAROUND: Until the following can be addressed. This means using
+# multi-disk bcachefs as a boot/root partition is not possible in its current
+# form with disko
+# https://github.com/koverstreet/bcachefs-tools/issues/308
+# https://github.com/systemd/systemd/issues/8234#issuecomment-1868238750
+##############################################################################
```
Silverdev2482, line 193: NixOS has some special work around for this, though I am not sure what exactly it is. I do have a working multi-device bcachefs rootfs server that has been running for months, though currently I am struggling to get a working bootloader on an arch install for a friend.

ProjectInitiative: If you are able to point me to that work around please post it here. I also worked around this by having an external systemd mount for my non-root pools with a no-fail attribute and retry counter. However due to this weird boot-dependency issue this PR is blocked as it can't pass the VM tests which test root and reboots.

Silverdev2482: I'll try to figure out what it is, probably something with the initramfs, though I don't know much about thoses, but it probably will help with the Arch Install. My config is here: https://github.com/Silverdev2482/Router-Server bcachefs systems just works out of the box, I also can get you any other info about that server, though I am currently on my phone.

Silverdev2482: Well I got it working on the arch install with systemd-boot, I still get some mounting errors on boot, but they appear to be mostly ignorable, besides the delay waiting for it to pass. I think what did it was removing fsck from some config file for the initramfs.

Silverdev2482: I'm not too good with VMs, but I could learn, what VM tests are failing? maybe if I try to get the VM's to boot I could understand why it doesn't work. Though I'm not familiar with disko, just manual formatting, but I would like to help.

ProjectInitiative: One of the problems is found [here](https://github.com/nix-community/disko/pull/961/files#r1992265462)
When disko does it's provisioning script, it tries to mount the FS to verify it works. However the kernel used in the testing VM is not the most up-to-date (This would also fail for nixos-anywhere provisioning if it doesn't have the most up-to-date kernel either). For the bcachefs to be mounted by UUID, I believe kernel 6.13 or newer is required. Otherwise it just fails with "no device found". The other test that fails is when the VM does reboot, it isn't mounted via the systemd mount due to the issues linked at the top of this PR. Systemd is not handling multi-disk FS mounting very well at the moment. 

In theory, we could skip these tests, but I am opposed because it would require users to explicitly provide their own post boot mounting strategy, and that is not maintainable, reproducible or easily debuggable. Furthermore, it means a multi-disk setup can't be used as a boot FS (at least at this present time)

See the [testing guide](https://github.com/nix-community/disko/blob/master/docs/testing.md) for more on how the disko test VM works. It is derived from a Nixos [testing framework](https://nixos.org/manual/nixos/stable/#sec-running-nixos-tests-interactively)

Silverdev2482: I seem to be a bit in over my head, but I haven't given up hope, Right now I'm just trying to spin up the VM, and then I will try to set up a multi device root VM.

nothingnesses:
This is superseded by #1013 and can probably be closed.

ProjectInitiative:
Awesome work, thanks for picking this up!

PR #972, opened 2025-02-15
add treefmt and format all files

DavHau:
- **configure treefmt**
- **treewide: format all files**


Lassulus:
we need to not include the inputs in every downstream user of disko

.git-blame-ignore-revs
```diff
@@ -0,0 +1,35 @@
+# This file contains a list of commits that are not likely what you
+# are looking for in a blame, such as mass reformatting or renaming.
+#
+# If a commit's line ends with `# !autorebase <command>`,
+# where <command> is an idempotent bash command that reapplies the changes from the commit,
+# the `maintainers/scripts/auto-rebase/run.sh` script can be used to rebase
```
Enzime, line 6: Can you change the reference to the script from nixpkgs?

Enzime (approved):
LGTM

PR #992, opened 2025-03-12
Fix jq filter syntax error

marcphilipp:
The trailing comma causes the following error:

```
+ jq -r --arg disk_to_clear /dev/sda -f /nix/store/4vgl19l2swbfvq2x8d7qrsf8qchrypbj-disk-deactivate/disk-deactivate.jq
jq: error: syntax error, unexpected ']' (Unix shell quoting issues?) at <top-level>, line 49:
    ]
jq: 1 compile error
```

### DIFF
--- a/disk-deactivate/disk-deactivate.jq
+++ b/disk-deactivate/disk-deactivate.jq
@@ -45,7 +45,7 @@ def deactivate:
     ]
   elif .type == "swap" then
     [
-      "swapoff \(.path)",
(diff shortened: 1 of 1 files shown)


Mic92:
I thought I already had this fix merged, but I didn't: 
https://github.com/nix-community/disko/pull/993

I also included a regression test in my fix, so I am going with my version.
But thanks for the effort!

Issue #983, opened 2025-02-23
disko-install error "efiSysMountPoint = '/boot' is not a mounted partition"


nothingnesses:
Hello. I've been trying to use disko-install to install NixOS on to my system, but I'm running into an error. Here is my set up:

## /tmp/install/disko/disko-install/configuration.nix
```nix
{ config, lib, pkgs, ... }:
{
  imports =
    [
      ./hardware-configuration.nix
    ];
  boot.loader.systemd-boot.enable = true;
  boot.loader.efi.canTouchEfiVariables = true;
  networking.wireless.iwd.enable = true;
  time.timeZone = "Europe/London";
  i18n.defaultLocale = "en_GB.UTF-8";
  console = {
    keyMap = "uk";
  };
  services.xserver.enable = false;
  services.xserver.xkb.layout = "gb";
  services.pipewire = {
    enable = true;
    pulse.enable = true;
  };
  users.users.a = {
    isNormalUser = true;
    extraGroups = [ "wheel" ];
    packages = with pkgs; [
    ];
  };
  environment.systemPackages = with pkgs; [
  ];
  system.stateVersion = "24.11";
}
```

## /tmp/install/disko/disko-install/flake.nix
```nix
{
  inputs.nixpkgs.url = "github:nixos/nixpkgs?ref=nixos-unstable";
  inputs.disko.url = "github:nix-community/disko/latest";
  inputs.disko.inputs.nixpkgs.follows = "nixpkgs";
  outputs = { self, disko, nixpkgs }: {
    nixosConfigurations.mymachine = nixpkgs.legacyPackages.x86_64-linux.nixos [
      ./configuration.nix
      disko.nixosModules.disko
      {
        disko.devices = {
          disk = {
            main = {
              type = "disk";
              device = "/dev/nvme0n1";
              content = {
                type = "gpt";
                partitions = {
                  ESP = {
                    size = "8G";
                    type = "EF00";
                    content = {
                      type = "filesystem";
                      format = "vfat";
                      mountpoint = "/boot";
                      mountOptions = [ "umask=0077" ];
                    };
                  };
                  root = {
                    size = "100%";
                    content = {
                      type = "luks";
                      name = "crypted";
                      passwordFile = "/tmp/secret.key";
                      settings = {
                        allowDiscards = true;
                      };
                      content = {
                        type = "filesystem";
                        format = "bcachefs";
                        mountpoint = "/";
                      };
                    };
                  };
                };
              };
            };
          };
        };
      }
[...]

PR #242, opened 2023-05-20
Add bcachefs type with support for encryption

onny:

PR #998, opened 2025-03-18
Add bcachefs type that supports encryption

nothingnesses:
This continues work from #242 but uses bcachefs' `-k` option so it uses the keyring session, instead of using the `keyctl` command. It also adds an example and test for an encrypted bcachefs setup and modifies the existing example for the simple, non-encrypted setup to reflect the API change. This PR doesn't aim to support other features of bcachefs such as subvolumes or multi-device setups, which I think deserve their own separate PRs.

This may (partially) address #983, although I haven't tested yet.


nothingnesses:
Superseded by #1013.

PR #1007, opened 2025-04-01
enable basic macos support

Mic92:
We will probably never support native formatting but this is useful when installing NixOS machines and printing formatting script.


Enzime:
Where do you use `disko` on macOS?

Mic92:
It let's me do this:

```
% ./result/bin/disko -m format --flake /Users/joerg/.homesick/repos/dotfiles#turingmachine --dry-run
disko version 1.10.0-dirty
/nix/store/7xf0bk39h1rap48073zsl8dn77kpw5nj-disko-format/bin/disko-format
```

PR #1008, opened 2025-04-02
make-disk-image: add `useVirtualDevices` option

JuneStepp:
I've been using Disko to both configure filesystems on my phone and create the initial images that are flashed. The only problem was that the installed config would have its device names overwritten. This PR adds an option to chose if the virtual/cleaned device names should be installed or not. They're still always used during partitioning.

For those curious, here's an example phone Disko config:
<details>
  <summary><i>expand me</i></summary>
  
```nix
# This config is NOT meant to be used with the Disko CLI. It should only be used with
# the Disko image builder. It represents the boot and root images that will be
# created with Disko image builder and flashed to the phone. NixOS is then able to find
# the partitions using /dev/disk/by-label based on this config.
{
  disko.devices.disk = {
    boot = let
      label = "nixos-boot";
    in {
      type = "disk";
      device = "/dev/disk/by-label/${label}";
      imageName = label;
      # vfat can't be auto-expanded by NixOS, so the initial image size must be
      # the final desired size.
      imageSize = "2G";
      content = {
        type = "filesystem";
        format = "vfat";
        mountpoint = "/boot";
        mountOptions = [
          # See https://github.com/nix-community/disko/issues/527
          "umask=0077"
          # Continuous Discard/trim For SSDs. This partition is small and not written
          # to a ton, so I doubt performance will be an issue.
          "discard"
        ];
        extraArgs = [
          "-n"
          label
          # Sector size.
          "-S"
          "4096"
        ];
      };
    };
    root = let
      label = "nixos-root";
    in {
      type = "disk";
      device = "/dev/disk/by-label/${label}";
      imageName = label;
      imageSize = "5G";
      content = {
        type = "luks";
        name = "crypt";
        settings = {
          # Enable discard/TRIM support. fstrim shouldn't be needed, b/c BTRFS
          # discard=async is default.
          # See https://wiki.archlinux.org/title/Dm-crypt/Specialties#Discard/TRIM_support_for_solid_state_drives_(SSD)
          allowDiscards = true;
          # Improve SSD performance.
          # See https://wiki.archlinux.org/title/Dm-crypt/Specialties#Disable_workqueue_for_increased_solid_state_drive_(SSD)_performance
          bypassWorkqueues = true;
        };
        extraFormatArgs = [
          "--label"
          label
          "--sector-size"
          "4096"
        ];
        # Just for initial encryption in the VM.
        passwordFile = "/tmp/nixos-root.key";
        content = {
          type = "btrfs";
          extraArgs = [
            "--sectorsize"
            "4096"
            "--force" # Force overrite existing partition
          ];
          subvolumes = {
            ...
          };
        };
      };
    };
  };
}
```
</details>

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -134,6 +134,15 @@ in
         description = "QEMU image format to use for the disk images";
         default = "raw";
       };
+
+      useVirtualDevices = lib.mkOption {
+        type = lib.types.bool;
+        description = ''
+          Replace device names with ones usable in VMs. Ex: `/dev/vda`.
+          Disable this if the disk image is for use on real hardware.
+        '';
+        default = true;
+      };
     };
 
     memSize = lib.mkOption {
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -55,26 +55,29 @@ let
       customQemu = cfg.qemu;
     }
   );
+
   cleanedConfig = diskoLib.testLib.prepareDiskoConfig config diskoLib.testLib.devices;
-  systemToInstall = extendModules {
+  virtualDevicesModule = {
+    disko.devices = lib.mkForce cleanedConfig.disko.devices;
+    boot.loader.grub.devices = lib.mkForce cleanedConfig.boot.loader.grub.devices;
+  };
+  system = extendModules {
     modules = [
       cfg.extraConfig
       {
         disko.testMode = true;
-        disko.devices = lib.mkForce cleanedConfig.disko.devices;
-        boot.loader.grub.devices = lib.mkForce cleanedConfig.boot.loader.grub.devices;
       }
     ];
   };
-  systemToInstallNative =
+  cleanedSystem = system.extendModules {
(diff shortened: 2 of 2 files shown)


module.nix
```diff
@@ -121,6 +121,15 @@ in
         description = "QEMU image format to use for the disk images";
         default = "raw";
       };
+
+      useVirtualDevices = lib.mkOption {
```
Mic92, line 125: I still quite don't understand why this is needed after partitioning. Are we not using partition labels in any case?

JuneStepp: There's also `boot.loader.grub.devices`.

JuneStepp: @Mic92 Thoughts?

JuneStepp:
@Enzime Are you perhaps able to look at this or know who can?

JuneStepp:
I'd still really love to get this merged. I see you've been active @Lassulus, do you mind taking a look?

PR #1015, opened 2025-04-13
Add destroy option

pschmitt:
1. This introduces a `destroy` option (which defaults to `true`). Setting it to `false` will prevent the corresponding disk from being wiped during the destroy stage.

2. The second commit (72383cd20fc788fe2a288246b1f010b009b2b074) improves the detection and handling of already existing luks devices. The previous check didn't check for a luks device explicitly, and the luks dev was never opened if the check failed.

Let me know if you prefer splitting these in 2 separate PRs.

### DIFF
--- a/lib/types/disk.nix
+++ b/lib/types/disk.nix
@@ -22,6 +22,11 @@
       type = diskoLib.optionTypes.absolute-pathname; # TODO check if subpath of /dev ? - No! eg: /.swapfile
       description = "Device path";
     };
+    destroy = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "If false, disko will not wipe or destroy this disk's contents during the destroy stage";
+    };
     imageName = lib.mkOption {
       type = lib.types.str;
       default = config.name;
--- a/lib/types/luks.nix
+++ b/lib/types/luks.nix
@@ -130,7 +130,7 @@ in
     _create = diskoLib.mkCreateOption {
       inherit config options;
       default = ''
-        if ! blkid "${config.device}" >/dev/null || ! (blkid "${config.device}" -o export | grep -q '^TYPE='); then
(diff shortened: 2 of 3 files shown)


lib/types/disk.nix
```diff
@@ -22,6 +22,11 @@
       type = diskoLib.optionTypes.absolute-pathname; # TODO check if subpath of /dev ? - No! eg: /.swapfile
       description = "Device path";
     };
+    destroy = lib.mkOption {
+      type = lib.types.bool;
+      default = true;
+      description = "If false, disko will not wipe or destroy this disk's contents during the destroy stage";
```
Mic92, line 28: We already have format. Is this not the same as not destroying?

Mic92: Ok. Maybe this one might be a bit different actually.

lib/types/luks.nix
```diff
@@ -130,7 +130,7 @@ in
     _create = diskoLib.mkCreateOption {
       inherit config options;
       default = ''
-        if ! blkid "${config.device}" >/dev/null || ! (blkid "${config.device}" -o export | grep -q '^TYPE='); then
```
Mic92, line 133: This seems to break stuff:

vm-test-run-disko-luks-btrfs-raid> machine # Device p1 already exists.

pschmitt: yeah I'll add a check since the script is run twice.

pschmitt: Added in a8ab39de646794b1fa6c71b2dd16df93a3a375b9

lib/types/luks.nix
```diff
@@ -130,7 +130,7 @@ in
...
+        if ! blkid "${config.device}" >/dev/null || ! (cryptsetup isLuks "${config.device}"); then
```
Mic92, line 133:
```suggestion
        if ! blkid "${config.device}" >/dev/null || ! cryptsetup isLuks "${config.device}"; then
```

PR #1021, opened 2025-04-25
Make bcachefs subvolumes boot-time mount tests pass + add more bcachefs tests

nothingnesses:
@Mic92 I manually tested our earlier commit and it seems like boot-time mounting of subvolumes actually was working, contrary to what I previously thought. The issue was that the automated test actually needed to unlock each of the mounted subvolumes individually (I didn't realise the boot process presented unlock prompts for each subvolume to be mounted, so the test was only unlocking the first one), which I've now amended with this new commit.

The OCR for the unlock prompts in the automated test seem to take a while though (and also causes my laptop's fans to spin like crazy), so hopefully it doesn't cause any timeouts in the CI.

### DIFF
--- a/example/bcachefs.nix
+++ b/example/bcachefs.nix
@@ -23,7 +23,7 @@
               content = {
                 type = "bcachefs";
                 # This refers to a filesystem in the `bcachefs_filesystems` attrset below.
-                filesystem = "unmounted_subvolumes_in_multi";
+                filesystem = "mounted_subvolumes_in_multi";
                 label = "group_a.vdb2";
                 extraFormatArgs = [
                   "--discard"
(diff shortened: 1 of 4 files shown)


example/bcachefs.nix
```diff
@@ -188,61 +125,40 @@
...
-      #     # This subvolume will be created but not mounted
-      #     "test" = {
-      #     };
-      #   };
-      # };
-
       # Example showing another bcachefs filesystem.
       no_reliance_on_external_subvolume = {
         type = "bcachefs_filesystem";
         mountpoint = "/sometestdir";
```
nothingnesses, line 153: I just realised that this particular example doesn't make sense, since this mount point actually does rely on another one in a different subvolume. Tomorrow (or later today, I should say), I'll remove it or maybe move it to a separate test file altogether.

nothingnesses: Resolved by eae05708d8217dab44ef4e38deeaa73a5008d5f4.

nothingnesses:
@Mic92 I've updated the PR to remove the example that didn't make sense and to include more tests.

gozzarda:
@nothingnesses sorry for the ping, but I think the bot timed out and just needs a sharp poke?

PR #1022, opened 2025-04-27
fix: Use and configure grub for mirrored efi partitions

provokateurin:
This works on my machine, but for some odd reason grub-install fails when running the check for the example:
```
       > machine # Installing for i386-pc platform.
       > machine # /nix/store/6934ywwmhnxzrim4f8hwff6c22v0sf6i-grub-2.12/sbin/grub-install: warning: File system `ext2' doesn't support embedding.
       > machine # /nix/store/6934ywwmhnxzrim4f8hwff6c22v0sf6i-grub-2.12/sbin/grub-install: error: embedding is not possible, but this is required for cross-disk install.
       > machine # /nix/store/j07f0cnyhdx46j9n3a40xhjk7z9d5njc-install-grub.pl: installation of GRUB on /dev/vda failed: No such file or directory
       > machine # Failed to install bootloader
```

I'm not sure what's wrong there, as the example I added is almost exactly like my real config (minus LUKS and BTRFS subvolumes) and I also tried my real config as an example which doesn't work either.
In my NixOS configuration I have `boot.loader.grub.efiSupport = true;` set, but that doesn't seem to make a difference here as the error from the example does not go away with it.

@Lassulus @Enzime @iFreilicht @Mic92 @phaer can you help me out with this one?

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -1033,13 +1033,37 @@ let
             '';
             default =
               with lib;
-              let
-                configKeys = flatten (
-                  map attrNames (flatten (map (dev: dev._config) (flatten (map attrValues (attrValues devices)))))
-                );
-                collectedConfigs = flatten (map (dev: dev._config) (flatten (map attrValues (attrValues devices))));
-              in
-              genAttrs configKeys (key: mkMerge (catAttrs key collectedConfigs));
+              (
+                let
+                  configKeys = flatten (
+                    map attrNames (flatten (map (dev: dev._config) (flatten (map attrValues (attrValues devices)))))
+                  );
+                  collectedConfigs = flatten (map (dev: dev._config) (flatten (map attrValues (attrValues devices))));
+                in
+                genAttrs configKeys (key: mkMerge (catAttrs key collectedConfigs))
+              )
+              // (
+                let
+                  efi_partitions = builtins.filter (part: part.type == "EF00") (
+                    flatten (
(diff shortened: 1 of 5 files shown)


lib/default.nix
```diff
@@ -1033,13 +1033,37 @@ let
...
+                    flatten (
+                      map (disk: map (part: part // { device = disk.device; }) (attrValues disk.content.partitions)) (
+                        flatten (map attrValues (attrValues devices))
+                      )
+                    )
+                  );
+                in
+                (optionalAttrs (builtins.length efi_partitions >= 2) {
+                  # Mirrored boot partitions are not supported on systemd-boot.
+                  boot.loader.grub.enable = mkForce true;
```
provokateurin, line 1057: This should be an assertion for `system.boot.loader.id == "grub"`, but I'm not sure how to access `config` here.

provokateurin:
I fixed the issue with my new example/test (see first commit), however it breaks the legacy tables for some reason. The second commit also doesn't work with legacy tables, but it is probably possible to make it work somehow. It just needs to detect the EFI partitions, but I'm not sure how that would work with the legacy tables.

So, would it be allowed to get rid of legacy tables support? AFAICT it has been deprecated for over two years (https://github.com/nix-community/disko/pull/294), so it should be fine to remove it now.

I would greatly appreciate your input and help on this issue @Mic92 @Lassulus!

PR #1028, opened 2025-05-07
bcachefs: fix ENOENT during boot

ProjectInitiative:
Addresses: https://github.com/koverstreet/bcachefs/issues/812
Ran into this issue, adding the full path seems to mitigate for the time being.
@nothingnesses


nothingnesses:
@ProjectInitiative ProjectInitiative#1

PR #1031, opened 2025-05-08
diskoImageScript: fix compat with nixpkgs-unstable

Mic92:


zackw:
Hey, thanks for improving the documentation here!  Can I make just one small additional suggestion: please add instructions for making your test VM overlay-mount `/nix/store` from the host, so people don't have to use `copyNixStore = true`?

Mic92:
> Hey, thanks for improving the documentation here! Can I make just one small additional suggestion: please add instructions for making your test VM overlay-mount `/nix/store` from the host, so people don't have to use `copyNixStore = true`?

Not quite sure what this means. Currently this just tests the built image. Why would we share state with the host system?

Mic92:
What you are proposing sounds more like the qemu-vm.nix profile.

zackw:
Unless #778 gets merged (and currently I don't think it _should_ be merged, for reasons discussed over there), the `vmWithDisko` output is not useful to me. You added nice clear instructions for how to use the disk image produced by `diskoImagesScript` for testing, in the same way one might use `vmWithDisko`. But because I'm only using it for testing, I would like those instructions to explain how to share `/nix/store` with the host, so I don't have to twiddle my thumbs while it's copied into the disk image, repeat on every test cycle.  Does that make sense?

PR #1056, opened 2025-06-06
Better handling of bind mounts in make-disk-image

zaninime:
The device of the bind mounts is incorrectly taken verbatim while making the image. The root hierarchy is mounted at `config.rootMountDir` (`/mnt` by default) during the process, and this commit prepends that path to the `device` in these cases.


jgus:
Just chiming in with a +1; I just hit this and would love to see this get in

PR #1062, opened 2025-06-09
Allow disabling copying of udev rules

zaninime:
Currently, the builder builds the whole `system.build.etc`, which is a rather enormous closure in some configurations. Also, it requires the build of all packages declared in `systemPackages` for the build architecture, which is in many cases unneeded, and can fail if the packages are platform-specific.

### DIFF
--- a/lib/make-disk-image.nix
+++ b/lib/make-disk-image.nix
@@ -121,7 +121,9 @@ let
     ln -sfn /proc/self/fd/2 /dev/stderr
     mkdir -p /etc/udev
     mount -t efivarfs none /sys/firmware/efi/efivars
-    ln -sfn ${systemToInstallNative.config.system.build.etc}/etc/udev/rules.d /etc/udev/rules.d
+    ${lib.optionalString cfg.useUdevRules ''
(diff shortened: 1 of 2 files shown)


Lassulus:
We usually need the udev rules to create the /dev/disk/by-partlabel symlinks. So without them disko falls apart a bit, so I'm curious what your usecase is and why you don't need them :)

maybe a better approach would be, to just copy the udev rules we need instead of all of them?

zaninime:
My usecase was building an image for Raspberry Pi, and with this (and the other PRs I opened) I was successful. :)

I do see in my fstab:

```
/dev/disk/by-partlabel/system /system ext4 x-initrd.mount,noatime 0 2
```

where `system` is one of the partitions managed by disko, so this has to be working also without udev rules. I omitted the other partitions for brevity.

In my case particularly there's some software that is not available in caches and I do not wish to rebuild it in the cross configuration just for the sake of building the image. It has no impact on udev rules. One example: https://github.com/raspberrypi/rpicam-apps depends on a custom ffmpeg, qt, ... and eventually even fails to compile in the cross config.

PR #1091, opened 2025-07-14
Add jq as dependency to diskoScriptNoDeps

hgl:
NixOS installer by default doesn't include `jq`. Using `nixos-anywhere --no-disko-deps` can fail with `jq` not being found.

I'm not sure this is the right fix, but is the easiest one, and `jq` is small enough to not defeat the purpose of nodeps scripts.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -860,6 +860,7 @@ let
                   })
                     "disko"
                     ''
+                      export PATH=${lib.makeBinPath [ pkgs.jq ]}:$PATH
                       ${cfg.config._disko}
                     '';
               };


Lassulus:
hmm, nodeps should actually mean no deps, I usually create a new iso with jq, I guess the right fix would be to port from jq to awk or something else that is available on the iso.

But what is the usecase here? why not use the version with dependencies or is this called from another tool?

Lassulus:
the actual problem is also, that this induces a dependency on the system, currently noDeps can be build on aarch64-linux and still executes on x86, with this change, the jq package will depend on the architecture

hgl:
> But what is the usecase here?

When using the version with dependencies in `nixos-anywhere`, it can fail if the remote server has limited amount of memory, since all the dependencies are installed to the ram on the NixOS installer.

> the jq package will depend on the architecture

Do you think this PR should be made to `nixos-anywhere`? Maybe use `nix profile install` to install jq before calling this script? Wonder what @Mic92 thinks.

hgl:
jq has been made available in the official NixOS installer.

PR #1094, opened 2025-07-19
allow subtype definitions to be broken up

hgl:
Currently `subtype` suffers from a few issues:

1. If the definition is broken up, it mistakenly thinks [a partial definition is invalid](https://github.com/NixOS/nixpkgs/pull/254790/files#r1358073458).
2. Only the last partial definition is used.

This PR addresses them.

### DIFF
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -23,21 +23,40 @@ let
       lib.mkOptionType {
         name = "subType";
         description = "one of ${lib.concatStringsSep "," (lib.attrNames types)}";
-        check =
-          x:
-          if x ? type then
-            types.${x.type}.check x
-          else
-            throw "No type option set in:\n${lib.generators.toPretty { } x}";
+        check = x: lib.isAttrs x;
         merge =
-          loc:
-          lib.foldl' (
-            _res: def:
-            types.${def.value.type}.merge loc [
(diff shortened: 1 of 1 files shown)


Lassulus:
this looks pretty good, can we add a test that shows that we can split it up now? :)

Enzime:
Not sure if this is going to be very heavy because it uses `evalModules` as part of the `merge` function

Lassulus:
hmm, the test framework is really too opinionated to test this. I don't think computation will be heavy with this, since the evalModules should be fast with the few options available. So I'm gonna merge this for now and try to integrate it into the nixpkgs PR at a later point.

Thanks a lot!

Enzime:
Just tested this and it works well 👍 Thanks

One example using this PR is for having a different FDE password for your VM:

```nix
{
  virtualisation.vmVariantWithDisko = {
    disko.devices.disk.primary.content.partitions.luks.content.passwordFile =
      lib.mkForce (toString (pkgs.writeText "password" "apple"));
  };
}
```

Enzime:
This PR allows us to finish #480 which I think should refactor `makeDiskoTest` to instead use `evalModules` and be brought in line with the `virtualisation.vmVariantWithDisko` interface

PR #1095, opened 2025-07-20
feat: add zfs-encrypted-key-on-luks-zvol.nix

ncrmro:
This file creates root on ZFS where the dataset is encrypted by a key that lives on a LUKS encrypted zvol. Allowing multiple passwords/TPM to unlock.

The only problem with it right now is that when using nixos-anywhere to run the install it hangs until I run `cryptsetup luksClose /dev/mapper/credstore`, which to `postCreateHook` is supposed to do.

When I run this I get

```
[root@nixos:~]# cryptsetup luksClose /dev/mapper/credstore
Device /dev/zd0 does not exist or access denied.
Device /dev/zd0 does not exist or access denied.
Device /dev/zd0 does not exist or access denied.

[root@nixos:~]# 
Broadcast message from root@nixos (Sun 2025-07-20 16:48:34 UTC):

The system will reboot now!
```

I understand it may not fit an example here but I'd like to see if anyone has any ideas for the `luksClose` issue..

### DIFF
--- a/example/zfs-encrypted-key-on-luks-zvol.nix
+++ b/example/zfs-encrypted-key-on-luks-zvol.nix
@@ -1,0 +1,208 @@
+{ lib, config, utils, ... }:
+{
+  disko.devices = {
+    disk.disk1 = {
+      type = "disk";
+      device = "/dev/vda";
+      content = {
+        type = "gpt";
+        partitions = {
+          esp = {
+            name = "ESP";
+            size = "500M";
+            type = "EF00";
+            content = {
+              type = "filesystem";
+              format = "vfat";
+              mountpoint = "/boot";
+            };
+          };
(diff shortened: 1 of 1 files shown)


almereyda:
This is interesting and exactly how Ubuntu does. I'd suggest to replicate the behaviour in an accompanying `tests/zfs-encrypted-key-on-luks-zvol.nix` file and work it from there.

yushisora:
Thanks for the config.

I believe disko does [add](https://github.com/nix-community/disko/blob/5ad85c82cc52264f4beddc934ba57f3789f28347/lib/types/zfs_fs.nix#L177) `zfsutil` options to the filesystems so perhaps there is no need to do it again at the end of the config?

PR #1098, opened 2025-07-24
docs: add btrfs raid on luks devices non-root example

SolidEva:
useful for creating and managing a non-root, encrypted data raid on a headless server

the major difference between this and https://github.com/nix-community/disko/blob/master/example/luks-btrfs-raid.nix
is that this configuration is not expected to be decrypted at boot time

because of that, we set options
```
            initrdUnlock = false; # dont try to unlock this in the initrd
            "noauto" # ensure that systemd doesnt try to mount this at boot
            "nofail" # ensure that systemd failing to mount this doesn't send us to emergency mode
```

tested using a server board with 4x nvme drives

happy to do additional testing


SolidEva:
testing with `checkScripts = false` for now, but before this merges, a fix for https://github.com/nix-community/disko/issues/1099 will have to be merged.

Issue #1102, opened 2025-08-09
Different password can pass LUKS creation


Eason20000:
I have a disko configuration file like this:

``` nix
{
  disko.devices = {
    disk = {
      nvme-Lexar_SSD_<filtered> = {
        type = "disk";
        device = "/dev/disk/by-id/nvme-Lexar_SSD_<filtered>";
        content = {
          type = "gpt";
          partitions = {
            ESN20-Z390_boot = {
              size = "1G";
              type = "EF00";
              content = {
                type = "filesystem";
                format = "vfat";
                mountpoint = "/boot";
                mountOptions = [ "umask=0077" ];
              };
            };
            ESN20-Z390_system-0 = {
              size = "100%";
              content = {
                type = "luks";
                name = "ESN20-Z390_system-0";
                settings.allowDiscards = true;
                content = {
                  type = "lvm_pv";
                  vg = "ESN20-Z390_system";
                };
              };
            };
          };
        };
      };

      nvme-Samsung_SSD_990_PRO_1TB_<filtered> = {
        type = "disk";
        device = "/dev/disk/by-id/nvme-Samsung_SSD_990_PRO_1TB_<filtered>";
        content = {
          type = "luks";
          name = "ESN20-Z390_system-1";
          settings.allowDiscards = true;
          content = {
            type = "lvm_pv";
            vg = "ESN20-Z390_system";
          };
        };
      };

      ata-ST4000VX007-<filtered> = {
        type = "disk";
        device = "/dev/disk/by-id/ata-ST4000VX007-<filtered>";
        content = {
          type = "luks";
          name = "ESN20-Z390_data-0";
          content = {
            type = "lvm_pv";
            vg = "ESN20-Z390_data";
          };
        };
      };

      ata-WDC_WD60EDAZ-<filtered> = {
        type = "disk";
        device = "/dev/disk/by-id/ata-WDC_WD60EDAZ-<filtered>";
        content = {
          type = "luks";
          name = "ESN20-Z390_data-1";
          content = {
            type = "lvm_pv";
            vg = "ESN20-Z390_data";
          };
        };
      };

    };

    lvm_vg = {
      ESN20-Z390_system = {
        type = "lvm_vg";
        lvs = {
          nix = {
            size = "64G";
            content = {
              type = "filesystem";
              format = "ext4";
              mountpoint = "/nix";
            };
          };
          persistent = {
            size = "100%FREE";
            content = {
              type = "filesystem";
              format = "ext4";
[...]

PR #1103, opened 2025-08-10
fix: the method of veryfing password during LUKS creation

Eason20000:
Fixed issue:
#1102

### DIFF
--- a/lib/types/luks.nix
+++ b/lib/types/luks.nix
@@ -143,19 +143,21 @@ in
 
             askPassword() {
               if [ -z ''${IN_DISKO_TEST+x} ]; then
-                set +x
                 promptSecret "Enter password for ${config.device}: " password
                 promptSecret "Enter password for ${config.device} again to be safe: " password_check
                 export password
                 [ "$password" = "$password_check" ]
-                set -x
               else
+                set -x
                 export password=disko
+                set +x
               fi
             }
+            set +x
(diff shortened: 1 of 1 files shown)


Lassulus:
ah the `set -x` is used to enable debug output (and set +x disables it again), we probably need to call exit 1 if the passwords don't match, not sure why it is working if we rearrange the debug outputs

Eason20000:
Maybe it's because `until` catches the last command's exit code in the function? And `set -x` is the last so it is always `0` and no matter what `[ $password = $password_check ]`'s exit code is...?

Eason20000:
Also, is it possible to automatically test the function of checking if the password is identical? I know little programming so I don't know how to do it...

lib/types/luks.nix
```diff
@@ -143,19 +143,21 @@ in
 
             askPassword() {
               if [ -z ''${IN_DISKO_TEST+x} ]; then
-                set +x
                 promptSecret "Enter password for ${config.device}: " password
                 promptSecret "Enter password for ${config.device} again to be safe: " password_check
                 export password
                 [ "$password" = "$password_check" ]
```
Lassulus, line 149: probably better to use exit 1 if the password don't match instead of setting the debug flags

Lassulus:
should be fixed by https://github.com/nix-community/disko/pull/1105

PR #1122, opened 2025-09-08
Try an auto-generated doc for disko

EliRibble:
This really shouldn't be considered a "real" PR. I noticed that the auto-generated reference documentation has been considered broken since [commit ea3e29fe7d02e2d951c9deea93057330d0e18fec](https://github.com/nix-community/disko/commit/ea3e29fe7d02e2d951c9deea93057330d0e18fec), which essentially means it's never worked.

I generated this documentation with:

```shell
nix build github:Thunderbottom/nix-options-doc
./result/bin/nix-options-doc --path lib/ --out docs/reference.md
```

I don't expect this to be accepted, but wanted to use it to start a discussion - at a high level, what do you want the auto-generated reference documentation to look like? Are you willing to add comments to the disko code in order to improve auto-generated documentation?

## Summary by CodeRabbit

* **Documentation**
  * Replaced the CLI-centric “Reference Manual: disko” with a structured NixOS Module Options reference.
  * Added comprehensive, typed options including defaults, descriptions, and examples.
  * Documented deeply nested schemas (e.g., partitions, LVM, GPT, LUKS, RAID, ZFS, bcachefs, swap, subvolumes, pools).
  * Removed legacy command-line options section and manual narratives.
  * Introduced consistent, auto-generated formatting for easier discovery and configuration.

### DIFF
--- a/docs/reference.md
+++ b/docs/reference.md
@@ -1,47 +1,1045 @@
-# Reference Manual: disko
+# NixOS Module Options
 
-## Module Options
 
-We are currently having issues being able to generate proper module option
-documentation for our recursive disko types. However you can read the available
-options [here](https://github.com/nix-community/disko/tree/master/lib/types).
-Combined with the
-[examples](https://github.com/nix-community/disko/tree/master/example) this
-hopefully gives you an overview.
+## [`options.type`](types/lvm_pv.nix#L12)
 
-## Command Line Options
+Type
 
+**Type:** `lib.types.enum [ "lvm_pv" ]`
+
+## [`options.device`](types/lvm_pv.nix#L17)
+
+Device
+
+**Type:** `lib.types.str`
+
+**Default:** `device`
+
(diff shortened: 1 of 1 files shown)


Enzime:
Seems like this didn't understand the different types and merged all the options for each of the different types together

Mic92:
@EliRibble something that could seperate the different types might be a good start and later soemthing that gives an overview what types can be embedded in what types (maybe some sort of diagram)?

EliRibble:
> soemthing that gives an overview what types can be embedded in what types (maybe some sort of diagram)?

Yeah, I think I can do something like that, though I'm sure I'll have to generate it myself rather than rely on an automated system.

How do y'all feel about potentially changing the way that comments are done in the code to work better with auto-generated documentation?

Enzime:
Ideally if you could make it work without relying on comments that would be great

PR #1140, opened 2025-09-30
Fix zfs with encryption mount/unmount

### LINKED ISSUE #1139: Unmount script on zfs-encrypted example fails at unload-key
Error `cannot open 'root': dataset does not exist` when executing the disko unmount script on the example configuration `zfs-encrypted-root.nix`.

Image

# Reproduction

1. Deploy the zfs-encrypted-root.nix configuration on top of a (virtual) disk
    * Use the diskoScript attribute script to format+mount
2. Cleanup the disko managed mounts
    * Use the unmount attribute script to unmount
3. Observe error message after the `zfs unload-key` command

## Expected to happen

Key for dataset `zroot/root` is sucessfully unloaded IF it was previously loaded.

## What happens

Command `zfs unload-key root` fails because the filesystem argument is not zroot/root (poolname + dataset path).

# Possible fix

I think the filesystem name argument for `zfs unload-key` has to be changed to become the poolname + dataset path. AKA `"${config._parent.name}/${config.name}"` ?
RELATED; The `zfs load-key` in the mount script is also broken probably? This flies under the radar (I guess) due to testing sequence "create" + "mount"; where "create" already does correct encryption setup.

https://github.com/nix-community/disko/blob/67ff9807dd148e704baadbd4fd783b54282ca627/lib/types/zfs_volume.nix#L91-L100

Image

Bert-Proesmans:
Fixes #1139 

- Adds unmount validation to example tests (module mode)
- Updates zfs_fs and zfs_volume to use the full dataset path consistently

Currently running all tests locally, new/other issues might pop up due to change to test script.

### DIFF
--- a/lib/tests.nix
+++ b/lib/tests.nix
@@ -348,6 +348,9 @@ let
               machine.succeed("${lib.getExe nodes.machine.system.build.format}")
               machine.succeed("${lib.getExe nodes.machine.system.build.mount}")
               machine.succeed("${lib.getExe nodes.machine.system.build.mount}") # verify that mount is idempotent
+              machine.succeed("${lib.getExe nodes.machine.system.build.unmount}")
+              machine.succeed("${lib.getExe nodes.machine.system.build.unmount}") # verify that unmount is idempotent
+              machine.succeed("${lib.getExe nodes.machine.system.build.mount}") # verify that mount is idempotent
               machine.succeed("${lib.getExe nodes.machine.system.build.destroyFormatMount} --yes-wipe-all-disks") # verify that we can destroy and recreate again
               machine.succeed("mkdir -p /mnt/home")
               machine.succeed("touch /mnt/home/testfile")


Bert-Proesmans:
You can cherry-pick my changes to the ZFS types, otherwise decisions have to be made about how to handle the failing tests.

Enzime:
I suspect all the tests are failing because all the other types don't have idempotent unmounts

Bert-Proesmans:
i removed the change to the partprobe command in zfs_volume create step. The argument to partprobe is not related to the dataset path but to the device path **EDIT;** in the disko abstraction layer.

> I suspect all the tests are failing because all the other types don't have idempotent unmounts

Yeah, I understand i'm pushing on invariants/guarantees with this PR. But the lack of tests for unmounting practically comes down to "unmounting is not guaranteed", let alone that it happens clean/idempotent (to match naive user expectations).

That's why I commented to cherry-pick or lay out decisions, because this is just a drive-by contribution and I'm not a maintainer. I can adjust the PR.

Enzime:
If you're interested in updating the rest of the types to support idempotent unmounts, that would be great 👍

Bert-Proesmans:
Rebased and got the errors for the examples down to <10. Fixing the leftover issues is certainly not easy 😅
I'll take another look soon

Bert-Proesmans:
Closing this PR to push separate PR's instead grouped by type of changes.

I thought I was clever by stacking my PR's but it looks like I'm misinterpreting how that should work?

PR #1147, opened 2025-11-05
Unmount fixes

Bert-Proesmans:
Part of stacked pull requests to enable unmount testing. This is PR 1 out of 4.

These commits are grouped on the idea that they are non-breaking fixes to existing working config, and also unbreak non-working config.

Next PR #1148 
See #1140 for ealier discussion.

PR #1148, opened 2025-11-05
Bugfix unmounting encrypted ZFS volume

### LINKED ISSUE #1139: Unmount script on zfs-encrypted example fails at unload-key

Bert-Proesmans:
This PR introduces new option and proper unmounting in case of encryption on ZFS volumes.

Part of stacked pull requests to enable unmount testing. This is PR 2 out of 4. Stacks on top of #1147.
Only the last two commits apply to this specific PR.

Previous PR #1147, next PR #1149
See #1140 for ealier discussion.

Closes #1139.

### DIFF
--- a/lib/types/zfs_volume.nix
+++ b/lib/types/zfs_volume.nix
@@ -88,8 +88,8 @@
       default = {
         dev = ''
           ${lib.optionalString (config.options.keylocation or "none" != "none") ''
-            if [ "$(zfs get keystatus ${config.name} -H -o value)" == "unavailable" ]; then
-              zfs load-key ${config.name}
+            if [ "$(zfs get keystatus ${config._name} -H -o value)" == "unavailable" ]; then
+              zfs load-key ${config._name}
             fi
           ''}
 
@@ -102,9 +102,11 @@
       inherit config options;
       default = {
         dev = ''
-          ${lib.optionalString (
-            config.options.keylocation or "none" != "none"
-          ) "zfs unload-key ${config.name}"}
+          ${lib.optionalString (config.options.keylocation or "none" != "none") ''
+            if [ "$(zfs get keystatus ${config._name} -H -o value)" == "available" ]; then
+              zfs unload-key ${config._name}
+            fi
+          ''}
 
           ${config.content._unmount.dev or ""}
         '';
--- a/lib/types/zfs_volume.nix
+++ b/lib/types/zfs_volume.nix
@@ -13,6 +13,12 @@
       default = config._module.args.name;
       description = "Name of the dataset";
     };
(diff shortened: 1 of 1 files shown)


Bert-Proesmans:
The PR (and entire stack) is rebased on current master and tests are running at the moment. My spot checking didn't see any discrepancies since last push. I don't expect failing tests here nor in #1150. 

I hope this PR can get merged soon!

Enzime:
I'm still confused why these PRs are stacked, the PRs are independent? are they just stacked because you want to be able to test all of them?

Bert-Proesmans:
> I'm still confused why these PRs are stacked, the PRs are independent?

The first three PR's are indeed independent. I could/should have simplified the approach; not stack 1147/1148/1149 and stack only in 1150.

> are they just stacked because you want to be able to test all of them?

Yes

---

The 2 lines of code I want to add(#1150) to the tests don't work without dependency work (refactoring, fixes, missing functionality) to the disko types. The changes to these types, especially the mdam change (#1149), need independent/isolated discussion.
By stacking the branches, the CI run on the last PR tests all code at once. So you and me know ahead of time that the end-result remains in good state whatever happens in the middle.

I don't know how to explain differently without sounding patronizing, it's just a means to reach a goal with waypoints instead of straight start -> finish. I don't fully understand/never worked with many of the disko supported filesystems, for me this is unknown territory.
Best case this all gets merged as-is, worst case this opens heavy discussion and multiple commits need rework. I can't predict this, so i'm proactive in setting the environment to track progress.

PR #1161, opened 2025-11-20
stop using pkgs.path

Mic92:
pkgs.path is an expensive copy operation that slow down evaluation by a lot.

### DIFF
--- a/module.nix
+++ b/module.nix
@@ -4,6 +4,7 @@
   pkgs,
   extendModules,
   diskoLib,
+  modulesPath,
   ...
 }:
 let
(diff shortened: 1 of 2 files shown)


module.nix
```diff
@@ -245,8 +246,9 @@ in
     _module.args.diskoLib = import ./lib {
       inherit lib;
       rootMountPoint = config.disko.rootMountPoint;
-      makeTest = import (pkgs.path + "/nixos/tests/make-test-python.nix");
-      eval-config = import (pkgs.path + "/nixos/lib/eval-config.nix");
+      makeTest = import "${modulesPath}/../tests/make-test-python.nix";
```
Enzime: Isn’t `modulesPath` already pointing to the `niĩos` directory?

PR #1166, opened 2025-11-27
disko-images: switch back from xcp to cp for improved stability

Mic92:


brianmcgillion:
so far there have not been any repeat issues. All the PRs where this is applied, seem to be working well.

Mic92:
@brianmcgillion Vadim reported more issues. So now I made configurable. if set to 1, it should be the old behavior.

PR #1167, opened 2025-11-28
flake: export lib with references to nixpkgs input

wolfgangwalther:
This previously pointed at `<nixpkgs/...>`, which might be anything. It caused this error for me, when I tried to use `lib.makeDiskoTest` downstream:

```
error: cannot look up '<nixpkgs/nixos/tests/make-test-python.nix>' in
pure evaluation mode (use '--impure' to override)
```

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -24,6 +24,9 @@
 
       diskoLib = import ./lib {
         inherit (nixpkgs) lib;
+        makeTest = import (nixpkgs + "/nixos/tests/make-test-python.nix");
+        eval-config = import (nixpkgs + "/nixos/lib/eval-config.nix");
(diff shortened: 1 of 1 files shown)


Enzime:
Not sure if it would be better to change these to be in terms of `pkgs`: 

https://github.com/nix-community/disko/blob/aecba248f9a7d68c5d1ed15de2d1c8a4c994a3c5/tests/default.nix#L1-L6

like:

```nix
 { 
   pkgs ? import <nixpkgs> { },
   makeTest ? import (pkgs.path + "/nixos/tests/make-test-python.nix"),
   eval-config ? import (pkgs.path + "/nixos/lib/eval-config.nix"),
   qemu-common ? import (pkgs.path + "/nixos/lib/qemu-common.nix"),
 }:
```

PR #1171, opened 2025-12-04
Propagate bcachefs_filesystem mountOptions to subvolumes

VlaDexa:
Make subvolume mountOptions also include options for the whole
filesystem, but only when the filesystems mountpoint is not specified

My reasoning behind this is that if user specifies mountOptions for the filesystem, but doesn't mount it, it's likely they want those options to apply to subvolumes of that filesystem.

### DIFF
--- a/lib/types/bcachefs_filesystem.nix
+++ b/lib/types/bcachefs_filesystem.nix
@@ -35,6 +35,9 @@
       description = ''
         Options to pass to mount.
         The "X-mount.mkdir" option is always automatically added.
+
+        These options will also be applied to subvolumes of this filesystem,
+        if the filesystem itself isn't mounted directly.
       '';
       example = [
         "noatime"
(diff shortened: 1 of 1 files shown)


nothingnesses:
It might be worth either modifying the `mounted_subvolumes_in_multi` example in https://github.com/nix-community/disko/blob/master/example/bcachefs.nix (which doesn't have a `mountpoint`, so it should be suitable for this) or adding a new entry in `bcachefs_filesystems`, to add one or more `mountOptions` that will be applied to all its subvolumes, as well as updating or adding the corresponding test in https://github.com/nix-community/disko/blob/master/tests/bcachefs.nix , to demonstrate this works properly.

Alternatively, it might instead be worth adding an explicit `subvolumesMountOptions` configuration option for the `bcachefs_filesystem` type that can be used to apply the mount options listed to all its subvolumes, which I think would be less confusing. It'd be good to have corresponding examples and tests for this too.

Alternatively, you can just set the `mountOptions` for each of the subvolumes manually, which would require no code changes.

PR #1174, opened 2025-12-09
Add extraArgs for mdadm 

fidgetingbits:
This PR adds a simple `extraArgs` option for `mdadm` that allows you to pass something like `--assume-clean` during disk creation. A simple test that passes an argument is also included.

Being able to pass `--assume-clean`, for the sake of example, allows you to do things like more easily avoid initial resync during raid array creation, which can be multi-hour process depending on the size of the array, and otherwise a bit non-intuitive to cancel using `/proc/mdstat`.

### DIFF
--- a/example/mdadm-raid5-args.nix
+++ b/example/mdadm-raid5-args.nix
@@ -1,0 +1,66 @@
+{
+  disko.devices = {
+    disk = {
+      disk1 = {
+        type = "disk";
+        device = "/dev/my-disk";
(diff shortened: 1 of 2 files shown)


Enzime:
Can you run `nix fmt .` on each commit?

fidgetingbits:
Done

Enzime:
LGTM

Thanks for the contribution :)

PR #1177, opened 2025-12-17
bcachefs: fix empty label formatting failure

crasm:
Fixes an argument handling error with the bcachefs module where the ultimate command ends up as `bcachefs format --label= …`.

This removes the label flag entirely when the nix configured label is the empty string.

### DIFF
--- a/example/bcachefs.nix
+++ b/example/bcachefs.nix
@@ -99,7 +99,8 @@
         type = "disk";
         content = {
           type = "bcachefs";
-          filesystem = "empty_default_label";
+          label = "";
+          filesystem = "empty_label";
         };
       };
     };
@@ -148,9 +149,9 @@
       };
 
       # Example bcachefs filesystem without a label
-      empty_default_label = {
+      empty_label = {
         type = "bcachefs_filesystem";
-        mountpoint = "/empty_default_label";
+        mountpoint = "/empty_label";
       };
     };
   };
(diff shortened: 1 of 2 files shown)


Enzime:
Can you update the bcachefs test?

Also what’s the error here, is `--label=` not supported?

crasm:
> Also what’s the error here, is `--label=` not supported?

The bcachefs tool was giving an error when the literal argument `--label=` was passed without any value.

I’m not sure a test for this is worth keeping around another bcachefs partition/instance with no label? 

The label is also currently the only special case option passed to `bcachefs format` that can be configured inadvertently to be empty.

edit: having a hard time verifying this works; will see about adding a test

crasm:
Let me know if you think I should grep the output of `bcachefs show-super’ to confirm there’s no label. In my opinion, testing that the default empty label works should be enough.

Enzime:
I think it would be good if your disko configuration explicitly sets the label to empty so that it explicitly tests the empty label situation rather than relying on the default and if you could assert on the output of `bcachefs show-super` that would be good

crasm:
The test is needed for the implicit default as much for the volume creation. It's the most likely scenario a user would stumble upon this, and how I tripped on it. Since the field is optional, leaving it out shouldn't break the config.

Let's say the default label were to be removed at some point (or made null)... Then the test will fail. Which is correct from the perspective of a user relying on the default in their config.

Enzime:
Maybe two tests then, one to verify the most basic default config and one to verify what happens when you have the field set to empty

crasm:
I want to avoid cluttering the example, so I changed the test to use the empty string specifically

PR #1181, opened 2025-12-28
Fix broken bcachefs test

nothingnesses:

PR #1183, opened 2025-12-31
Fix update flake lock action

nothingnesses:
Supersedes #1181.


Enzime:
It looks like this is missing the flake lock update

nothingnesses:
> It looks like this is missing the flake lock update

@Enzime I cherry-picked the flake.lock commit from the upstream update_flake_lock_action branch, but it looks like the x86_64-linux.zfs-over-legacy check is still hanging/timing out on "emergency.service: Unable to locate executable 'plymouth': No such file or directory".

Issue #331, opened 2023-08-26
btrfs example and enabling compression failing


sedlund:
https://github.com/nix-community/disko/blob/4eed2457b053c4bbad7d90d2b3a1d539c2c9009c/example/btrfs-subvolumes.nix#L31

The `/rootfs` subvolume is the first mounted in this example, and it does not have compression enabled so the device will not have compression enabled by btrfs.  The following subvolumes will fail to enable compression with the mount script disko creates/uses.

See: https://github.com/numtide/nixos-anywhere/issues/188

When installing with nixos-anywhere the install will not be compressed.  Once the system is installed and rebooted, compression is correclty enabled on the whole device with however fstab filesytems are mounted.

So there are a couple things:

1. The example should be updated 

2. Since `mount` silently fails to enable compression when mounting a fs with the option if the device is already mounted with it disabled (mount exits with 0 in this case), should the disko mount script verify the mount options that were requested are in fact set?

In this example it can `-o remount` the subvolume mount and it will cause the device to be remounted with compression enabled.

octylFractal:
`btrfs` itself mentions that setting most mount options for subvolumes after the first will have no effect, so this should definitely be checked and error/warn: https://btrfs.readthedocs.io/en/latest/ch-mount-options.html

DerRockWolf:
Also just stumbled upon this. Does somebody know what "most mount options" actually means 😅

PR #1200, opened 2026-01-26
docs(btrfs): update example to better explain btrfs mount options

sedlund:
ref: https://github.com/nix-community/disko/issues/331


sedlund:
Prefer #1221  (when ready)

PR #1208, opened 2026-02-12
Support more disks in VM

jgus:
This adds support for much larger numbers of disks in VMs. In my case, I wanted this so I could verify a config before deploying to a multi-array server.

The core limitation is that qemu virtio-blk-pci disks occupy a PCI slot each, and a large array can quickly exhaust available slots. This uses a PCI-to-PCI bridge to add another 32 slots, and, if we have more than 31 disks, starts daisy-chaining more bridges onto that.

- Expanded the device name list up to `/dev/vdzz`
- Added a PCI bridge (or a daisy chain of them), and added flags to each disk device to attach it to the appropriate bridge
- Added an example with a zpool with 36 disks to test/demonstrate

### DIFF
--- a/lib/interactive-vm.nix
+++ b/lib/interactive-vm.nix
@@ -34,10 +34,11 @@ let
     deviceExtraOpts.bootindex = "1";
     deviceExtraOpts.serial = "root";
   };
-  otherDisks = map (disk: {
+  otherDisks = lib.imap0 (i: disk: {
     name = disk.name;
     file = ''"$tmp"/${lib.escapeShellArg disk.imageName}.qcow2'';
     driveExtraOpts.werror = "report";
+    deviceExtraOpts.bus = "bridge${toString (i / 31 + 1)}";
   }) (builtins.tail disks);
 
   diskoBasedConfiguration = {
@@ -68,6 +69,19 @@ in
   virtualisation.useDefaultFilesystems = false;
   virtualisation.diskImage = null;
   virtualisation.qemu.drives = [ rootDisk ] ++ otherDisks;
+
+  # Using `networkingOptions` instead of `options` here because it's added _before_ the drive options, where `options` is added at the end :-P
+  virtualisation.qemu.networkingOptions = lib.mkAfter (
+    let
+      # A PCI bridge takes one slot and adds 32, so we'll want one for every 31 drives
+      count = ((builtins.length otherDisks) + 30) / 31;
+      parentBridge = i: lib.optionalString (i > 0) "bus=bridge${toString i},";
+    in
+    (builtins.genList (
+      i: "-device pci-bridge,id=bridge${toString (i + 1)},${parentBridge i}chassis_nr=${toString (i + 1)}"
+    ) count)
+  );
+
   boot.zfs.devNodes = "/dev/disk/by-uuid"; # needed because /dev/disk/by-id is empty in qemu-vms
   boot.zfs.forceImportAll = true;
   boot.zfs.forceImportRoot = lib.mkForce true;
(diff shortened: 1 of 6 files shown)


lib/interactive-vm.nix
```diff
@@ -68,6 +69,19 @@ in
   virtualisation.useDefaultFilesystems = false;
   virtualisation.diskImage = null;
   virtualisation.qemu.drives = [ rootDisk ] ++ otherDisks;
+
+  # Using `networkingOptions` instead of `options` here because it's added _before_ the drive options, where `options` is added at the end :-P
```
CRTified:
```suggestion
  # Using `networkingOptions` instead of `options` here because it's added before the drive options, where `options` is added at the end
```

Looking only at the diff - is the order of arguments important here? If so, it would be nice to mention that it needs to be before the drive options (which I am guessing here)

lib/interactive-vm.nix
```diff
@@ -68,6 +69,19 @@ in
...
+  virtualisation.qemu.networkingOptions = lib.mkAfter (
+    let
+      # A PCI bridge takes one slot and adds 32, so we'll want one for every 31 drives
+      count = ((builtins.length otherDisks) + 30) / 31;
+      parentBridge = i: lib.optionalString (i > 0) "bus=bridge${toString i},";
+    in
+    (builtins.genList (
+      i: "-device pci-bridge,id=bridge${toString (i + 1)},${parentBridge i}chassis_nr=${toString (i + 1)}"
```
CRTified, line 81: Just a note: Assuming more than one bridge gets added, they will be daisy-chained. This is performance-wise likely not "the best" approach, but given that this is only relevant for rare instances of 62 drives or more, it's nothing to worry about IMHO, but the "perfect" solution would be some kind of balanced 32-tree.

Might be useful to leave a comment about this, though.

CRTified (changes requested):
At least from my point of view, this PR is fine.
It solves the issue of qemu being limited in the number of drives (see [my comment on the issue](https://github.com/nix-community/disko/issues/1225#issuecomment-5655804460)) and it introduces a test that implicitly validates that the mechanism for one additional controller successfully works.

I cannot state anything about the ZFS setup, as I do not use it and don't have any experience with it.

**Requested Change:** It would be great to test that it works for 2 additional controllers, and not only one, to make sure that the calculations for indicies are correct (even though they look correct to me).

Should be sufficient to bump the number of drives to something like 32/32/32 for each raidz3 group to make this clear.

jgus:
> **Requested Change:** It would be great to test that it works for 2 additional controllers, and not only one, to make sure that the calculations for indicies are correct (even though they look correct to me).
> 
> Should be sufficient to bump the number of drives to something like 32/32/32 for each raidz3 group to make this clear.

The existing test already contains 36 drives, which does require 2 additional controllers.

lib/interactive-vm.nix
jgus: Maybe. Depends entirely on how the VM implements these things. In any case, it's pretty irrelevant for our use case, where all the software cares about is the view of disks provided by Linux, and nothing about their physical topology.

CRTified:
> The existing test already contains 36 drives, which does require 2 additional controllers.

Ah, I understood it as "~31 drives are possible without additional controllers", so that it needs >=64 drives to get two additional controllers. Thanks, then I'd be happy as it is right now :)

CRTified (approved):

lib/interactive-vm.nix
CRTified: That's what I meant with "not the best approach". It was just a thought that I had while looking at the code, totally fine for me to leave it as-is :)

PR #1220, opened 2026-03-02
Fix #331: Fix btrfs compression in example and remount subvolumes

### LINKED ISSUE #331: btrfs example and enabling compression failing

sliortega295-ops:
Fixes #331

Brief description of changes:
- Added `compress=zstd` to the `/rootfs` subvolume in `example/btrfs-subvolumes.nix`.
- Modified `lib/types/btrfs.nix` to append a `mount -o remount` command immediately after the initial mount of each subvolume (and the base mountpoint). This ensures that compression and other mount options are correctly applied, even if they were omitted from the first mounted subvolume on the device.

### DIFF
--- a/lib/types/btrfs.nix
+++ b/lib/types/btrfs.nix
@@ -193,7 +193,7 @@ in {
               (subvol.mountpoint != null)
               (
                 let
-                  mergedOptions = lib.unique (config.mountOptions ++ subvol.mountOptions ++ ["subvol=${subvol.name}"]);
+                  mergedOptions = lib.unique (config.mountOptions ++ subvol.mountOptions ++ [ "subvol=${subvol.name}" ]);
                 in {
                   ${subvol.mountpoint} = ''
                     if ! findmnt "${config.device}" "${rootMountPoint}${subvol.mountpoint}" > /dev/null 2>&1; then


sedlund:
One concern: since most Btrfs mount options are filesystem-wide, doing a remount per mount effectively makes the last mount determine the active options for the whole filesystem. In disko, I think the order is derived from attribute iteration (lexicographic subvolume keys), which isn’t very explicit. 🤔

This improves the “first mount wins” issue, but it shifts the behaviour to “last mount wins” without defining that policy clearly.

sliortega295-ops:
Thank you for the feedback @sedlund! You're absolutely right about the concern with the \"last mount wins\" behavior.

I've been thinking about this, and I see a few potential approaches:

## Current Issue
The remount approach I implemented does solve the immediate problem (compression not being applied), but it creates a new issue where the mount order (determined by lexicographic attribute iteration) implicitly determines which options take effect.

## Potential Solutions

### Option 1: Explicit filesystem-wide options
Add a top-level `filesystemOptions` field that explicitly defines options that should apply to the entire btrfs filesystem, separate from per-subvolume `mountOptions`. This would make it clear which options are filesystem-wide.

### Option 2: Merge and validate
Collect all mount options from all subvolumes, merge them, and validate that there are no conflicts. If conflicts exist, fail with a clear error message explaining which subvolumes have conflicting options.

### Option 3: First mount with merged options
Apply all unique mount options from all subvolumes on the first mount, then subsequent mounts just specify the subvolume. This ensures all options are applied consistently.

I'm leaning towards **Option 3** as it's the most straightforward and doesn't require schema changes. What do you think?

I'll also fix the treefmt issues in the next update.

sedlund:
I implemented the main filesystem being the default options in #1221.  Subvolumes inherit these and some options can then be added.

PR #1223, opened 2026-03-03
feat: add flake-parts installer image options

different-name:
Creating an offline installer configuration is [already demonstrated in the disko docs](https://github.com/nix-community/disko/blob/master/docs/disko-install.md#example-for-a-nixos-installer), however it's pretty dense and I think a module would be a great replacement for copying that large chunk of code from the docs

This PR adds the `disko-install` option set to the flake-parts module, which generates installer images with dependencies & install script included

Minimal config:

```nix
disko-install.hosts = {
  my-host.disks.main = /dev/disk/by-id/some-disk-id;
};
```

Exhaustive config:

```nix
disko-install.hosts = {
  my-host = {
    nixosConfiguration = "my-host";
    disks.main = /dev/disk/by-id/some-disk-id;
    image = {
      formats = [ "iso-installer" "hyperv" ];
      modules = [ /*<nixos module>*/ ];
    };
    script = {
      name = "my-cool-installer";
      extraArgs = [ "--dry-run" ];
    };
  };
};
```

### TODO

This works to generate installer images (I have tested with `iso-installer`) with a `install-nixos-unattended` script, however it still needs work:

- Stricter typing for formats option, as it currently is just `listOf str`
- Documentation
- Tests?

### DIFF
--- a/flake-module.nix
+++ b/flake-module.nix
@@ -1,43 +1,211 @@
-{ lib, ... }:
 {
-  options.flake.diskoConfigurations = lib.mkOption {
-    type = lib.types.lazyAttrsOf lib.types.raw;
-    default = { };
-    description = "Instantiated Disko configurations. Used by `disko` and `disko-install`.";
-    example = {
-      my-pc = {
-        disko.devices = {
-          disk = {
-            my-disk = {
-              device = "/dev/sda";
-              type = "disk";
-              content = {
-                type = "gpt";
-                partitions = {
-                  ESP = {
-                    type = "EF00";
-                    size = "500M";
-                    content = {
-                      type = "filesystem";
-                      format = "vfat";
-                      mountpoint = "/boot";
-                      mountOptions = [ "umask=0077" ];
-                    };
-                  };
-                  root = {
-                    size = "100%";
-                    content = {
-                      type = "filesystem";
-                      format = "ext4";
-                      mountpoint = "/";
+  lib,
+  config,
+  inputs,
+  self,
+  ...
+}:
+let
+  inherit (lib)
+    mkOption
+    types
+    ;
+
+  nixpkgs = inputs.nixpkgs or (throw "No nixpkgs input found");
+
+  flakeOutPaths =
+    let
+      collector =
+        parent:
+        map (
+          child:
+          [ child.outPath ] ++ (if child ? inputs && child.inputs != { } then (collector child) else [ ])
+        ) (lib.attrValues parent.inputs);
+    in
+    lib.unique (lib.flatten (collector self));
+
+  mkInstallClosureModule =
+    nixosConfiguration:
+    { pkgs, ... }:
+    let
+      dependencies = [
+        nixosConfiguration.config.system.build.toplevel
+        nixosConfiguration.config.system.build.diskoScript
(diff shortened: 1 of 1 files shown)


different-name:
Marking this as ready for review, since the install script is implemented & I'm happy with the way it's configured

mightyiam:
How about some tests?

different-name:
Tests is on the todo above, I don't have much experience with CI, and wanted to gauge if this is something desirable before continuing with docs & tests

Jayman2000:
I had actually started working on implementing a similar feature a while ago (if you want, you can check out an old version of what I had been working on [here](https://codeberg.org/JasonYundt/disko-backup/commit/8d151ca6cc214a2e11bbfa9fd15bd4681ac5db40)). I had gotten stuck trying to write tests for the functionality that I had added. Specifically, I thought that I needed [this Nixpkgs pull request that adds a `lib.strings.escapePythonBytes` function](https://github.com/NixOS/nixpkgs/pull/458502) to be merged before I would be able properly write tests for the functionality that I had added.

More recently, I figured out workarounds that would allow me to write correct tests even though the `lib.strings.escapePythonBytes` pull request still hasn’t been merged yet. As a result, I have finished my pull request, and [I have submitted it as an alternative][my-pr] to this pull request. Compared this this pull request, I think that my pull request has a few advantages:

- With this pull request, users need to do several different things which are normally optional: users must use **disko** for their partitions and filesystems, users must use flakes, and users must use flake-parts.

    With my pull request, users only need to do one thing which is normally optional: users must use **disko** for their partitions and filesystems.

- With this pull request, the deployee configuration (the NixOS configuration that you are going install onto a system) gets evaluated at least twice. First, the deployee configuration is evaluated when the install medium is created. Then, the deployee configuration gets reevaluated every time an installation is performed  (when `disko-install` is run).

    With my pull request, the deployee configuration gets evaluated once when the install medium is created. You can use that install medium as many times as you want without ever having to reevaluate the deployee configuration.

- With this pull request, you need to specify the same disk(s) twice in different locations. The first time is in the deployee NixOS configuration as `disko.devices.disk.<name>.device = "<path>";`. The second time is in the flake-parts configuration as:

    ```nix
    disko-install.hosts = {
      my-host.disks.<name> = <path>;
    };
    ```

    With my pull request, you only need to declare disks once.

- With this pull request, you have to explicitly add additional configuration (i.e., set the `disko-install.hosts.<name>` option).

    With my pull request, you don’t need to explicitly add any additional configuration. You can just create a NixOS configuration that uses **disko** for its partitions and filesystems, and then run this command:

    ```bash
    nixos-rebuild build-image --image-variant disko-unattended-install-iso
    ```

Since you have been working on a similar feature already, I would love it if you could take a look at [my pull request][my-pr] and share your thoughts.

Also, sorry for coming out of nowhere and submitting a conflicting PR. I wish that I had submitted my work as a WIP pull request without tests a while ago so that we both didn’t end up putting in a bunch of duplicated effort.

[my-pr]: https://github.com/nix-community/disko/pull/1272

different-name:
You're fine! Unfortunately life has gotten pretty busy for me, and I don't have much mental bandwidth to spare. So I won't be working on this for the foreseeable future

PR #1284, opened 2026-08-06
gpt: reject '0<unit>' partition start values that crash sgdisk

### LINKED ISSUE #702: sgdisk behaves confusingly when specifying `partitions.${name}.start = "0M"`
`partitions.${name}.start = "0M"` causes `sgdisk` to begin at the start of next free space on disk, which crashes `disko` with confusing errors.

This is despite the manpage setting a distinction between plain 'N' as sector numbers and 'N[K/M/G]' as logical size/byte positions, and stating that '0' is the 'next free space' position, not 0 with a letter.

In any case it would be nice to have a check for `start = '0*'` in `disko` as the `sgdisk` is an implementation detail. Start should be 1M (or whatever is physically possible minimum considering GPT and such) or not specified for the first partition, with only size being specified.

yottanami:
`sgdisk` treats a bare `"0"` for `start` as "next available range", but `"0"` directly followed by a size unit (`0M`, `0K`, `0G`, `0T`, `0P`) is parsed as a literal zero-byte offset instead, which collides with the GPT header/table area and makes `sgdisk` fail with a confusing error instead of doing what the user expects (see the errors quoted in the issue).

This adds a type check on the `start` option (`lib/types/gpt.nix`), following the same pattern used by the `size` option a few lines above, that rejects this specific footgun at Nix-eval time with a clear description, while still allowing a bare `"0"` and any non-zero offset (e.g. `"1M"`, `"2048"`, `"-1M"`).

Checked all in-repo usages of `start` on gpt-typed partitions (examples and tests): they all use non-zero offsets like `"1M"` already, so this shouldn't affect any existing configuration.

I don't have a local Nix environment to run `nix flake check` / the test suite against this change, so flagging that so a maintainer/CI can verify the syntax and formatting (`nixfmt`) directly.

Fixes #702

### DIFF
--- a/lib/types/gpt.nix
+++ b/lib/types/gpt.nix
@@ -183,9 +183,19 @@ in
                 description = "Alignment of the partition, if sectors are used as start or end it can be aligned to 1";
               };
               start = lib.mkOption {
-                type = lib.types.str;
+                type = lib.types.addCheck lib.types.str (
+                  start: builtins.match "[+-]?0+[KMGTP]" start == null
+                );
                 default = "0";
-                description = "Start of the partition, in sgdisk format, use 0 for next available range";
+                description = ''
+                  Start of the partition, in sgdisk format, use 0 for next available range.
+
+                  Note: sgdisk treats a bare "0" as "next available range", but "0"
+                  followed directly by a size unit (e.g. "0M") is a literal
+                  zero-byte offset that collides with the GPT header, causing a
+                  confusing sgdisk failure. Use a bare "0" or a non-zero offset
+                  (e.g. "1M") instead.
+                '';
               };
               end = lib.mkOption {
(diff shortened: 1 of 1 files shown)


yottanami:
Pushed a follow-up commit: the `jsonTypes` CI check failed because `lib/default.nix`'s doc-generation stub lib doesn't implement `lib.types.addCheck` (confirmed by reading `typesSerializerLib`). Guarded the lookup with `or` so the stub degrades to a no-op there while real evaluation still gets full `addCheck` validation.

Also noting: the `treefmt` failures on this PR (`openTempFileWithDefaultPermissions: permission denied`) appear to be a pre-existing CI infra issue unrelated to this change — the same failure shows up on #1280, an unrelated open PR from another contributor.

PR #1286, opened 2026-08-07
bugfix: Detached headers idempotence check

Hussainity:
## Summary
Detached LUKS headers were not handled correctly by the `_create` idempotence check in `lib/types/luks.nix`.

The old logic still probed `config.device`. For detached-header setups, that meant rerun could mis-detect an already-initialized LUKS device and fall back into luksFormat.

## Fix
Probe the actual LUKS header target instead of the payload device, and make the create guard depend only on `cryptsetup isLuks` for that target.

## Test
Add a detached-header regression test that exercises the direct-mode idempotence path.


Hussainity:
@Lassulus requesting maintainer review!

PR #1291, opened 2026-08-22
feat(btrfs,luks): better default checksum collision resistance with xxhash

nikelborm:
Choose xxHash, which will catch errors with a higher probability. A single encrypted bit flip causes a 128-bit flip on the decrypted side. This makes the CRC32C property of being certain to catch 100% of errors in 1/2/3-bit flip use cases disappear and become probabilistic `1-2⁻³²`. And since we're already in a non-deterministic zone, it's better to choose the algorithm that has stronger collision resistance.

### DIFF
--- a/example/luks-btrfs-raid.nix
+++ b/example/luks-btrfs-raid.nix
@@ -54,8 +54,12 @@
                 content = {
                   type = "btrfs";
                   extraArgs = [
-                    "-d raid1"
+                    "--data raid1"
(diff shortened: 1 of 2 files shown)


nikelborm:
It probably would've been better to move this deeper into abstraction, like making disko detect btrfs-on-luks/btrfs-on-lvm-on-luks stacking and automatically add the necessary formatting option, but I'm not competent enough at the moment to do that.

PR #1292, opened 2026-09-01
install-cli: source nixos configurations from package outputs

bricked-dev:
`disko-install` only searched for NixOS configurations within the `nixosConfigurations` flake output. Related tools like `nixos-install` and `nixos-rebuild` additionally search for NixOS configurations within the `packages` and `legacyPackages` flake outputs. This commit adds this behavior to `disko-install`.

### DIFF
--- a/install-cli.nix
+++ b/install-cli.nix
@@ -7,7 +7,11 @@
   rootMountPoint ? "/mnt",
 }:
 let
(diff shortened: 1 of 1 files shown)


bricked-dev:
For instance, my NixOS configuration is stored at `legacyPackages.x86_64-linux.nixosConfigurations.carrot`.

